PR #44, opened 2011-03-27
Added simple method that forces the application gain focus

onedayitwillmake:
Just a tiny method

    void AppBasic::getFocus()
	{
	#if defined( CINDER_COCOA )
		[[NSApplication sharedApplication] activateIgnoringOtherApps:YES];
	#elif defined( CINDER_MSW )
		// Something else
	#endif
	}


andrewfb:
Apologies for the delay - I think for the time being we'll leave this functionality out of App, in part because it's not a concept that can be matched on all operating systems (ie iOS).

PR #102, opened 2011-09-22
Added App::isMinimized() (OSX)

rsodre:
I use this on my Syphon apps, when I don't need to draw to the screen when the app minimized, increasing my framerate a little bit.
I implemented only for OSX, sorry i have no Windows or Linux here.
But I'm not sure if you guys like this kind of pull.
Should I pull an OSX only solution, or not?
If not, please close this.


andrewfb:
Sorry for the minor delay Roger :). This should be addressed in a cross-platform and multi-window-compatible way with App's getSignalWillResignActive() and getSignalDidBecomeActive() these days.

PR #106, opened 2011-10-19
Add OpenUrlEvent to AppCocoaTouch

RandomEtc:
Adds an `OpenUrlEvent` and a `urlOpened` method to `AppCocoaTouch` to allow Cinder apps to respond to requests from other iOS apps. This allows a Cinder app to be opened as a result of a web link, for example. We're using this to get back to our app after completing an OAuth request in mobile safari.

I have also adjusted AppCocoaTouch's delegate to use `application:didFinishLaunchingWithOptions` which is the recommended place to complete your app setup on iOS these days.

The intention is that `setup` will run before the first `urlOpened`. Once your app is running, or in the background, it might receive more than one `urlOpened` call or event. The OpenUrlEvent is a class rather than a string in case we want to add more values to it later. I only needed the URL, but there are other notifications that can wake up an app.

To receive a URL notification, your app's plist must declare the url-scheme supported by your app. You can edit this in Xcode but the resulting values are something like:

```xml
	<key>CFBundleURLTypes</key>
	<array>
		<dict>
			<key>CFBundleURLSchemes</key>
			<array>
				<string>my-fancy-scheme</string>
			</array>
		</dict>
	</array>
```

Your app then receives notifications whenever someone opens a `my-fancy-scheme://something` URL.

This Stack Overflow post provided some guidance on real world implementation, but honestly it seemed pretty straight forward: http://stackoverflow.com/questions/3612460/lauching-app-with-url-via-uiapplicationdelegates-handleopenurl-working-under

This is not an urgent request but I wanted to clean it up and get it out there. Thoughts welcome!

### DIFF
--- a/include/cinder/Buffer.h
+++ b/include/cinder/Buffer.h
@@ -78,6 +78,6 @@ class Buffer {
 };
 
 Buffer compressBuffer( const Buffer &aBuffer, int8_t compressionLevel = DEFAULT_COMPRESSION_LEVEL, bool resizeResult = true );
-Buffer decompressBuffer( const Buffer &aBuffer, bool resizeResult = true );
+Buffer decompressBuffer( const Buffer &aBuffer, bool resizeResult = true, bool useGZip = false );
 
 } //namespace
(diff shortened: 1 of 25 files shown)


andrewfb:
Hey Tom - a very belated thank you for this. We'll use this as a reference in the future but will close it as an active PR until we're in a position to tackle it for all platforms.

PR #107, opened 2011-10-23
OpenGL ES: Unintuitive Texture::getCleanWidth() Behavior

superquadratic:
For OpenGL ES, ```Texture::getCleanWidth()``` does _not_ return ```mCleanWidth```, but ```mWidth```; the same applies to ```Texture::getCleanHeight()```. This came as a little surprise to me.

The behavior was introduced in 512eb51, and I can imagine it being implemented to circumvent cases where ```mCleanWidth``` was not initialized correctly. However, this has been fixed in f6ffd88, so I think 512eb51 should now be reverted -- or did I overlook something?


andrewfb:
Apologies for the delay on this - in the recent GL rewrite we redesigned the distinction between clean bounds and actual size, so this should be working more intuitively now.

PR #118, opened 2011-12-11
drawBillboard generalizations

artm:
Extra variants accepting texture coordinates in different forms and unrotated billboard variants. Supplying texture coordinates is useful for implementing sprite sheets, where multiple sprites or multiple animation frames occupy the same texture.


andrewfb:
Thanks for these and sorry for the delayed response; I think for the time being we're keeping billboarding to the basics in terms of the convenience methods in the new GL API though.

PR #143, opened 2012-02-08
Support for Tablets (mac only)

lichtfaktor:
This will enable basic tablet support on mac. Pressure and Tilt will be supported by calling getTabletPressure() or getTabletTilt().

### DIFF
--- a/include/cinder/Display.h
+++ b/include/cinder/Display.h
@@ -43,6 +43,8 @@
 
 namespace cinder {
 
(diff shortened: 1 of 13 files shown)


andrewfb:
Thanks for this; we'll definitely keep this as reference but I am going to close it as an active issue since we're not ready to tackle cross-platform tablet support yet. We'll revisit this when we eventually pursue this with MSW / WinRT and possibly other platforms.

PR #174, opened 2012-05-17
Changed the default attentuation values of Light.cpp

timmb:
Changed the default attentuation values of Light.cpp to match those specified in http...://www.glprogramming.com/red/chapter05.html#name4 (0 for linear and quadratic factors, 1 for the constant factor).


andrewfb:
Thanks and obviously apologies for a bit of a delay :) - this made sense at the time but in light of the 'glNext' changes now merged into master I'm going to close it.

PR #262, opened 2012-12-18
new easing EaseStep

q-depot:
this is a very simple easing, it holds the current value until the end, then it updates to the target value.


andrewfb:
Closing this in favor of referenced future pull request.

PR #362, opened 2013-07-31
Bin packing

paulhoux:
Added classes for efficient bin packing, both a single bin implementation and a multi bin implementation that can create additional, equally-sized bins. Using these classes we can write things like improved texture fonts, shadow map atlasses etc.

See the BinPackerTest application for implementation details.

### DIFF
--- a/include/cinder/BinPacker.h
+++ b/include/cinder/BinPacker.h
@@ -74,6 +74,9 @@ public:
 	//!
 	virtual BinPackerBase&	setSize( unsigned width, unsigned height ) = 0;
 	virtual BinPackerBase&	setSize( const Vec2i &size ) = 0;
+
+	//
+	virtual std::vector<BinnedArea>	pack( const std::vector<Area> &rects, bool allowRotation = false ) = 0;
 	
 	//!
 	Vec2i		getSize() const { return Vec2i( mBinWidth, mBinHeight ); }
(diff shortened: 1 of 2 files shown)


simongeilfus:
Hello gents,

Sorry for re-opening this but I was wondering if it was still planned to add bin packing to Cinder? 

Paul's (bin_packing branch)[https://github.com/paulhoux/Cinder/tree/bin_packing] looks great and could still be easily merged to master. (There's just some missing ```const``` [here](https://github.com/paulhoux/Cinder/blob/bin_packing/include/cinder/BinPacker.h#L128) that ```std::sort``` doesn't seem to like but appart from that it works really nicely).

Is there any reason this was dropped? Too specific for the core?

simongeilfus:
Hey @paulhoux !

I found a small issue with your code. The packing will fail if you insert too much Area of the same size in a row. Try inserting 5 Area( 0, 0, 64, 64 ) in a 128x128 packer to see what I mean. The last Area will be inserted in the result vector but at the origin, therefore creating overlapping and ruining the whole packing. 

The issue comes from [this part](https://github.com/paulhoux/Cinder/blob/bin_packing/src/cinder/BinPacker.cpp#L83). Basically inserting the 4th area will fill perfectly the available area making the mAvailable empty. Any following insert will be considered by this if as the first one and will fail your ```BinPackerTooSmallExc``` detection.

This can easily be fixed by moving the code that insert the initial available area into both the constructor and the clear method.

Any update on my last comment?

PR #405, opened 2014-01-25
Added clear function to Surface

wdlindmeier:
Calling clear() sets the value of every pixel to the Color that's passed in.

### DIFF
--- a/src/cinder/Surface.cpp
+++ b/src/cinder/Surface.cpp
@@ -339,6 +339,19 @@ void SurfaceT<T>::copyFrom( const SurfaceT<T> &srcSurface, const Area &srcArea, 
 	else
 		copyRawRgb( srcSurface, srcDst.first, srcDst.second );
 }
+	
+template<typename T>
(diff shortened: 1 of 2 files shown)


src/cinder/Surface.cpp
```diff
@@ -339,6 +339,19 @@ void SurfaceT<T>::copyFrom( const SurfaceT<T> &srcSurface, const Area &srcArea,
...
+void SurfaceT<T>::clear( const ColorAT<T> &color )
+{
+	SurfaceT<T>::Iter iter = getIter();
+    while( iter.line() )
+    {
+        while( iter.pixel() )
+        {
+			setPixel(Vec2f(iter.getPos()), color);
+        }
+    }
```
drewish: Seems like some spaces crept in here, be good to swap them for tabs.

andrewfb:
OK - at least for the time being I think we're going to stick with the current arrangement where Surface manipulation is done through separate free functions. Open to revisiting this at some future date but going to close for now.

PR #425, opened 2014-05-15
Allow Serial to find virtual COM ports on Windows

heisters:
This fixed an issue I was having where a virtual serial device was not being found. I believe this should fix issues mentioned on the forum:

https://forum.libcinder.org/topic/serial-getdevices-always-returns-0
https://forum.libcinder.org/topic/serial-communication-with-arduino

### DIFF
--- a/src/cinder/Serial.cpp
+++ b/src/cinder/Serial.cpp
@@ -179,15 +179,13 @@ const std::vector<Serial::Device>& Serial::getDevices( bool forceRefresh )
 
 	ZeroMemory( &devInfo, sizeof( devInfo ) );
 	devInfo.cbSize = sizeof( devInfo );
-	while( ::SetupDiEnumDeviceInfo( devInfoSet, deviceIndex++, &devInfo ) )
-	{
+	while( ::SetupDiEnumDeviceInfo( devInfoSet, deviceIndex++, &devInfo ) )	{
 		devInfo.cbSize = sizeof( devInfo );
 
 		::DWORD deviceInterfaceIndex = 0;
 
 		devInterface.cbSize = sizeof( devInterface );
-		while( ::SetupDiEnumDeviceInterfaces( devInfoSet, &devInfo, &guid, deviceInterfaceIndex++, &devInterface ) )
-		{
+		while( ::SetupDiEnumDeviceInterfaces( devInfoSet, &devInfo, &guid, deviceInterfaceIndex++, &devInterface ) ) {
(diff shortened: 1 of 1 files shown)


pizthewiz:
I've created [Cinder-Serial](http://github.com/pizthewiz/Cinder-Serial), a super thin wrapper on [wjwwood/serial](https://github.com/wjwwood/serial) and didn't have a problem communicating to and from an Arduino Uno when running a Windows 8.1 guest in VMware Fusion. My Uno was connected at COM3 and I did install the Arduino app so I don't know if that skipped the whole virtual COM port issue that was called out here or not.

totalgee:
Just curious-- is this `Serial::getDevices()` fix dead?  So the recommended solution is just to use [Cinder-Serial](https://github.com/pizthewiz/Cinder-Serial) (alternatively, just use [wjwwood/serial](https://github.com/wjwwood/serial) directly)?

Because, on Windows, even in 0.9.0, `getDevices()` doesn't list virtual COM ports, such as one for an Arduino MKR1000.

Thanks.

richardeakin:
@totalgee are you saying that the mods in this PR didn't work for you? This isn't dead, we just didn't find time to review it and get it in.

I think using wjwwood/serial / Cinder-Serial is a good interim solution. There has been discussion on rewriting some (or most) of Serial.cpp to use asio as that will be part of the C++ standard, however that doesn't cover enumerating devices so maybe there is something we can learn or use from wjwwood there.

totalgee:
Sorry, I didn't get a chance to test this PR yet, I just took the
enumerating stuff (all that I needed) from wjwwood/serial. It was trivial
to use -- required no changes, just removing some code I didn't need and
adding a link library (#pragma) on Windows. The port enumeration code is
nicely isolated in a single file (one for each platform).

I've used my own serial communication stuff before (using Boost asio), but
in this case I just needed the port enumeration; as you say, that's the
platform-dependent bit!

I can test this PR later, if there is still a chance it may be used. My
serial (USB) device is a Genuino MKR1000.

ryanbartley:
Oh that's great news. That's basically exactly what we were talking about and most likely the way forward (a mix of asio and wjwwood enumeration, license permitting).

richardeakin:
Well license looks to be MIT so it should be fine. Yes I agree - let's get in the wjwwood enumeration improvements first so that you can properly find devices on all of our platforms, afterwards we can work towards moving to use asio for data transfer.

PR #523, opened 2014-10-03
Support compute shaders while still targeting GL 3.2

clinthidinger:
This is a suggestion of how compute shaders could be supported while still targeting GL 3.2.  I replaced the shader string member variables in GlslProg::Format with a map.  I also added a shaderStage function that takes the shader type and the source.  Users can then use compute shaders by doing:

mUpdateProg = gl::GlslProg::create( gl::GlslProg::Format().shaderStage( loadAsset( "particleUpdate.cs.glsl" ), GL_COMPUTE_SHADER ) );

To test this change, I converted the ParticlesGPU sample to use compute shaders in ParticlesCS.  It works under Visual Studio 2013.  I also ran several of the existing samples under _opengl on both osx and windows. 

image


BanTheRewind:
I like this approach. We've been using our own CS wrapper since we haven't really cracked the best way to get it into GlslProg.

We have a particle project which uses the same logic in TF and CS versions. We can run about 2.5 more particles at 60fps with CS than TF.

Compute and SSBOs have become a critical part of our work here. Beyond simulation, we use it for things like tiled rendering and just any place that can benefit from parallelism (which is almost everywhere).

clinthidinger:
Thanks!  Do you know if there is a chance of some of this being merged in?  Also, do you have any videos/images of the particle work that you have done?  I'd like to check it out.

andrewfb:
At this point this PR is too old to be merged (sorry about that - at the time we just weren't in a position to absorb the work to integrate it). However I think we'd be interested in shipping with compute shader support in the upcoming release. I have to admit I'm too new to SSBOs to comment on an interface for them, but I can do the reading. However in the new GlslProg design I think adding Compute should be relatively straightforward, if someone wants to start there...

BanTheRewind:
I need to look again, but is there an explicit requirement for SSBO from the compute update, or would it make sense to just use a generic BufferObj and set the target?

BanTheRewind:
I ask because it's common to use SSBOs with compute for the sole purpose of changing its target later, or pointing a VBO or texture at it.

BanTheRewind:
If you want to PR these changes against a more recent commit of Cinder, I can pull and try them out. I have a few compute usages ready to test against it.

clinthidinger:
BanTheRewind:  I don't want to duplicate your effort, but I'll be happy to get the PR and samples working again.  I probably won't be able to set up my windows box to work on this until tomorrow night, though.

The actual changes to GlslProg::Format were really simple.  I just replaced all of the individual member variables for each shader stage with a map.  I also added a shaderStage function so that users could add member stages to that map, including ones not directly supported by the library.  I did a small wrapper for ssbo's that extends BufferObj too.

The bulk of the work was in the samples.  I'm not sure if there would be any interest in having samples that only work on windows?  Maybe it would be better to have the compute shader sample be optionally  #define'd on inside the ParticlesGPU sample for windows only.

BanTheRewind:
Oh, I really just wanted to try out your gl:: and gl::GlslProg updates. The class we're using is totally separate from GlslProg and is not intended to go into Cinder. It would be nice to try out something designed for the core.

andrewfb:
The internal architecture has changed some for GlslProg but I believe adding compute should be straightforward. Part of our initial negligence in this PR was due to anticipation of those changes, but that class has settled now. And I believe there'd be room for a Windows-only sample(s) - not much to be done about Apple's lack of support.

BanTheRewind:
Right. If iOS gets its own samples, why can't Windows? ;)

richardeakin:
Also worth mentioning is that latest android phones (with ES 3.1) support compute shaders.  I think a separate example for compute shaders would be ideal, as to avoid overcomplicating the existing ParticleSystemGpu sample.

I'm happy to see this feature getting attention.

clinthidinger:
A new pull request has been submitted.

andrewfb:
Closing this as it has been superseded by #985 and #989

PR #525, opened 2014-10-05
Update tutorials

lithium-snepo:
Mostly just fixed the issues introduced by switching to glm, so they're still not representative of "modern" cinder usage, but at least they'll compile. :)


drewish:
looks like this needs a rebase.

richardeakin:
I think the consensus is that we need to rewrite the tutorials for the current state of ci::gl (glNext), so closing this PR and as we'll readdress with a fresh new tutorial.

PR #538, opened 2014-10-20
Added PBOReadBack sample

thyrrestrup:
This sample demonstrates asynchronous texture read back from the GPU to the CPU.
It is using the newly added Pbo class.
It has only been tested in Xcode.
I hope this sample is useful enough to make it into the next release.
pboreadback


andrewfb:
Apologies for the delay - just now looking at this. This seems like a good demonstration of the technique and also gives a justification for its usage. I wanted to minimize the expense of the other operations though to make the FPS difference more pronounced as, at least on my machines, the end framerate was almost the same due to other bottlenecks. Consequently I removed the grayscale portion, but ultimately this is a technical sample, so the look is probably unimportant. Additionally, I made the FBO 1024x1024 to make the readback expense greater. This work is on the `pboReadBackSample` branch. I'll likely merge this sometime tomorrow unless anyone else has suggestions. Thanks for this contribution.

richardeakin:
Icons.

andrewfb:
Icons addressed; merged. Thanks again for this @thyrrestrup

thyrrestrup:
No problem ;-)

PR #542, opened 2014-10-22
Added the ability to specify a custom deleter for your Texture...

paulhoux:
...using the Format's ```setDeleter()``` method. If a custom deleter is specified, the texture's shared pointer will use it. This is a cleaner solution than the one I proposed earlier in #541 .

This raises the question whether or not we should also add this to Cinder's other Format classes (Fbo, Vbo, etc).


andrewfb:
Addressed by bb3eaa8da4ef83818b5be54b2e1a544d28f0bfbe - adding a deleter to the Texture*::Format().

PR #613, opened 2014-11-27
Added per-vertex color support to OBJLoader

wdlindmeier:
Some programs export .OBJ files with per-vertex color values rather than using materials. The values are tacked onto the end of a vertex line. E.g.

Without color data:
v 1.463737 -1.050420 -0.000000

With color data:
v -0.104546 0.865881 -0.440595 0.498039 0.498039 0.498039

This commit adds support for parsing the latter format.

### DIFF
--- a/include/cinder/ObjLoader.h
+++ b/include/cinder/ObjLoader.h
@@ -94,13 +94,12 @@ class ObjLoader : public geom::Source {
 		std::vector<int32_t>	mVertexIndices;
 		std::vector<int32_t>	mTexCoordIndices;
 		std::vector<int32_t>	mNormalIndices;
-		std::vector<int32_t>	mColorIndices;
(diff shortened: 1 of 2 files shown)


richardeakin:
Some background info: http://gamedev.stackexchange.com/a/66270

So it seems that

a) .obj is half baked and doesn't include all necessary features such as per-vertex colors and
b) some companies decide to add them in their own way, specifically [MeshLab](http://meshlab.sourceforge.net/)

Here's a similar request for [Three.js's obj loader](https://github.com/mrdoob/three.js/issues/2083), wherein they decide it isn't worth adding complexity for a non-standard feature addition. I sort of agree, instead a alt. solution would be to implement a loader for a better format (such as .ply).

Closing for now.. can reopen if opinions change.

PR #624, opened 2014-12-04
Stereoscopic sample updated.

paulhoux:
XCode project has not yet been updated.


richardeakin:
Merged - looking good on mac.  Please remember that we no longer add CinderApp.icns files to each sample as they are around half a meg each, instead all samples should reference the one at samples/data/CinderApp.icns (I made the change before pushing).

PR #645, opened 2014-12-23
User can specify GlslProg for use when rendering TextureFont.

sansumbrella:
Pass as `TextureFont::DrawOptions().glslProg( shader )`

Sample result:
screen shot 2014-12-23 at 5 40 26 pm

Also includes the bugfixes from #644


richardeakin:
Tested with TextureFont sample - looks good to merge! Thanks David.

PR #649, opened 2015-01-07
fix full screen in multiple monitors under windows

bgbotond:
In case of full screen the window always jumped to the first monitor even it was requested from another one.


andrewfb:
Thanks for this find & fix...

PR #655, opened 2015-01-14
LoggerFile: Create the file only if/when needed.

simongeilfus:
Two lines modification to address the “issue” explained here:
https://forum.libcinder.org/topic/using-the-new-log-feature-in-glnext#23
286000002245021

Thanks to std::ofstream::is_open, it is easy to check whether the file has already been created or not (http://www.cplusplus.com/reference/fstream/ofstream/is_open/).


richardeakin:
Thanks!

PR #658, opened 2015-01-15
Panning behaves correctly regardless of camera orientation.

num3ric:
Panning behaves correctly regardless of camera orientation.


num3ric:
Oh, are the billboard vectors guaranteed to be normalized? It's not immediately apparent by looking at the implementation.

num3ric:
To answer myself, yes they are unit vectors. The view matrix (3x3) is orthogonal.

andrewfb:
Nice fix.

PR #659, opened 2015-01-16
Added support for glDrawRangeElements.

paulhoux:
Allows you to draw only part of a vertex buffer, by specifying a start and end index (inclusive). I needed this for a particle system, where the buffer had room for 256 particles per emitter, but sometimes only a few were active. I did not want to recreate the buffer every time the number of particles changed. Now I can simply call ```mBatch->drawRange(0, mParticles.size() - 1);``` to draw all active particles, no more, no less.

This may also answer a question in this post:
https://forum.libcinder.org/#Topic/23286000002235007

Only tested this on Windows. The ```glDrawRangeElements()``` call is not supported on ES2, but you could use the non-indexed version ```glDrawArrays()``` instead.


richardeakin:
Nice one!  I found myself needing similar funcitonality on the last project I worked on - I wanted to draw portions of point sprite stars, each portion with different textures bound. I ended up having to use glDrawArrays() directly, but of course I would prefer if it was built into both Batch and VboMesh.

One suggestion: it seems to me that it would be closer to the OpenGL spec, and therefore more intuitive, to use start + count, instead of start + end.

andrewfb:
Paul and I talked a bit about this exact change, and also that `glDrawElements()` is probably the call we want rather than `glDrawRangeElements()`.

andrewfb:
I've added support for this with optional parameters for 'first' and 'count' on Batch::draw() as well as gl::draw( VboMesh ).

Issue #663, opened 2015-01-21
Add Blob arguments in OSC block


indefinit:
Feature request.  I'd like to send and receive blob arguments in the OSC Block.  Since osc/OscTypes.h already defines a Blob type, I feel like it should be relatively trivial to implement the arguments in the Block wrapper.  I've started working on it in my own fork but depending on general consensus (i.e. if others think they'd benefit from a blob data type in OSC) maybe I could present that here?  Thanks!

pizthewiz:
The OSC blob type is one of the four types required by the OSC 1.0 spec, so a PR sounds good to me.

Relatedly, I think there was some discussion of ASIO-based OSC handling on the forum and possibly even a CinderBlock, but I can't seem to find it.

pizthewiz:
The block I was thinking of might be [djmike/OscDev](https://github.com/djmike/OscDev), but it doesn't appear to be very active.

indefinit:
hmm interesting about the ASIO OSC handling.  

I'm happy to open up a PR.  One quick question though.  in the osc::ReceivedMessageArgument class I see two methods for receiving blobs, but I'm a little confused as to what to actually pass as parameters to these methods since the other ReceivedMessageArgument methods take no parameters.  

```
void ReceivedMessageArgument::AsBlob( const void*& data, osc_bundle_element_size_t& size ) const
{
    if( !typeTagPtr_ )
        throw MissingArgumentException();
	else if( *typeTagPtr_ == BLOB_TYPE_TAG )
		AsBlobUnchecked( data, size );
	else
		throw WrongArgumentTypeException();
}
void ReceivedMessageArgument::AsBlobUnchecked( const void*& data, osc_bundle_element_size_t& size ) const
{
    // read blob size as an unsigned int then validate
    osc_bundle_element_size_t sizeResult = (osc_bundle_element_size_t)ToUInt32( argumentPtr_ );
    if( !IsValidElementSizeValue(sizeResult) )
        throw MalformedMessageException("invalid blob size");

    size = sizeResult;
	data = (void*)(argumentPtr_+ osc::OSC_SIZEOF_INT32);
}
 ```

If I extend the ProcessMessage method in OscListener it would look something like:
```
else if(arg->IsBlob())
    // where do I reference data and generic bundle element size?
    message->addBlobArg(arg->AsBlobUnchecked(const void *&data, osc_bundle_element_size_t &size);
```

pizthewiz:
[...]

PR #664, opened 2015-01-21
Feature osc blob

indefinit:
This commit implements the osc blob type from oscpack, per Issue #663.


pizthewiz:
It looks like you merged glNext into your master branch in 583ffaf, a change I don't think we'd want to make quite yet :wink:.

indefinit:
Oy! not sure how that happened.  definitely don't want that...

PR #665, opened 2015-01-22
Feature: implements osc blob arg 

indefinit:
After a little git rebase hell, here is an attempt at an OSC blob argument.  Implements Issue #663 .

### DIFF
--- a/blocks/OSC/src/OscArg.h
+++ b/blocks/OSC/src/OscArg.h
@@ -107,28 +107,25 @@ typedef enum _ArgType
 	private:
 		std::string value;
 	};
-    
-    class ArgBlob : public Arg
-    {
-    public:
-        ArgBlob( ci::Buffer _value ) { value = _value; }
-        
-        /// return the type of this argument
-        ArgType getType() const { return TYPE_BLOB; }
-        std::string getTypeName() const { return "blob"; }
-        
-        /// return value
-        ci::Buffer get() const { return value; }
-        /// set value
-        void set( ci::Buffer _value ) { value = _value; }
-        
-        void set( const char *_value, unsigned int size) {
-            ci::Buffer buff = Buffer(&value, size);
-            value = buff;
-        }
-    private:
-        ci::Buffer value;
-    };
+	
+	class ArgBlob : public Arg
+	{
+	public:
+		ArgBlob( const ci::Buffer _value ) { value = _value; }
+		/// return the type of this argument
+		ArgType getType() const { return TYPE_BLOB; }
+		std::string getTypeName() const { return "blob"; }
+		/// return value
+		ci::Buffer get() const { return value; }
+		/// set value
+		void set( ci::Buffer _value ) { value = _value; }
+		void set( const char *_value, unsigned int size) {
+			ci::Buffer buff = Buffer(&value, size);
+			value = buff;
+		}
+	private:
+		ci::Buffer value;
+	};
 
-} // namespace osc
-} // namespace cinder
+}// namespace osc
+}// namespace cinder
--- a/blocks/OSC/src/OscArg.h
+++ b/blocks/OSC/src/OscArg.h
@@ -14,7 +14,7 @@
  * Redistributions in binary form must reproduce the above copyright notice, this list of conditions and
  the following disclaimer in the documentation and/or other materials provided with the distribution.
  
- THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED
+ THIS SOFTWARE IS PROVID&&ED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED
  WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
  PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
  ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
@@ -107,6 +107,28 @@ typedef enum _ArgType
 	private:
 		std::string value;
 	};
+    
+    class ArgBlob : public Arg
+    {
+    public:
+        ArgBlob( ci::Buffer _value ) { value = _value; }
+        
+        /// return the type of this argument
+        ArgType getType() const { return TYPE_BLOB; }
+        std::string getTypeName() const { return "blob"; }
+        
+        /// return value
+        ci::Buffer get() const { return value; }
+        /// set value
+        void set( ci::Buffer _value ) { value = _value; }
+        
+        void set( const char *_value, unsigned int size) {
+            ci::Buffer buff = Buffer(&value, size);
+            value = buff;
+        }
+    private:
+        ci::Buffer value;
+    };
 
 } // namespace osc
 } // namespace cinder
--- a/blocks/OSC/src/OscBundle.h
+++ b/blocks/OSC/src/OscBundle.h
@@ -32,7 +32,7 @@
 namespace cinder { namespace osc {
 
 class Bundle {
-  public:
+public:
 	Bundle();
 	~Bundle();
 	Bundle(const Bundle& other){copy(other);}
@@ -51,7 +51,7 @@ class Bundle {
 	const Bundle& getBundleAt(int index) const { return bundles[index]; }
 	const Message& getMessageAt(int index) const {return messages[index];}
 	
-  private:
+private:
 	std::vector<Message> messages;
 	std::vector<Bundle> bundles;
 };	
--- a/blocks/OSC/src/OscListener.cpp
+++ b/blocks/OSC/src/OscListener.cpp
@@ -132,14 +132,14 @@ void OscListener::ProcessMessage( const ::osc::ReceivedMessage &m, const IpEndpo
 			message->addFloatArg(arg->AsFloatUnchecked());
 		else if (arg->IsString())
 			message->addStringArg(arg->AsStringUnchecked());
-        else if(arg->IsBlob()){
-            const char *data;
-            ::osc::osc_bundle_element_size_t size = 0;
-            arg->AsBlobUnchecked((const void *&) data, size);
-            ci::Buffer buffer(&data, size);
-            message->addBlobArg(buffer);
-        }
-        else {
+		else if(arg->IsBlob()){
+			const char *data;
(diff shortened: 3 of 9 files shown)


pizthewiz:
Nice, I wasn't sure if it would be recoverable via `rebase`!

While Rich and Andrew will no doubt have feedback and I haven't actually pulled the PR down to try it out, I thought I'd quickly mention the indentation seems a little off, I think you are using spaces while [the project uses tabs](https://github.com/cinder/Cinder/blob/master/CONTRIBUTING.md#spacing) - it isn't a showstopper, but can make code a bit more difficult to read.

indefinit:
ok cool makes sense about the tabs.  I'll wait to hear more feedback and then change to tabs as well.

richardeakin:
I can't comment too much from an experienced OSC perspective, but I agree with piz that if it is in the spec then we should expose something like this.  Can you provide some test code? Perhaps just some snippets that we can add to OscSender / OscListener samples.

Regarding style:  the OSC block was written before there was any dicreed cinder style, which will be redone as a whole at some point. For now its probably best just to match the surrounding style of the code, and it looks like you've done well, once you switch to tabs instead of spaces.

pizthewiz:
Looks like the includes changed from glNext to master as `ci::Buffer` is undefined in _OscMessage.h_ without an explicit include:
```diff
diff --git a/blocks/OSC/src/OscMessage.h b/blocks/OSC/src/OscMessage.h
index 1f42ee7..f8de5fd 100644
--- a/blocks/OSC/src/OscMessage.h
+++ b/blocks/OSC/src/OscMessage.h
@@ -27,6 +27,7 @@
 #pragma once
 
 #include "cinder/Cinder.h"
+#include "cinder/Buffer.h"
 #include "cinder/Exception.h"
 
 #include "OscArg.h"
```

indefinit:
Oh boy!  So sorry for the delay on these changes!  Didn't mean to delay the changes this long but I got stuck on the tests and then had to shift gears to a different installation...

Anyway, full disclosure on these tests, I think my lack of understanding for the IStream and OStream implementation in Cinder might be getting in the way since I haven't been able to produce a successful test.

Test 1 OscBlobListener: https://gist.github.com/indefinit/132fc6084e734b490d6c
Test 2 OscBlobSender: https://gist.github.com/indefinit/354f0c956865cbaec3bc

Osc Listener Blob seems to work when I send blobs from Processing but I must be doing something wrong when I blob send from Cinder.  On the listener test app, I create a IStreamMemRef from the blob data, which seems to work when sending byte arrays from Processing but I get SIGBART errors from Cinder when I try to send buffer data as a blob.  

Any help or further critique you might have are much appreciated!  Thanks!

daan:
The blob sending needs to be updated to make things work. see example in oscpack, OscSendTests.cpp ):

in  OscSender.cpp, void OscSender::appendMessage( const Message& message, ::osc::OutboundPacketStream& p ){

just put the "data" in a oscpack "Blob"

}else if (message.getArgType(i) == TYPE_BLOB){
    p << ::osc::Blob( message.getArgAsBlob(i).getData(),  message.getArgAsBlob(i).getDataSize() );
}

thank you for adding blob support, i was waiting for this a long time.

pizthewiz:
It'll probably need an additional tweak once `ci::Buffer` migrates to an explicit `shared_ptr` type as well, see #763.

richardeakin:
It's probably worth mentioning, that a new OSC implementation is in the works, although there is no ETA on that yet. So, if you folks need this and are willing to get this feature working across the board, we can look at it getting included for 0.9.0 release, though we can't commit much time to testing it ourselves.

@pizthewize where are thinking the `ci::Buffer` tweak would be needed?

daan:
I think it is a nice feature. I tested Test2 OscBlobSender with pyOsc and receive data correctly on the other end. I will test some more next week when I implement it in my current project ( both windows and mac)

pizthewiz:
@richardeakin I was thinking `ci::Buffer` -=> `ci::BufferRef` but you are right that it wouldn't be required. If you guys are looking for another set of eyes on the new OSC implementation, I'd love to help out - I have a bit of OSC experience after writing the Objective-C lib [PonyExpress](https://github.com/pizthewiz/PonyExpress) and a SLIP encoder/decoder to allow OSC over serial in OF.

indefinit:
I'm still interested in this as well, but I default to Piz who has a lot
more experience in this implementation than I do.  For my installation I
ended up using a workaround which didn't need to rely on OSC so I don't
have an immediate need for blobs in OSC right now, but I still think it's
super important/cool.

On Thu, Apr 30, 2015 at 12:36 PM, pizthewiz <notifications@github.com>
wrote:

> @richardeakin <https://github.com/richardeakin> I was thinking ci::Buffer
> -=> ci::BufferRef but you are right that it wouldn't be required. If you
> guys are looking for another set of eyes on the new OSC implementation, I'd
> love to help out - I have a bit of OSC experience after writing the
> Objective-C lib PonyExpress <https://github.com/pizthewiz/PonyExpress>
> and a SLIP encoder/decoder to allow OSC over serial in OF.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cinder/Cinder/pull/665#issuecomment-97872044>.
>



-- 

Kevin Siwoff . data + sound + vision
// http://kevinsiwoff.com // http://studioindefinit.com

pizthewiz:
I personally haven't needed to use OSC data blobs in Cinder yet, but I'd be happy to help out - I'll post some comments in the changes.

blocks/OSC/src/osc/OscTypes.h
```diff
@@ -47,7 +47,7 @@ typedef std::int64_t	int64;
 typedef std::uint64_t	uint64;
 typedef std::int32_t	int32;
 typedef std::uint32_t	uint32;
-
+typedef std::uint8_t    uint8;
```
pizthewiz: It doesn't look like this new `typedef` is use and if so, it might be better to leave the file unchanged.

blocks/OSC/src/osc/OscReceivedElements.h
```diff
@@ -261,7 +261,7 @@ class ReceivedMessageArgument{
     bool IsSymbol() const { return *typeTagPtr_ == SYMBOL_TYPE_TAG; }
     const char* AsSymbol() const;
     const char* AsSymbolUnchecked() const { return argumentPtr_; }
-
+
```
pizthewiz: Probably not worth changing either ;0)

blocks/OSC/src/OscArg.h
```diff
@@ -14,7 +14,7 @@
  * Redistributions in binary form must reproduce the above copyright notice, this list of conditions and
  the following disclaimer in the documentation and/or other materials provided with the distribution.
  
- THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED
+ THIS SOFTWARE IS PROVID&&ED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED
```
pizthewiz: Probably just a stray insertion cursor

blocks/OSC/src/OscBundle.h
```diff
@@ -32,7 +32,7 @@
 namespace cinder { namespace osc {
 
 class Bundle {
-  public:
+public:
```
pizthewiz: Probably not worth only making indentation changes to this file.

blocks/OSC/src/osc/OscReceivedElements.cpp
```diff
@@ -444,9 +444,10 @@ void ReceivedMessageArgument::AsBlobUnchecked( const void*& data, osc_bundle_ele
     osc_bundle_element_size_t sizeResult = (osc_bundle_element_size_t)ToUInt32( argumentPtr_ );
     if( !IsValidElementSizeValue(sizeResult) )
         throw MalformedMessageException("invalid blob size");
-
+
```
pizthewiz: Probably not worth making just indentation changes to this file too.

blocks/OSC/src/OscListener.cpp
```diff
@@ -132,6 +132,13 @@ void OscListener::ProcessMessage( const ::osc::ReceivedMessage &m, const IpEndpo
 			message->addFloatArg(arg->AsFloatUnchecked());
 		else if (arg->IsString())
 			message->addStringArg(arg->AsStringUnchecked());
+		else if(arg->IsBlob()){
```
pizthewiz: This file seems to have inconsistent style but we might as well be consistent with the `for` loop body and add a space between `else if` and `(arg-> …`

blocks/OSC/src/OscMessage.h
```diff
@@ -58,11 +58,13 @@ namespace cinder { namespace osc {
 		int32_t getArgAsInt32( int index, bool typeConvert = false ) const;
 		float getArgAsFloat( int index, bool typeConvert = false ) const;
 		std::string getArgAsString( int index, bool typeConvert = false ) const;
-		
+		ci::Buffer getArgAsBlob(int index, bool typeConvert = false) const;
```
pizthewiz: Need the one space padding within the parens `ci::Buffer getArgAsBlob( int index, bool typeConvert = false )`

blocks/OSC/src/OscMessage.cpp
```diff
@@ -77,26 +77,34 @@ float Message::getArgAsFloat(int index, bool typeConvert) const{
 		else
 			throw OscExcInvalidArgumentType();
 	}else
-        return ((ArgFloat*)args[index])->get();
+		return ((ArgFloat*)args[index])->get();
 }
 
 std::string Message::getArgAsString( int index, bool typeConvert ) const{
-    if (getArgType(index) != TYPE_STRING ){
```
pizthewiz: Probably want to revert the indentation changes to non-blob stuff.

pizthewiz:
Syntactic knitpicks, I'll test out some interoperability too.

pizthewiz:
In trying this out, it looks like a message sent from Cinder with a blob is not encoded properly, it is represented as `T` in the message's type tag which indicates the arg-less `True` type instead of `b` - in a quick look, I don't see how the Cinder OSC block is mapping its message arguments into oscpack…

pizthewiz:
Looks like the sender uses oscpack's `OscOutboundPacketStream` but for the blob type it is sending the backing buffer address:
```C++
p << message.getArgAsBlob(i).getData();
```

while it probably needs to use this:
```C
OutboundPacketStream& operator<<( const Blob& rhs );
```

daan:
hi,
that was also my conclusion, the solution is simple ( see my previous comment )

in OscSender.cpp, void OscSender::appendMessage( const Message& message, ::osc::OutboundPacketStream& p ){

just put the "data" in a oscpack "Blob"
 
    }else if (message.getArgType(i) == TYPE_BLOB){
    p << ::osc::Blob( message.getArgAsBlob(i).getData(), message.getArgAsBlob(i).getDataSize() );
    }

pizthewiz:
For what it is worth, the blob type is also implemented in @ryanbartley's new OSC block PR'ed at #1179.

richardeakin:
Yes how about we close this as it is superseded by Ryan's new implementation.

PR #666, opened 2015-01-22
Update git clone guidance in Git Setup Guide, fixes #489

### LINKED ISSUE #489: Git install guide uses nonexistent branch
The [Git install guide](http://libcinder.org/docs/welcome/GitSetup.html) instructs users to clone the dev branch, which no longer exists on Github.

pizthewiz:
Fixes #489

Update the provided `git clone` commands and a few other small elements. There are several other things on this page that could use updating, but since this doc is linked off the main README, I thought the cloning bit at least should be accurate until a more substantial overhaul.

Other stuff that needs tweaking:
- Screenshots
- Git Installers
- VC Version Support

### DIFF
--- a/docs/welcome/GitSetup.html
+++ b/docs/welcome/GitSetup.html
@@ -39,7 +39,7 @@ $(function() {
 
 	<div class="contents">
 		<h1>Cinder + Git</h1>
-		<p>If you're interested in keeping up with the latest Cinder development you'll want to setup a clone of Cinder's git repository. Cinder's home on Github is at <a href="http://github.com/cinder">http://github.com/cinder</a>.</p>
+		<p>If you're interested in keeping up with the latest Cinder development you'll want to setup a clone of Cinder's git repository. Cinder's home on GitHub is at <a href="http://github.com/cinder">http://github.com/cinder</a>.</p>
 		<br>
 		<p>In general Cinder's active development is done in branches and the <em>master</em> branch is designed to reflect minor updates to the latest official release.</p>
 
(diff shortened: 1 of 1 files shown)


richardeakin:
Overrall these look like nice improvements, thanks for this, though one comment:

I think we should leave the SSH repo links - I don't think we should promote people using http protocal for git checkouts. This forces you to have to suppy a username and password on each git remote command, which can be extremely annoying.  I know that every time I happen upon a dev who has setup his repos with http, the first thing I instruct them to do is switch it to git or ssh protocal, setting up sha keys if necessary.

pizthewiz:
Oh weird, I thought it was the other way around, that SSH links require a user/pass?

richardeakin:
No - it requires setting up ssh keys instead for many operations. I routinely refer people to [this link](https://help.github.com/articles/generating-ssh-keys/) to do that (considering the target audience of cinder, probably not necessary or even belittling to include in the doc).

pizthewiz:
My error, that was exactly what I was trying to avoid - I'll revert 6906ef0.

pizthewiz:
It should be all sorted now, I went the evil route and rebased to remove the commit instead of reverting it.

richardeakin:
Looks good, thanks for the updates!

sansumbrella:
I think sharing the link for ssh keygen is always welcome. Anyone who sets up ssh-keys every day can ignore it, otherwise it's the kind of thing that's easy to forget (and for most, hardly worth remembering).

sansumbrella:
Also, glad to see the docs getting some attention. Thanks!

Issue #667, opened 2015-01-23
Enable comments support in JsonTree by default, add ParseOption to disable 


richardeakin:
After further investigation into json and comments following #643, I discovered that the reason we can't parse some comments is that they are explicitly disabled in JsonTree (see [here](https://github.com/cinder/Cinder/blob/master/src/cinder/Json.cpp#L270)).

By default, jsconcpp aims to support comments, I think we should too because it is surprising to many users, not all of whom are technical, when they write a comment in a json file and it blows up on them. To disable comment support, we can add another [ParseOption](https://github.com/cinder/Cinder/blob/master/include/cinder/Json.h#L55).

PR #668, opened 2015-01-23
Upgrade jsoncpp to release 1.3.0

### LINKED ISSUE #643: update jsoncpp
Comments in json can be very useful, and jsoncpp is one of the few json libraries that supports them (even though not part of the spec). It seems that there is a couple bugs in the version of jsoncpp we used, though I believe they are fixed in its master branch (see [this commit](https://github.com/open-source-parsers/jsoncpp/commit/fd06bfca792d327c9fc990bc1052c1bd0cbf7836)).

richardeakin:
__TODO: test on Visual Studio 2012__
This version of jsoncpp now requires C++11, and we need to make sure that VS 2012 can build it without problems before merging into master. If it doesn't, I will rebase against glNext (where we use VS2013 as a minimum on windows). 

##### Steps taken to upgrade:
checkout out 1.3.0 tag from: https://github.com/open-source-parsers/jsoncpp
and ran:
```
cd jsoncpp
python amalgamate.py
```

Using the resulting amalgamated .h and .cpp instead.  Added some code to JsonTest around comments (not yet passing because comments are disabled in JsonTree, see #667).

Fixes #643.


richardeakin:
I've just pushed project file updates for Visual Studio 2013, everything is OK there. I can't do vc2012 / vc2012_winrt, hoping someone else can.. these were the steps:

- remove old jsoncpp headers and cpp's from project file (there are 6 or 7 each)
- add new, amalgamated header and source
- add a header include path for `..\include\jsoncpp` for all targets

andrewfb:
Just tested this with VC2012 and everything appeared correct. JsonTest behaved as expected, failing in testComments().

andrewfb:
Just logging that I also merged these changes into glNext.

PR #675, opened 2015-01-25
mallocScoped() utility

richardeakin:
I stumbled upon [this article](http://www.codeproject.com/Articles/820931/Using-std-unique-ptr-RAII-with-malloc-and-free) not too long ago, which shows a nice way to make a scoped malloc. It is a bit long winded but an improvement over using a shared_ptr() in times where you just want to make a local chunk of data and ensure that it gets freed at the end of scope.

So, today I made this utlity function to simplify matters. Here is an example where I'm using it in Cinder-Dart ([this commit](https://github.com/richardeakin/Cinder-Dart/commit/49cd87d5b0ddc62d848f4f8a619cb8fb0a11e787):

This:
```c++
const char **vmCFlags = (const char **)malloc( mVMFlags.size() * sizeof( const char * ) );
for( size_t i = 0; i < mVMFlags.size(); i++ )
  vmCFlags[i] = mVMFlags[i].c_str();

Dart_SetVMFlags( mVMFlags.size(), vmCFlags );
free( vmCFlags );
```

turns into:
```c++
auto vmCFlags = mallocScoped<const char *>( mVMFlags.size() );

for( size_t i = 0; i < mVMFlags.size(); i++ )
   vmCFlags.get()[i] = mVMFlags[i].c_str();

Dart_SetVMFlags( mVMFlags.size(), vmCFlags.get() );
```

Making the unique_ptr without the addition of `mallocScoped` util would look like:
```c++
auto vmCFlags = unique_ptr<const char *, decltype( &free )>( (const char **)malloc( vmFlagsBytes ), free );
```

Which is a bit hard to remember, IMO.

There are several places in cinder's codebase where I think we can make use of this util to improve the code a bit.

### DIFF
--- a/include/cinder/Utilities.h
+++ b/include/cinder/Utilities.h
@@ -93,6 +93,13 @@ inline double fromString( const std::string &s ) { return atof( s.c_str() ); }
 //! Returns a stack trace (aka backtrace) where \c stackTrace()[0] == caller, \c stackTrace()[1] == caller's parent, etc
 std::vector<std::string> stackTrace();
 
+//! Mallocs `count * sizeof( T )` bytes and returns it in a std::unique_ptr so that `free` will be called when it goes out of scope.
+template <typename T>
+std::unique_ptr<T, decltype( &free )> mallocScoped( size_t count )
+{
+	return unique_ptr<T, decltype( &free )>( (T*)malloc( count * sizeof( T ) ), free );
+}
+
 // ENDIANNESS
 inline int8_t	swapEndian( int8_t val ) { return val; }
 inline uint8_t	swapEndian( uint8_t val ) { return val; }


sansumbrella:
Is there a reason to prefer this to `unique_ptr`?

sansumbrella:
But as a malloc-free alternative, how about:
```c++
auto vmCFlags = unique_ptr<char[]>( new char[mVmFlags.size] );
```

richardeakin:
I've seen andrew use the above, but yes the problem is this doesn't cover things that must be malloc'ed, which we use quite a few times when interfacing with C-apis  (win32 apis and core audio are two that come to mind). 

Also, last time I checked new doesn't provide any aligment help, although many implementations of malloc allocate things on 16byte boundaries so the compiler can optimize a little easier (this could be total BS, I haven't really investigated it in detail).

sansumbrella:
I think this could be a handy thing to have (though it is funnily akin to `salloc`). Maybe bring it into the Cinder naming world as `scopedMalloc`?

I don't understand why data would _need_ to be `malloc`'d if the C library isn't in charge of freeing it, though that isn't to say it wouldn't happen. I thought that new/delete and malloc/free shared the same heap, even though they affect it differently. Does it have something to do with destructors causing problems for the library memory (event no-op ones)?

richardeakin:
Hm, never heard of salloc. There may or may not be any concrete need to use malloc + free instead of new[] + delete, I haven't looked in detail, but I do know that malloc is used within the cinder codebase quite frequently, and relies on a separate call to free. Adding this would at minimum make it very easy to update those calls to malloc to be more leak-safe.

Seems like there aren't any objections, so I'm going to merge this soon.

richardeakin:
Ah, I forgot to reply about the suggestion to name it `scopedMalloc`: my thinking was that you'd get nice autocompletion with mallocScoped over plain malloc, though I'm open to the alternative. This case is a tad different to the existing Scoped* classes though, in that it returns a unique_ptr that you must catch in order to keep the data in scope. Either one works for me.

richardeakin:
Discussed with Andrew and we talked about making some adjustments to this, specifically around memory aligment and also about specifying bytes instead of count in the size argument. More work than I have time for at the moment, so closing for now.

PR #678, opened 2015-01-29
Fix for memory leak when quitting.

paulhoux:
When calling ```quit()```, the application's windows were not properly closed and destroyed, causing memory leak detectors to go nuts on application exit. To fix this, the ```quit()``` method now simply closes all open windows, forcing the application to quit as soon as the last window is closed.

I've implemented and tested this code only on Windows.

### DIFF
--- a/include/cinder/app/AppImplMswBasic.h
+++ b/include/cinder/app/AppImplMswBasic.h
@@ -43,7 +43,7 @@ class AppImplMswBasic : public AppImplMsw {
 
 	class AppBasic*		getApp() { return mApp; }
 	
-	void	quit() { mShouldQuit = true; }
+	void	quit();
 
 	float	setFrameRate( float aFrameRate );
 	void	disableFrameRate();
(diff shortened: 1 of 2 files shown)


richardeakin:
We have sort of been gaining this policy that we don't necessarily consider leaks at application shutdown to actually be leaks, mainly because freeing memory at that time can slow down the shutdown process.  That said, we still need to be able to find actual memory leaks, and many tools for this detect false positives at shutdown.

What about an application setting like `isShutdownMemoryLeaked()` which is off by default, but can be turned on when checking for leaks?  I can add this as part of the [app_refactor](https://github.com/richardeakin/Cinder/tree/app_refactor) changes, as I'm already making quite a few mods to App::Settings there.

paulhoux:
I see your point, but I still don't like the idea that a ```quit()``` takes a different code path as opposed to closing the last window. It could lead to a crash that occurs only when using ```quit()```.

Having said that, I just ran into an issue caused by this pull request: because all windows are actively closed, any instance that relies on the window being still present when ```shutdown()``` is called will/can now crash. I had to connect to the window's close signal to get rid of the crash.

richardeakin:
The consistency made sense. Also, we discussed offline that any sort of 'fastShutdown' parameter would be non-default and user opts into it, that by default we should be freeing memory to ensure we are leak free.

I think it makes sense that you can't rely on there being any Windows left when shutting down.

In other words, looks good for merging. :)

PR #699, opened 2015-02-06
Missing implementation of Color / "std::ostream& operator<<"

simongeilfus:
Apparently this one has been left out when updating the Color class. There's the function signature for the 4 color types, but only the implementation for ColorA<uint8_t> and ColorA<float>.

This PR fixes this.


simongeilfus:
ps: This means that any code that has a toString( color ) will not link because of missing symbols.

andrewfb:
Thanks for this...

PR #701, opened 2015-02-10
Json tree additions

ryanbartley:
Helpers to making arrays easier. 
- Added Variadic Constructor and Variadic static method.
- Also added initializer_list version of the static method.

### DIFF
--- a/include/cinder/Json.h
+++ b/include/cinder/Json.h
@@ -263,6 +263,17 @@ private:
 
 	explicit JsonTree( const std::string &key, const Json::Value &value );
 	
+	template<typename Arg>
+	static void pushType( JsonTree & parent, const Arg &argument, std::true_type )
+	{
+		parent.pushBack( argument );
+	}
+	template<typename Arg>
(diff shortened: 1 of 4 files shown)


paulhoux:
IMHO, I think we should remove ```JsonTree``` altogether and simply use jsoncpp's own interface. Adding and finding nodes is done using ```operator[]``` overloads, which is way easier than using iterators.

pizthewiz:
:+1: I am a fan of [JsonCpp](https://github.com/open-source-parsers/jsoncpp) as well. Out of scope for Cinder-0.9.0?

richardeakin:
You can use it directly now (we do for our configuration projects on our current project). I'm thinking to move the header to cinder/include/jsoncpp/json.h though so it is slightly easier (currently it lives at cinder/include/jsoncpp/__json__/json.h).

richardeakin:
Going to close this since I think we're in agreement that we don't want to complicate JsonTree any further, but instead start working with jsoncpp directly. Also with #935, I've made the header change so you now can use `#include "jsoncpp/json.h"`.

PR #704, opened 2015-02-10
Const-correctness for svg::findByIdContains.

sansumbrella:
Enables use of findByIdContains() with nodes that were found using any of the svg::find* methods.


sansumbrella:
(correcting my mistake from 2012)

PR #706, opened 2015-02-12
PolyLine area and centroid

### LINKED ISSUE #679: Compute area of a PolyLine?
I needed to compare the area of some shapes defined using PolyLines. I didn't see an obvious way to compute the area so I wrote this function up: 
```
// Assumes line does not self intersecting
// Treats the line as closed (you wanted area right?)
// http://www.mathsisfun.com/geometry/area-irregular-polygons.html
float area( PolyLine2f line ) {
	float sum = 0.0;
	if ( line.size() > 1 ) {
		PolyLine2f::iterator it;
		Vec2f prev, curr;
		for ( it = line.begin(), prev = *it++; it != line.end() ; it++ ) {
			curr = *it;
			sum += ( prev.y + curr.y ) / 2.0 * ( curr.x - prev.x );
			prev = curr;
		}
	}
	return sum;
}
```
https://gist.github.com/drewish/d5d95cdf398e5f7e7240

If there's any interested I'd be happy to try to put together a PR to clean this up and add it as method of the class

drewish:
This exposes boost::geometry's [area](http://www.boost.org/doc/libs/1_57_0/libs/geometry/doc/html/geometry/reference/algorithms/area/area_1.html) and [centroid](http://www.boost.org/doc/libs/1_57_0/libs/geometry/doc/html/geometry/reference/algorithms/centroid/return_centroid_1.html) algorithms

This fixes #679.

### DIFF
--- a/src/cinder/PolyLine.cpp
+++ b/src/cinder/PolyLine.cpp
@@ -173,6 +173,14 @@ double PolyLine<T>::area() const
 }
 
 template<typename T>
+T PolyLine<T>::centroid() const
+{
+	point p = boost::geometry::return_centroid<point>( convertPolyLineToBoostGeometry ( *this ) );
+	return T( p.x(), p.y() );
+}
+
+
+template<typename T>
 std::vector<PolyLine<T> > PolyLine<T>::calcUnion( const std::vector<PolyLine<T> > &a, std::vector<PolyLine<T> > &b )
 {
 	if( a.empty() )
--- a/src/cinder/PolyLine.cpp
+++ b/src/cinder/PolyLine.cpp
@@ -157,8 +157,6 @@ polygon convertPolyLinesToBoostGeometry( const std::vector<PolyLine<T> > &a )
 template<typename T>
 std::vector<PolyLine<T> > PolyLine<T>::calcUnion( const std::vector<PolyLine<T> > &a, std::vector<PolyLine<T> > &b )
 {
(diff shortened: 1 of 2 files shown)


src/cinder/PolyLine.cpp
```diff
@@ -157,8 +168,6 @@ polygon convertPolyLinesToBoostGeometry( const std::vector<PolyLine<T> > &a )
 template<typename T>
 std::vector<PolyLine<T> > PolyLine<T>::calcUnion( const std::vector<PolyLine<T> > &a, std::vector<PolyLine<T> > &b )
 {
-	typedef boost::geometry::model::polygon<boost::geometry::model::d2::point_xy<double> > polygon;
```
drewish: These can just use the typedef in the anonymous namespace where the conversion functions are defined.

src/cinder/PolyLine.cpp
```diff
@@ -107,6 +107,17 @@ bool PolyLine<T>::contains( const Vec2f &pt ) const
 	return (crossings & 1) == 1;
 }
 
+template<typename T>
+double PolyLine<T>::area() const
+{
+	typedef boost::geometry::model::polygon<boost::geometry::model::d2::point_xy<double> > polygon;
+
```
drewish: I realize it might be good to explicitly check for an empty polyline and just return 0 here.

andrewfb:
OK - in light of the move away from Boost, I've merged an implementation similar to what you proposed at #679 ccdb941e0b2bc25b6bca89a0fb7582e4d5ae1707

Issue #694, opened 2015-02-03
Toggling to bordeless and back on mac has no resize buttons


richardeakin:
First reported on [forum here](https://forum.libcinder.org/topic/glnext-multi-window-app-framerate-drop#23286000002266019), repro'd with [this gist](https://gist.githubusercontent.com/simongeilfus/db9dee222d67cceed285/raw/a6a1741a878610e8543998ad7dcb16b28711535a/gistfile1.cpp). Hit b twice and you have no resize buttons.

lithium-snepo:
If a window title is set, it also disappears when returning from borderless mode. I'm able to get the buttons to return by manually resizing the window, or programmatically by calling `[mWin setFrame]` to a different size and then back again. Unfortunately this results in an ugly pop, so it's not a good solution. mWin also appears to deallocate its title string upon entering borderless mode, so you can't just call `[self setTitle:[self getTitle]]` to get it to come back, either. Long story short, boo at yosemite.

PR #709, opened 2015-02-14
Fixes missing titlebar buttons when exiting borderless mode

lithium-snepo:
Addresses issue reported at https://github.com/cinder/Cinder/issues/694 (excluding window title problem)


richardeakin:
Thanks for this. We're in the middle of a hefty app refactoring pass (#710), but will try to get your fix in there once that goes through.

richardeakin:
I moved your fix over to app_refactor branch #734, since many things in app:: changed there. Thanks again!

Issue #647, opened 2014-12-31
Assets folder: sometimes, searching 5 folders up is not enough


richardeakin:
(In reference to [this method](https://github.com/cinder/Cinder/blob/master/src/cinder/app/App.cpp#L210-L262)).

Sometimes, and I'm not really sure when, xcode decides to bury the app it builds even deeper than 5 folders in. I just hit this again when trying to add an asset to JsonTest and it couldn't find the assets folder in the usual place, sitting next to the xcode folder.

I'd like to propose two changes around this:
1) Increase the default search to 7, which seems to cover this xcode weirdness
2) make the depth settable in App::Settings. This way at least the user can extend the depth without modifying cinder itself.

There's some heavy app refactoring underway (paused, but about to resume), I'd be happy to address the above in that branch as the prepareAssetsDirectory will be moving to platform specific files..

richardeakin:
Update: Andrew and I discussed that we should just increase the depth to 10, less user maintanence that way

richardeakin:
Fixed in app_refactor branch, #710

Issue #619, opened 2014-11-28
Replace boost::signals2 with C++11 implementation


richardeakin:
boost::signals2 is probably our largest dependancy on boost (other than filesystem and asio, which are both proposed for standardization), and it is also quite beefy in the realm of signals / slots implementations. Swapping it out is a breaking change, although only minor, in that method and class names will slightly change (ex. `ci::signals::signal` ->  `ci::signals::Signal`).

Right now the favorite replacement candidate is a modified version of [SimpleSignal](https://testbit.eu/cpp11-signal-system-performance/), modifed in the sense that we still need to make up for a couple features. Specifically it doesn't support connections or scoped connections, and I'm not sure if its 'collector' notion equates to boost::signal2's 'combiner' notion.  I have been using a modified version of SimpleSignal in private for some months now, and plan to implement these features before we make the switch.

**Important: boost::signals2 is thread safe, SimpleSignal is not**. This is not a problem for any of cinder's internal usage of signals, and for the most part signals are used from UI threads so it isn't a problem there either. However, if it is a problem for someone else, first please reconsider your design :), then let's discuss if this is actually a requirement for a ci::Signal implementation. Most likely, we would recommend that you keep using boost::signals2 directly, but we do want to know of any use cases that users could have and we don't know about.

Related to #611 
Depends on #618 being accepted.

richardeakin:
Fixed in app_refactor branch, #710

Issue #611, opened 2014-11-26
Move signal related stuff to cinder/Signals.h


richardeakin:
This is a breaking change, and how things will be organized once we are using a non-boost signals implementation. Will also move the EventCombiners here from  [App.h](https://github.com/cinder/Cinder/blob/master/include/cinder/app/App.h#L80-L125).

richardeakin:
Fixed in app_refactor branch, #710

Issue #610, opened 2014-11-24
Should app::loadResource() be taking a fs::path() for macPath?


richardeakin:
Just checking if there are any esoteric reasons why [loadResource( std::string macPath )](https://github.com/cinder/Cinder/blob/master/include/cinder/app/App.h#L376) couldn't handle an fs::path. I don't *think* it would break anyone's code to change it to fs::path, but I'm not entirely sure either..

Anyway, I'm looking at refactoring this part of the codebase right now, so I'd like to change that to:

static DataSourceRef loadResource( const fs::path &resourcePath );

And define it everywhere, since posix and winrt can make use of it too (though it would throw on win32).

richardeakin:
Fixed in app_refactor branch, #710

PR #710, opened 2015-02-14
app:: refactor

richardeakin:
We've done a lot in here, mostly around breaking changes that we have planned to get done before 0.9.0 release. It makes sense to merge this back into glNext before continuing on, as it is ready to be battle tested. Some of the main features addressed are:

- Refactored platform-specific code in App and AppBasic into subclasses to make it easier to maintain additional platforms in the future.
- In the spirit of heading towards a more cross-platform API, user's apps simply inherit from `App`, which is typedefed to the correct platform occurding to preprocessor directives. The base App class is now called `AppBase`, which is pure abstract.
- You now have the ability to use OpenGL functionality or assets from your app's constructor or member variables, which used to crash and was quite mysterious to new users.  To make this happen, we removed the virtual `prepareSettings()` methods, instead you pass in an optional `std::function` to the app instantiation macro (see #602 for details).
- The naming of app instantiation macros has been simplified - now you just use
         `CINDER_APP( ... )`.
- Nearly all of the platform-specific headers have been tucked away in implementations, significantly reducing the precompiled size of a basic app.
- A virtual app::Platform class has been added, along with PlatformCocoa and PlatformMsw implementations, which is to move things out of App that can be used in a non-app environment, such as a plug-in.  This will see more additions in the future.
- a new C++11 Signals implementation, replacing boost::signals2. It is based on SimpleSignal, though it has been heavily modified to support features we find useful, such as scoped connections and slot ordering. Not only is it much faster, it is also much smaller in code size, totalling ~600 lines and using nothing but standard headers. See #619 for details.
- various small changes to App::Settings to try to make the naming scheme more predictable. Now, you'd always start with `set*` to update a variable.
- fixes `MouseEvent::getPos()` during `mouseWheel()` (#259)

Aside from being a nice step towards more platforms, and organizing the ever-evolving app:: namespace while we have the chance, we are seeing some nice compile time performance improvements, some reports say 20% faster.

Fixes issues: #647, #619, #602, #604, #610, #611, #259


num3ric:
Please add an ivec2 variant to the free-standing app::setWindowSize() in AppBase.h. Thanks!

PR #712, opened 2015-02-17
Added log appending, daily rotating log, moved NSLog to syslog

lucasvickers:
Generally I like the straight forward logging technique.  I think we can get enough basic functionality out of it, and if we expose the logs to syslog and eventlog we can tie into more advanced reporting systems for production rollouts.

##### I've added a few features:
* Added log appending and set to true as default.  
  * I'm open to changing the default to false, but for a rollout with an auto-restart tool we'd want to append to the old logs in case of crash, etc.
* Added a daily rotating log.  
  * I gave an overloaded ```enableFileLogging``` function to keep the interface simple, and the user must provide a ```strftime``` formatted string for the log file.  There is no verification on this, and the log rolls daily independent of the ```strftime``` format.
  * Perhaps future control of log rolling granularity (weeks, days, hours) could be nice.
* I wrote a ```LoggerSysLog``` class that writes to the syslog.  
  * I let the user change the SystemLoggingLevel, but standard system /etc/asl.conf configurations will ignore anything below Warning.  This differs slightly from what ```NSLog``` was doing (all levels in the system.log).  This is normal behavior, and syslog shouldn't be for debug information unless you have a reason and specifically configure it.
  * In ```LoggerSysLog::LoggerSysLog()``` I get the logging name from ```NSFileManager```.  There may be a more cinder oriented way to get a logging name (app name of sorts).

##### Example usages:
```
log::manager()->enableFileLogging("/tmp", "loggingTests.%Y.%m.%d.log", false);
OR
log::manager()->enableFileLogging("/tmp/cinder.log", true);

log::manager()->enableSystemLogging();
log::manager()->setSystemLoggingLevel(log::LEVEL_ERROR);
```
##### What I would like to add later:
* Windows EventLog for system logging in windows.
* I do conversion from the ```ci::log::LEVEL``` enums to ```syslog::LEVEL``` in a few places, I'd like to combine these somehow.


richardeakin:
Hey Lucas, these are some nice features, thanks for this work. I had hoped to switch to syslog at some point - do we need `LoggerNSLog()` anymore?

About the `syslog::LEVEL`, are you referring to `syslog()`'s priority? I think it should be enough to just have two private utility methods that convert between ci::log::Level and the int priority that would mean the same thing to syslog(), or am I missing something?

One thing I'd like to suggest for future PR's is that you try to separate the different features / types of modifications into different commits, as it is easier to review and make sense of the changes this way. Also try to follow the [styleguide](https://github.com/cinder/Cinder/blob/master/CONTRIBUTING.md#style-guide) as best as you can so the code looks uniform. Overall it looks like you've done a pretty good job, though I see some spacing differences here and there.

I will make a couple other suggestions in-line..

include/cinder/Log.h
```diff
@@ -199,23 +213,37 @@ class LogManager {
...
+															: disableFileLogging(); }
+
 	bool isFileLoggingEnabled() const					{ return mFileLoggingEnabled; }
 
 	void enableSystemLogging();
 	void disableSystemLogging();
 	void setSystemLoggingEnabled( bool b = true )		{ b ? enableSystemLogging() : disableSystemLogging(); }
 	bool isSystemLoggingEnabled() const					{ return mSystemLoggingEnabled; }
+	void setSystemLoggingLevel( Level level );
+	Level getSystemLoggingLevel()						{ return mSystemLoggingLevel; }
```
richardeakin: `getSystemLoggingLevel()` should be const

src/cinder/Log.cpp
```diff
@@ -318,6 +409,44 @@ void LoggerNSLog::write( const Metadata &meta, const string &text )
...
+		case LEVEL_VERBOSE:		sysLevel = LOG_DEBUG;	break;
+		case LEVEL_INFO:		sysLevel = LOG_INFO;	break;
+		case LEVEL_WARNING:		sysLevel = LOG_WARNING;	break;
+		case LEVEL_ERROR:		sysLevel = LOG_ERR;		break;
+
+		// debatable if FATAL is a LOG_CRIT or LOG_ALERT
+		case LEVEL_FATAL:		sysLevel = LOG_CRIT;	break;
+		default: CI_ASSERT_NOT_REACHABLE();
+	}
+	syslog(sysLevel, "%s %s", meta.toString().c_str(), text.c_str());
```
richardeakin: I was just testing here and it occured to me that this priority int (`sysLevel`) should probably not ever be less than LOG_NOTICE, as it won't appear in Console / system.log unless you modify some system config files (found that info [here](http://apple.stackexchange.com/questions/13484/messages-issued-by-syslog-not-showing-up-in-system-logs)).  Our usage of `ci::log::Level` is a bit different to syslog's priority anyway - for example if `CI_MAX_LOG_LEVEL` is set to 4 then `CI_LOG_V()` calls will be compiled out (as they are in release mode by default).

That said, If I were to structure this switch statement, I would still correlate the LEVE_WARNING : LOG_WARNING, and above, but everything else would default to `LOG_NOTICE`.

src/cinder/Log.cpp
```diff
@@ -318,6 +409,44 @@ void LoggerNSLog::write( const Metadata &meta, const string &text )
...
 }
 
+// ----------------------------------------------------------------------------------------------------
+// MARK: - LoggerNSLog
+// ----------------------------------------------------------------------------------------------------
+
+LoggerSysLog::LoggerSysLog()
+{
+	// Most general approach for determining app name.  Perhaps there is something more cinder specific?
+	// https://developer.apple.com/library/mac/qa/qa1544/_index.html
```
richardeakin: I don't think we store an app name string anywhere, perhaps we should. Though if we keep the approach here, we should switch to using [PlatformCocoa::getBundle()](https://github.com/cinder/Cinder/blob/app_refactor/include/cinder/app/cocoa/PlatformCocoa.h#L60) once app_refactor is merged.

richardeakin:
File rolling looks good, I really like the way you did the format string. Yea we should add some sort of way to specify when and how to roll files in the future (perhaps by file size, too), but this is a great start.

I'm fine with appending enabled by default. The only thing I'd suggest there would be to make the method variable name a little more descriptive, perhaps `bool appendToExisting`.

It looks like writing a log to a folder that doesn't exist is broken, which was pre-existing though we should fix it at some point. I'll open a separate issue for that. By the way, I added a [DebugTest](dafff1475eaa0f1b0cefdde3808e41ac5929c862) project which I was using when I originally wrote the log stuff, feel free to add any tests you think are relevant to that.

Thanks againf or all of this, really nice additions..

lucasvickers:
Hi Richard,
Sorry for the slow response -- I was on site doing some install testing.

I've made a bunch of small changes that address everything we're discussing.  I'll respond with a PR shortly.  In the mean time, I get a 404 when trying to access your [DebugTest](https://github.com/cinder/Cinder/pull/dafff1475eaa0f1b0cefdde3808e41ac5929c862)

src/cinder/Log.cpp
lucasvickers: I had the same issue until I found some info about /etc/asl.conf configurations.  I tried to comment about this during my PR but I wasn't clear enough.

I'm going to change it so it's always LOG_NOTICE and up.  Part of me doesn't love flooding the syslog with everything.  Chances are if you use the syslog you're linking it to some type of process manager and you really only need high priority messages.

richardeakin:
Hm, don't know what I did with that previous link, but DebugTest is in glNext branch [here](https://github.com/cinder/Cinder/blob/glNext/test/DebugTest/src/DebugTestApp.cpp).

src/cinder/Log.cpp
richardeakin: True, but we'll still be filtering log messages via cinder's ci::log::Level mechanism, so verbose things should still get filtered out, in release mode at least. We can think about adding per-logger levels in the future, but that would have to be runtime filtered of course.

lucasvickers:
* Now totally understand why it's easier to split the PR's apart :) will do so in the future
* Formatting - feel free to call me out on any formatting / spacing issues that I may miss
* Added some test cases to your DebugTest

I'll respond to the other stuff inline.

richardeakin:
Cool - about the separate PR's, if it is super inconvenient then combining them into one PR is okay, but in that case trying to keep the commits separated to specific feature additions or fixes helps when navigating the diffs and trying to make sense of things.

All in all, just happy that you're contributing! It's looking good... I don't have time at the moment but I'll give this branch a try in the next couple days and will probably merge thereafter.

lucasvickers:
Thanks for calling out the spacing and explaining the PR's.  I'm using this contribution as a way to get the details figured out before I jump on anything larger.  I'll update this PR with the few small changes and then you can give it a test.  I'll work on windows Event Log in the near future (2 weeks likely)

richardeakin:
Ok great. It looks like there'll be some merge conflicts to solve to match this up with the recent merges of #737 and #710, though nothing major. cheers!

lucasvickers:
Okay this should be all set now -- and yeah the merge is just that if statement.  I figured it would be ok.

richardeakin:
Merged, and I added some minor updates. Thanks again, these additions are really nice!

PR #725, opened 2015-02-25
Added getters for Frustum Planes

ryanbartley:
A small change for the purpose of extracting Frustum Planes.
- This enables easy access of the Planes contained in Frustum specifically for use with aspects of gl including but not necessarily limited to [gl_ClipDistance](https://www.opengl.org/sdk/docs/man/html/gl_ClipDistance.xhtml) functionality.


paulhoux:
Hi Ryan, is there a reason you're providing a non-const getter for the planes, too? Wouldn't it break the code if you'd adjust a single plane? It would no longer be a frustum.

ryanbartley:
I think probably because I was a zombie when I wrote this. There's no reason for the none const version.

PR #734, opened 2015-03-06
mac borderless fixes

richardeakin:
Merging against app_refactor since app impls have moved in that, amongst other changes. This includes:

* @lithium-snepo's fix for coming back from borderless and the resize buttons being gone. Internally it appears we need to resize the window to make this work (I tried a number of other things as well) - probably a bug on apple's part, but seeing as how the window is already visually changing, doesn't have any real side effects.
* caching the window title so that we can restore it after coming back from borderless
* `App::Settings::setTitle()` also sets the default `Window::Format`'s title. This seems like expected behavior, and also I think this is how things already work on windows (on mac, you had to explicitly call `Window::setTitle()`)

Fixes issues in #694, supersedes #709

### DIFF
--- a/include/cinder/app/cocoa/AppImplMac.h
+++ b/include/cinder/app/cocoa/AppImplMac.h
@@ -87,6 +87,7 @@
   @public
 	AppImplMac*					mAppImpl;
 	NSWindow*					mWin;
+	NSString*					mTitle; // title is cached becasue sometimes we need to restore it after changing window border styles
 	CinderViewMac*				mCinderView;
 	cinder::app::WindowRef		mWindowRef;
 	cinder::DisplayRef			mDisplay;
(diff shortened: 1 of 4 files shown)


include/cinder/app/cocoa/AppImplMac.h
```diff
@@ -87,6 +87,7 @@
   @public
 	AppImplMac*					mAppImpl;
 	NSWindow*					mWin;
+	NSString*					mTitle; // title is cached becasue sometimes we need to restore it after changing window border styles
```
notlion: "because" ;)

richardeakin: Thanks, don't want no body knowing I don't have no good english.  Seriously, the review is appreciated.

PR #735, opened 2015-03-07
Blur sample update

sansumbrella:
Use two FBOs with multiple attachments (one FBO for each size needed).
GPU and CPU frame timing added to params.
Separate methods for each phase of the blur process should be easier to follow.
Update shaders to 330 core so we can specify layout for FBO attachments.

### DIFF
--- a/samples/_opengl/MotionBlurVelocityBuffer/assets/motion-blur.fs
+++ b/samples/_opengl/MotionBlurVelocityBuffer/assets/motion-blur.fs
@@ -1,4 +1,4 @@
-#version 150 core
+#version 330 core
 
 uniform sampler2D uColorMap; // texture we're blurring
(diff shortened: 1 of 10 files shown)


sansumbrella:
Only tested on OSX 10.9.5 with an AMD card at the moment.

sansumbrella:
Totally busted on Win7 / NVidia card.

sansumbrella:
Works on OSX 10.9 with an AMD GPU and on Windows 7/VS2013 with an NVIDIA GPU.

sansumbrella:
Rebased after the latest glNext/appRefactor merge.

richardeakin:
Ah right, thanks for updating!

sansumbrella:
Thanks for merging!

Issue #717, opened 2015-02-19
LoggerFile writing to file in non-existent folder should still work


richardeakin:
Can test by enabling [this line](https://github.com/cinder/Cinder/commit/dafff1475eaa0f1b0cefdde3808e41ac5929c862#diff-73bdc39eb7181f7144af0797e5338d5dR50) in DebugTest.

lucasvickers:
What's your take, should we create the folder structure?

richardeakin:
Yea I mean, I think it would be convenient and what the user expected, if they specify a folder/file combo and the folder doesn't exist, then we create the folder too. @andrewfb what do you think?

andrewfb:
I agree - seems like this is what the user would want in basically every scenario.

richardeakin:
Fixed in #737, thanks @lucasvickers

PR #737, opened 2015-03-08
creates containing folder for logging

lucasvickers:
Related to issue #717

* There may be a more appropriate way to get an error message to the user other than ```cerr```, but I didn't want to rely on any app contexts, etc.
* I hit the ```cerr``` when I call ```testEnableFile();``` and it gives an empty path, but not when I call ```testEnableBadFilePath();``` with a bogus path.  I'm a little confused about it, but not really worried.

### DIFF
--- a/src/cinder/Log.cpp
+++ b/src/cinder/Log.cpp
@@ -299,8 +299,19 @@ LoggerFile::~LoggerFile()
 
 void LoggerFile::write( const Metadata &meta, const string &text )
 {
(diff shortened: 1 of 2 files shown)


richardeakin:
I think I see why the call to `create_directories()` fails, because the default mFilePath is "cinder.path" - this works many times but I think it might function differently across platforms. I'm going to fix after merging with the newly added app::Platform.

PR #744, opened 2015-03-15
Allow access to environment variables on OS X and iOS

pizthewiz:
A very basic add of `getEnvironmentVariables()` to Utilities. Would it make sense to hoist this to `Platform` and `CocoaPlatform`?

As an aside, `CINDER_COCOA` is a little misleading since iOS does not have _Cocoa_ but instead _Cocoa Touch_ no? I'd almost rather `CINDER_OSX` and `CINDER_IOS` :wink:

### DIFF
--- a/include/cinder/Utilities.h
+++ b/include/cinder/Utilities.h
@@ -26,6 +26,9 @@
 
 #include <string>
 #include <vector>
+#if defined( CINDER_MAC )
+	#include <map>
+#endif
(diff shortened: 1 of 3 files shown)


andrewfb:
Thanks for this - I moved this implementation into Platform and also added a matching one for MSW.

pizthewiz:
Awesome, I'm glad that was helpful and I really like the `Platform` abstraction so far!

PR #751, opened 2015-03-19
Added getSubnetMask() to both System and NetworkAdapter instances

zeh:
As per [this discussion](https://forum.libcinder.org/#Topic/23286000002295005), this adds information about an adapter's subnet mask to NetworkAdapter instances and a static `System::getSubnetMask()` call that works similarly to `System::getIpAddress()`, but for the subnet mask of the most appropriate network adapter.

    console() << "Your IP is " + System::getIpAddress() + ", and your subnet mask is " + System::getSubnetMask() << endl;

The reason a subnet mask is needed is so a broadcast IP [can be properly calculated](http://learn-networking.com/network-design/how-a-broadcast-address-works). Calculating the broadcast IP is necessary so we can [establish OSC communications](https://forum.libcinder.org/topic/osc-broadcast) without having to hard-code the OSC listener IP, or pass it as a parameter; it allows applications in the same network to communicate without additional configuration or tools.

Calculating the broadcast IP is not part of this pull request.

The code should be good for WinRT (XP), Win Vista+, OSX, and iOS, since I just exposed properties that already existed without change to the logic. However, I have only tested in Windows 7 so far.

### DIFF
--- a/include/cinder/System.h
+++ b/include/cinder/System.h
@@ -77,17 +77,20 @@ class System {
 	  public:
 		const std::string&	getName() const { return mName; }
 		const std::string&	getIpAddress() const { return mIpAddress; }
+		const std::string&	getSubnetMask() const { return mSubnetMask; }
 
-		NetworkAdapter( const std::string &name, const std::string &ip )
-			: mName( name ), mIpAddress( ip ) {}
+		NetworkAdapter(const std::string &name, const std::string &ip, const std::string &subnetMask)
+			: mName( name ), mIpAddress( ip ), mSubnetMask( subnetMask ) {}
 
 	  private:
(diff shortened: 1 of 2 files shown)


andrewfb:
Hi - thanks for this. I made a few changes to get it working on OS X and iOS. Also PrefixLength maps to a different expression of subnet masks on WinRT. For now this will be an inconsistency but if someone wants to write the binary conversion math and make it IPv6 aware it would be welcome. I merged this against glNext instead since it's where active development is taking place primarily. Closing for now.

zeh:
Cool, thanks Andrew.

Wasn't fully aware of glNext's role on the dev efforts, will use it next time.

PR #757, opened 2015-03-20
extended texture3D

dahiro:
extended texture3D to update parts of 3D texture
extended texture3D to define other data types than UNSINGNED_BYTE


andrewfb:
Apologies for the delay and thanks for this addition...

PR #758, opened 2015-03-20
geom::Rect fixes

BanTheRewind:
Colors are optional for geom::Rect. Disabled by default. Assigned new default colors.


richardeakin:
Looks good to me, will write a couple things in line.

include/cinder/GeomIo.h
```diff
@@ -210,8 +210,12 @@ class Rect : public Source {
 	Rect( const Rectf &r );
 
 	Rect&		rect( const Rectf &r );
+	//! Enables default colors. Disabled by default.
+	Rect&		colors();
```
richardeakin: It looks like we handle this differently in different places, however I think the best way is to pass a bool in, i.e. `colors( bool enable = true )`, instead of having hte addition `disableColors()` method. This is how it is done in [Icosehedron](https://github.com/cinder/Cinder/blob/glNext/include/cinder/GeomIo.h#L269), but obviously not in [Cube](https://github.com/cinder/Cinder/blob/glNext/include/cinder/GeomIo.h#L237-L241).

Above all, I think consistency is important so we should probably pick one or the other and make them all match.

src/cinder/GeomIo.cpp
```diff
@@ -462,36 +462,22 @@ void Target::generateIndices( Primitive sourcePrimitive, size_t sourceNumIndices
 const float Rect::sNormals[4*3] = {0, 0, 1,	0, 0, 1,	0, 0, 1,	0, 0, 1 };
 const float Rect::sTangents[4*3] = {0.7071067f, 0.7071067f, 0,	0.7071067f, 0.7071067f, 0,	0.7071067f, 0.7071067f, 0,	0.7071067f, 0.7071067f, 0 };
 
-Rect::Rect()
+Rect::Rect() : mHasColors( false )
```
richardeakin: spacing: according to the [styleguide](https://github.com/cinder/Cinder/blob/glNext/CONTRIBUTING.md#example-class-implementation), should be:

```
Rect::Rect()
    : mHasColors( false )
{
...
```

include/cinder/GeomIo.h
BanTheRewind: I prefer enable with flag. I didn't notice that in ico. Was trying to match cube.

src/cinder/GeomIo.cpp
```diff
@@ -553,6 +552,24 @@ AttribSet Rect::getAvailableAttribs() const
 	return { Attrib::POSITION, Attrib::NORMAL, Attrib::TEX_COORD_0, Attrib::COLOR, Attrib::TANGENT };
 }
 
+void Rect::setDefaultColors()
+{
+	// upper-right, upper-left, lower-right, lower-left
+	mColors[0] = ColorAf( 0.0f, 1.0f, 0.0f, 1.0f );
```
richardeakin: I personally find it more readable to use `0` and `1` here instead of `0.0f` and `1.0f`, which converts perfectly to floating point precision and is unambigious in these cases, but really not a big deal.

andrewfb:
It looks like this change may have gotten buried in a larger list; I'll cherry-pick this and make Rich's suggested tweaks so you can focus on the deferred sample. I'm also going to align all of the primitives to match the colors( bool enable ) convention.

andrewfb:
Cherry-picked 06cdaf959afc1638d5a3dd9417c793864e82ff5b

PR #759, opened 2015-03-20
DeferredShading sample

BanTheRewind:
I've got this working in Xcode now (by downgrading to GLSL 410). I also got rid of the lambdas in ::draw() that were no longer necessary and replaced window setup with scoped objects.


richardeakin:
Nice to finally see this. Just tested on Xcode and appears to be working with your latest updates.

Thats a lot of passes!  Would be super to be able to enable / disable some of them from the tweak bar, to  get an idea of what each pass does and how much performance it costs.  By the way, I'm getting about 12fps on a mbp-retina (debug or release doesn't seem to make a difference), would be nice to get this up to interactive rates.   One easy thing I tried was halving the resolution to 960x540, which brought the framerate to ~45fps and feels pretty nice.

BanTheRewind:
I just updated the app to make it a little more interesting. Let me look into getting the frame rate up on OSX. I noticed it was slower over there, as well. Kinda sucks that I get 60fps in Windows 8.1 *on* my Bootcamped Mac. Gets like 37 on the OSX side.

Pass toggling can be a liiiittle bit tricky here because if you skip a pass, the next pass needs to know which FboRef to read. I'll figure out. Don't merge just yet. I'll keep at it.

BanTheRewind:
I'll add toggles for the other passes, but bloom and FXAA were easy to throw in. I'm curious what kind of improvement you get, if any. I would imagine that the SSAO and DoF passes are the most expensive.

BanTheRewind:
Also try reducing the number of shadow samples to improve FPS. I think I had it at 8, which is a lot.

richardeakin:
Ah yea, looks pretty cool now. Reducing the # shadow samples improves things a bit, I was also wondering if the shadow / blur texture sizes could be reduced without seeing much degradation, from my quick glimpse the shader buffer size is 2048x2048, and the blur buffers are full window size. In any event, reducing the window size a tad does help, too, although of course not for fullscreen.

BanTheRewind:
Yeah, I'll do those things. I'll run it on my 2010 mbp and try to get it to 60 there. :)

BanTheRewind:
I'll try to bring the gl version to 330, as well.

richardeakin:
Nice, well 30fps+ will probably work IMO, if it means it looks better.

One other thing, can you use the icons in [samples/data](https://github.com/cinder/Cinder/tree/glNext/samples/data) instead of baking in new ones? Minor attempt at reducing the size of project clone --depth=1

BanTheRewind:
Ah, will do. I was wonderng where they went.

BanTheRewind:
OK, give this a shot. Using "samples/data" for icons. I have flags for all passes. I can get 60fps on my 2010 MBP with some passes disabled at 900x500-ish, so at least 30 fps should happen on newer machines (more like 60). I set the initial resolution of the app to 1280x720 as I was running out of video memory on my MBP at 1920x1080. Oops!

simongeilfus:
Cool sample!

Regarding the performances (around 10/15 fps in fullscreen on my mbp retina) the first thing I will look into in any deferred renderer is tweaking the color format of my render targets. This might seems to be a minor issue, but as you're writing a ton of data for each pixels, fillrate is usually the first suspect when looking for the bottleneck... (you can easily verify that by rotating the camera in such a way that every pixels got written in the gbuffer)

In your case there's quite a lot of Fbos and most of them use the extremely expansive GL_RGBA32F. Memory wise, and more importantly fillrate wise you really don't want to use such a high precision. That's really too much information to write each frame. Doing a quick "replace all" GL_RGBA32F by GL_RGBA16F already gives a small performance boost, but you probably want to choose the right format for each pass. I would personally arrange my geometry buffer in a way that I could use GL_RGB10_A2 in most places and eventually GL_RGBA16F in the rare case where you need it. You'd be surprised how much it can help. (A quick replace all with GL_RGB10_A2 seems to break one of the passes but raise the framerate from 10 to 50 in fullscreen. This will probably need a bit more work than a ctrl-shift-f though).

Are you storing the depth twice in your geometry buffer? (depth attachment and normalDepthBuffer?)

Anyway the only place where you need so much precision is your position, normal and depth buffers. The rest should work nicely with 8bits. And actually I don't think anybody stores the position in the gbuffer because of this exact problem. So the next thing I will look into is getting rid of the position buffer (From the 7x32bit floats you can easily go down to 2x16bit floats for the normals and 1x32bit float for the depth, the rest can be reconstructed later). This will probably make the sample a tiny bit more complicated, but IMO that's the way to go.

I can definitely provide some code if you want.

simongeilfus:
So I did a quick test and changing every GL_RGBA32F to GL_RGB10_A2, except for mTextureFboGBufferNormalDepth and mTextureFboGBufferPosition that I set to GL_RGBA16F raise the framerate to 45/50fps in fullscreen with shadows disabled. I'm not sure what's the issue with shadow mapping but it seems to be the second thing slowing down everything...

BanTheRewind:
Let me check on the double depth storage. I'm using the depth buffer for shadows. I might have another for SSAO and DoF, but I don't think the values are the same.

BanTheRewind:
Also, I think I would like to leave the position buffer in. Maybe do a second version with more complex optimizations (like single pass shadows).

BanTheRewind:
And to be clear. I'm thinking it might make sense to strip out passes of this sample for the "basic" version. Then work from here for an advanced one. One that's a learning tool to explain the concept, and one that's more of a grab bag of techniques. 




One thing this really needs is single FBO with multiple attachments for all the post processing passes.

BanTheRewind:
I've mentioned this before, but I'm thinking it might make sense to strip out passes of this sample for a "basic" version, then work from here for an advanced one. One that's a learning tool to explain the concept, and one that's highly optimized. I don't know that we can really do both in one sample. The Right Way is going to be really tough to wrap one's mind around, whereas a basic version won't get us the bells and whistles.

BanTheRewind:
Let me know what you think of breaking them out like this. The advanced version is where we can trim down... use one FBO for all the post-processing passes, calculate position from depth, etc. I feel like if someone wraps their brain around the basic version, they'll be able to dig into the advanced one. But with so many shaders and passes, the "advanced" version just isn't a good learning tool.

simongeilfus:
Absolutely, it makes sense to split that into two different samples. 
I'd be happy to see if I can help on the advanced sample and I'm sure we can both have something easy to grasp and optimised.

sansumbrella:
This looks pretty slick.

I'm not seeing any shadows on my work machine. Otherwise runs well. Toggling them on/off has no effect.
AMD FirePro D300, OSX 10.10.2 (Yosemite)

BanTheRewind:
Will check in a few.

BanTheRewind:
Fixed. In adding a shadow flag, I inadvertently moved a ScopedTextureBind to inside of an if statement, thus it was always out of scope. I couldn't tell the shadows were missing on my old machine at lower res at home. Really obvious on my work machine.

BanTheRewind:
OK, I think the basic sample is dialed. I'm going to dive into the advanced one and start on cleanup and optimization. A quick bit of history... this rendering engine was a bit of a volley between a Cinder sample and a production project. We tried a few techniques for rendering shadows, and it looks like we have are storing a lot of unused data because of it. Namely, as @simongeilfus pointed out, depth is being stored twice (there was a point where we needed both to toggle between shadows to see which we liked more).

I'll work on cutting out some extra code, then get us to only use the depth buffer properly. This means A) no more "depth" value in the normal map and B) no more position map. This will require a tweak to how SSAO and DoF work, but it will be a big improvement for sure.

BanTheRewind:
Done for today. Check out the latest. The advanced sample now uses a single FBO for the post-processing passes and just ping pongs. Depth is no longer stored in a color texture. The DoF shader was re-written with bokeh. General clean up, as well.

@simongeilfus Do want to take a crack at calculating the position from the depth buffer in lieu of a position buffer?

After that, I would like to get a velocity buffer in there for motion blurs. Beyond that... just make the visuals a little more interesting.

simongeilfus:
Hey Stephen!
Of course, here you go: https://gist.github.com/simongeilfus/bf0938f74cbd7997bfeb
I'm sorry this is a Gist and not a PR, but I'm still unsure what is the proper solution when working on someone else fork. Git doesn't seem to allow forking a fork....

Anyway, I removed the gbuffer position attachment and updated the different shaders. Another common thing to do is to encode the normal, which in the case of your gbuffer could probably lower down the number of attachments to only two. Which would be a really tiny lightweight gbuffer...

simongeilfus:
So basically what I did is to replace the position Sampler in the different fragment shaders by the depth Sampler and added the following uniforms:

```c++
uniform mat4        uProjMatrixInverse;
uniform vec2        uProjectionParams;
```
Then added the following function that linearize the depth and reproject the position:
```c++
vec4 getPosition( vec2 uv )
{
    float depth             = texture( uSamplerDepth, uv ).x;
    float linearDepth 		= uProjectionParams.y / ( depth - uProjectionParams.x );
    vec4 posProj        	= vec4( ( uv.x - 0.5 ) * 2.0, ( uv.y - 0.5 ) * 2.0, 0.0, 1.0 );
    vec4 viewRay        	= uProjMatrixInverse * posProj;
    return vec4( viewRay.xyz * linearDepth, 1.0 );
}
```
From there you can ditch the extra attachment of mFboGBuffer and update the gl::drawBuffers call.
That's it!

ps: This is what I would use for the normals:
```c++
vec2 encodeNormal(vec3 n){
    vec2 enc = normalize(n.xy) * (sqrt(-n.z*0.5+0.5));
   	enc = enc*0.5+0.5;
    return enc;
}
vec3 decodeNormal( vec2 enc ){
    vec4 nn = vec4(enc.xy,0,0)*vec4(2,2,0,0) + vec4(-1,-1,1,-1);
    float l = dot(nn.xyz,-nn.xyw);
    nn.z = l;
    nn.xy *= sqrt(l);
    return nn.xyz * 2 + vec3(0,0,-1);
}
```

BanTheRewind:
Thanks much! Can't wait to give this a shot.

BanTheRewind:
I could give you write access if you prefer. My next couple days are starting to look packed so I probably can't get back into this until Thursday.

BanTheRewind:
Added. Thanks!

BanTheRewind:
Hey, quick note just to say don't sweat the normal encoding. I have it in there now and it works great. Thanks! I'm storing the normals and material IDs in a three channel float buffer now. Because of the data type change, I need to tweak the material part a bit. Once that's finished, I think the "tech" part of this will be done. I'm not going to add a V-buffer after all. I'd rather focus on blowing out the animation / scene into something really interesting.

BanTheRewind:
I'm now encoding normals. I had some issues storing material IDs and encoded normals in the same buffer. It actually uses a hair less memory to use two buffers with one channel (GL_R8I) for materials and two channels (GL_RG8) for encoded normals. Easier to work with, too, so that's nice.

I was able to get all color buffers down to RGB8 / GL_BYTE and still have it look good. Getting to work on making it look nice now...

BanTheRewind:
Hey, all. This is getting close to wrapped. I wanted to blow out the
visuals a little more, but I'm starting to get busy here. I would also like
my Cinder efforts to move back to gl::Effect.

Please pull the latest if you have a chance. Looking for some feedback.
Here's what's left (note there are a couple questions for the team in
there).

TODO list

   - BUG Shadow map goes wonky on resize
   - BUG Switching to debug view, then back, then turning off SSAO flips
   image
   - BUG Normal encoding not working on my old Mac. Might just have to live
   with it. Works fine elsewhere.
      - Anyone else seeing this?
   - Finish writing comments
   - Should I rename the shaders? Where did we net out on that?
      - *.vs/*.fs
      - *.vert/*.frag
      - *_vert.glsl/*_frag.glsl (current)
   - Add credits.
      - Michael Latzoni and myself under Wieden+Kennedy (99% of this was
      made here with their blessing)
      - Simon Geilfus
      - Recommended way to write this in?
   - Verify performance across machines
      - Let me know what FPS you're getting!

BanTheRewind:
Oh, I also wanted to explore shader includes in the advanced sample to cut
down on duplicate code.

sansumbrella:
The latest (dca7af1) no longer renders the final output on my work machine (Mac Pro, AMD FirePro D300).

Attached is the latest debug gbuffer contents, which hopefully helps:
screen shot 2015-03-26 at 5 26 37 pm

It seems like a few channels aren't being read/written correctly.

BanTheRewind:
The normal encoding/decoding isn't working for you, either. My Nvidia card displays it OK, but that's two AMDs where it isn't working. Let me investigate. I'm wondering if maybe GL_RG16F is not AMD friendly.

sansumbrella:
I don't think it's the GL_RG16F that's the issue; I use a GL_RG16F buffer for storing velocity in the motion blur sample which runs fine on my AMD machines.

sansumbrella:
However, there may be an issue where you still need to write 1.0 to the green and alpha channels of your RG16F buffer. I got mangled/no output when only writing to the first two channels of the two-channel buffer. Will try that here shortly...

BanTheRewind:
Ah, I'm in there right now. Will give it a shot.

BanTheRewind:
Try that.

sansumbrella:
The fix was simple, though it may not work on PC. It didn't require the weird buffer filling workaround, either.

Specify the layout location for the gBuffer outputs in the shader:
```
layout (location=0) out vec3	oAlbedo;
layout (location=1) out int	oMaterial;
layout (location=2) out vec2	oNormal;
```

screen shot 2015-03-26 at 6 38 11 pm

BanTheRewind:
Ah, weird. I'll get that there. I had the locations in there, but removed them while trying to reduce GLSL version. Didn't realize <4 supported these!

BanTheRewind:
Updated. I also just realized I still had the basic sample at 400.

richardeakin:
Nice, working here now too after the layout change. On a 2013 mbp with nvidia gpu, I'm seeing 20fps with shadows enabled, a solid 60fps with shadows disabled.  Going to full screen drops the framerate (with shadows still disabled) down to 30fps, but I think this is acceptable, given the nature of the sample.

One aesthetic suggestion, what about adding an additional geom so the scene has more than spheres? For example, I turned the rotating spheres into floating cubes in [this commit](https://github.com/richardeakin/Cinder/commit/3b7834ef4b36bdcbbefb3aabdee269b048b88e08).  To me, this adds a little bit of variance, also shows a bit more of the FXAA work:

screen shot 2015-03-27 at 4 23 55 am

Looking pretty good from here. It would be nice if the shadow pass could be optimized, but I don't have any suggestions on how to do that. Concerning the naming of the shaders, its your call, there are various naming schemes in the samples from different committers, but the majority seem to end in .vert / .frag, etc. If the shader is included or generic, I choose to give it the .glsl suffix.

BanTheRewind:
Yeah, I agree with the additional geometry. Icosahedrons look really nice and seem to kick out more bokeh in the DoF pass. Will update.




Heads up. I was working on reducing the overall amount of code and consolidating the buffers. All is well, except I have a small bug to fix where the depth is clearing before the DoF pass, so everything is blurry. I know exactly how to fix it, it was just late and I needed to get to sleep. Will fix in a few and swap out the rotating spheres for icosahedrons.




Thanks,

Steve

BanTheRewind:
Oh, and there is a way to optimize shadows. One method is to pass the shadow camera's matrices during the G-buffer pass and you store that depth instead of the depth from the camera view. Then, use a color channel to store the depth for SSAO, etc. You essentially bake the shadow map into the G-buffer. It's a little trickier than that reading everything back, but it can work.




I've also seen single-pass shadows using a similar technique, but done using the stencil buffer. My gut says this is the right way, but I just need to read up on how to do it.

BanTheRewind:
Check out the latest. I've done as much consolidating as is reasonable for the FBOs. I played with using a single FBO with attachments for everything, but managing when to clear what got out of hand. Now there are four FBOs: G-buffer, half window size (bloom and SSAO), shadow map, and a ping-pong FBO for doing everything else.

The shaders have changed a little bit. I need to update the Xcode version to reflect this. There is no more bloom shader. Instead, there is a light accumulation buffer followed by blur passes. This allows the creation of subtle light trails and an emission pass that doesn't require materials (ie, emission texture, dynamically drawing light, etc).

Also, the spheres on the floor are now shiny, spinning icosahedrons. ;)

pizthewiz:
It looks like `debug_frag.glsl` didn't get added to the repo?

BanTheRewind:
Added it for the advanced sample, but forgot about the basic one. It's in there now.

BanTheRewind:
Xcode samples tested successfully. Nothing to commit.

richardeakin:
Oh yea, that's looking really nice now.  I still see a lower framerate with shadows on, probably from fill rate, but it is quite up to the interactive level so I think it's good to go.  If someone wants to take on improving the shadow pass to get it up to 60fps, that'd be super, but it can be done after merging and is probably easier to do so. 

Thanks for these killer samples, Steve!

BanTheRewind:
np. I can take a stab at single pass shadows. I'm implementing this engine for the DK2 so need to get hyper efficient.




I'm thinking it's pretty buttoned up for now. If we can pull this, I'll start working on faster shadows in a separate feature branch.

PR #764, opened 2015-03-22
gl.h as umbrella header

### LINKED ISSUE #600: turn gl.h into an umbrella header
I've been wondering if we couldn't restructure the gl headers a bit to make our day to day lives easier in terms of having to #include things and what not.  This is a bigger deal in glNext because the number of headers in cinder/gl/* has grown from 12 to 21, and will likely grow a bit more before it is finished.

So, strawman proposal for gl headers:

* the current contents of gl.h is separated into components:
  * cinder/gl/platform.h: all of the glload and platform specific headers. Basically all the cross platform preprocessor craziness
  * cinder/gl/draw.h: convenience drawing utils
  * cinder/gl/scoped.h: all the RAII binding stuff
  * ~~cinder/gl/utils.h~~ cinder/gl/wrapper.h: gl functionality wrapped in namespace `ci::gl`
    * After beginning on the re-organization, a utils.h file doesn't make sense for things like 
* then, gl.h becomes an umbrella header: is made to include all of these, and also all of the other widely used gl headers
  * gl/Texture.h, gl/Fbo.h, gl/Shader.h, gl/GlslProg.h, etc..

*Note*: I've chosen to use lower case filenames as there are no public classes in these files. If others would instead prefer them to star  upper case than speak now. :)

This approach would be backwards compatible, but also allow people to include what they needed if thats how they want to do it.

One thing that got me thinking about this (again) was reading about [clang's experimental module support](http://clang.llvm.org/docs/Modules.html), and how such a structure would facilitate that eventually, if that feature makes it into the C++ standard.

richardeakin:
Closes #600, and follows the structure in that issue description, so I won't repeat it here.

Through initial usage, I've found the following differences.

__Pros:__
- For user apps, if you `#include "cinder/gl/gl.h"` then you no longer need to worry about whether you need cinder/gl/Shader.h, or some other gl header - they're all there for you to use.
- library code can include less and be more specific, as little as just cinder/gl/platform.h.

__Cons:__
- Files that were not directly including gl/gl.h, but instead pulling it in through some other header, like gl/Texture.h or gl/GlslProg.h, will need to be updated in some cases. For example if you're using gl::draw() functionality, you'll need to include cinder/gl/draw.h, or if you're using scoped RAII objects then you'll need to include cinder/gl/scoped.h.  To me, seems like something people will get used to and won't look back.
- The precompiled size of an app including gl.h will grow a tad, my tests with BasicApp.cpp show it going from 117,097 lines before to 126,274 lines.

__Profiling compilation times__

Below is some compilation profiling, which shows in most cases that compile times actually went down in some cases, even dramatically for the windows cinder project. I suspect that mac user apps compile time went up a tad because pch is on there (I didn't try without), but repeated builds will probably be faster since gl.h is usually in the pch.

On mac, profiled with the following command (only ‘real' output is listed here):
```
time xcodebuild -project ${PROJECT)
```
On windows, enable with: Tools -> Options -> Projects and Solutions -> VC++ Project Settings -> Build Timing

Results are in seconds, with 'before' being current glNext branch, and 'after' being this PR. comma separated values are just the results from separate runs.

cinder mac
before: 55.755, 50.67
after: 53.980, 55.49

cinder windows
before: 192, 200.91
after: 153.674, 149.643

BasicApp mac
before: 5.632, 3.614
after: 5.012, 3.673

BasicApp win
before: 8.824, 6.75
after: 6.375, 6.313

FallingGears mac
before: 9.481, 7.709
after: 10.711, 11.809

FallingGears win
before: 45.139, 40.107
after: 29.564, 26.852


sansumbrella:
This is great to see. It was a little tedious to include everything when just starting a project, and definitely slowed down new (Cinder) developers.

PR #767, opened 2015-03-25
Added GlslProg::Format getters for the shader paths

simongeilfus:
Would this be something interesting to add to the current GlslProg::Format implementation?
This is probably not going to be used a lot but I can definitely see this being useful in some specific cases.

Let me know what you think.
Cheers,
Simon.

### DIFF
--- a/include/cinder/gl/GlslProg.h
+++ b/include/cinder/gl/GlslProg.h
@@ -149,7 +149,19 @@ class GlslProg : public std::enable_shared_from_this<GlslProg> {
 		void				setLabel( const std::string &label ) { mLabel = label; }
 		//! Sets the debugging label associated with the Program. Calls glObjectLabel() when available.
 		Format&				label( const std::string &label ) { setLabel( label ); return *this; }
-		
+        
+        //! Returns the fs::path for the vertex shader. Returns an empty fs::path if it isn't present.
+        const fs::path&	getVertexPath() const { return mVertexShaderPath; }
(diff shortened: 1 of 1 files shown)


richardeakin:
I can't see how it would hurt anything, although technically us storing shader paths is an implementation detail to enable searching for includes relative to the shader. Internally, it would probably mean that GlslProg::Format didn't have to friend GlslProg, if that matters to anyone.

Have you found already found a usage for these getters?  One thing to note, is that the paths are currently forgotten once the GlslProg is constructed (they only live in GlslProg::Format). This may have to change if / when we support live shader updates, but that won't be this release.

Btw, looks to need one more tab for each of the added lines.

simongeilfus:
Hey Rich,

> Have you found already found a usage for these getters?

Yes, in the code I'm writing right now, you basically don't create GlslProgs yourself but instead provide an optional GlslProg::Format to my class constructor. The shader itself get constructed somewhere else, and having access to the paths allows me to not have to pass the paths twice... if that makes any sense... another way to put it is that it allows me to leave stuffs like "live shaders" totally separated from the rest...

> Btw, looks to need one more tab for each of the added lines.

Arf, yes I don't know why xcode keeps messing up my tabs... I'll update my fork!

> Internally, it would probably mean that GlslProg::Format didn't have to friend GlslProg, if that matters to anyone.

Probably has nothing to do with this PR, but I find myself doing this a lot with ::Format or ::Options subclasses... how would it be a problem?

richardeakin:
The use case makes sense to me, especially since it makes live-reloading easier for this release, when using something like a file watcher.

PR #770, opened 2015-03-27
Gl next debug break

richardeakin:
- Added `LoggerBreakpoint`, a logger that doesn't print anything but instead triggers a breakpoint. You don't need to use it directly, instead you can write:

```
log::manager()->enableBreakOnError();
// or
log::manager()->enableBreakOnLevel( log::LEVEL_WARNING );
```
- Added `CI_BREAKPOINT()`, a cross-platform macro to trigger breakpoints at runtime. Resides in cinder/Breakpoint.h 
- Added a new macro called `CI_LOG_EXCEPTION`, which takes a user description ostream and the exception itself, and prints something informative like:

```
|error  | FXAA::FXAA()[38] exception caught loading fxaa.vert / frag, exception type: cinder::gl::GlslProgCompileExc, what: FRAGMENT: ERROR: 0:44: '.0' : syntax error: syntax error
```

I find that I almost exclusively log exceptions with this now in my projects, as it is usually much more verbose and helpful than anything else you'd code on the spot.

### DIFF
--- a/include/cinder/Breakpoint.h
+++ b/include/cinder/Breakpoint.h
@@ -1,5 +1,5 @@
 /*
- Copyright (c) 2014, The Cinder Project
+ Copyright (c) 2015, The Cinder Project
 
  This code is intended to be used with the Cinder C++ library, http://libcinder.org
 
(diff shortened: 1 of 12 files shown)


paulhoux:
Hi Rich,

Visual Studio is complaining about the signatures for ```setFileLoggingEnabled``` : due to the fact they both have default parameters, the two versions of the call are exactly the same if you omit them:

```mgr->setFileLoggingEnabled( true, path );```
C2668: ambiguous call to overloaded function.

You can fix this by explicitly creating a version of the function with only 2 parameters.

richardeakin:
Thanks for pointing that out, fixed in #786.

PR #777, opened 2015-03-29
Signal Connections can be blocked/unblocked.

sansumbrella:
boost::signals2 provided a shared_connection_block type that let users temporarily block calling of a given connection. This was pretty useful for pausing/unpausing type behavior.

This adds a `block`/`unblock` method pair to `ci::signals::Connection` that lets users temporarily suspend calling of a connected function.

### DIFF
--- a/include/cinder/Signals.h
+++ b/include/cinder/Signals.h
@@ -42,7 +42,7 @@ namespace detail {
 struct SignalLinkBase {
   public:
 	SignalLinkBase()
-		: mRefCount( 1 )
+		: mRefCount( 1 ), mActive( true )
 	{}
 	virtual ~SignalLinkBase()
 	{
@@ -71,8 +71,24 @@ struct SignalLinkBase {
 		return mRefCount;
 	}
 
+	void block()
+	{
+		mActive = false;
+	}
+
+	void unblock()
+	{
+		mActive = true;
+	}
+
+	bool isActive() const
+	{
+		return mActive;
+	}
+
   private:
-	int	mRefCount;
+	int		mRefCount;
+	bool	mActive;
 };
 
 //! Base Signal class, which provides a concrete type that can be stored by the Disconnector
@@ -110,6 +126,10 @@ class Connection {
 	bool disconnect();
 	//! Returns whether or not this Connection is still connected to the callback chain.
 	bool isConnected() const;
+
+	void block();
+	void unblock();
+	bool isActive() const;
 
   private:
 	std::weak_ptr<detail::Disconnector>		mDisconnector;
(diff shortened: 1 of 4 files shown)


sansumbrella:
For reference, I am using this functionality in the Choreograph samples (which previously used boost::signals2 via ci::signals).

https://github.com/sansumbrella/Choreograph/commit/199b987865942d7fb78aa273764195b67adbf853?diff=split#diff-06b5efc9d4ac04ea3c3286af621d121fR46

richardeakin:
Looks useful to me, I think we should add this. Though first, would you mind adding a couple things?

* doxygen documentation for public API added to Connection
* some simple test code to [SignalsTest](https://github.com/cinder/Cinder/blob/glNext/test/SignalsTest/src/SignalsTest.cpp)
* it would be nice to ensure there's no serious benchmark regressions since SignalLink in particular is pretty finicky, but looking at the diffs I don't imagine there will be.

One other thing, what do you think about the names `activate()` and `deactivate()` instead of `block()` / `unblock()`?  This would match the `isActive()` method a bit better, and seems slightly clearer in intention IMO.

sansumbrella:
Happy to provide documentation and different names; `block`/`unblock` are really just carry-over from boost signals.

Are the benchmarks for this in SignalsTest, or do they live somewhere else?

sansumbrella:
As for different names, maybe: `enable()` and `disable()` (also changing the bool to `isEnabled`)? I think those words are pretty common throughout the Cinder codebase, while `active` is not.

richardeakin:
Benchmarks are in [SignalsBenchmarks.cpp](https://github.com/cinder/Cinder/blob/glNext/test/SignalsTest/src/SignalsBenchmark.cpp), you can run it from the [SignalsTest](https://github.com/cinder/Cinder/tree/glNext/test/SignalsTest) xcode or vxproj.  The tests are meant to ensure we have correct functionality when compilers change or if we decide to muck with the signals internals, benchmarks file is just to get a gauge on emission speeds.

I like the names enable() / disable(), they are widely used and make sense to me here as well.

sansumbrella:
Benchmark results:
```
// No check.
sizeof( signal ): 48 bytes
Benchmark: emmission (1 slot): OK
	per emission: 12.872ns
Benchmark: emmission (5 slots): OK
	per emission: 31.038ns, per slot: 6.2076ns
Benchmark: emmission with groups (2 groups, 4 slots): OK
	per emission: 33.544ns, per slot: 8.386ns
Benchmark: plain callback loop: OK
	per round: 1.893ns
Benchmark: std::function callback loop: OK
	per round: 2.238ns

Program ended with exit code: 0
// With isEnabled() check.

sizeof( signal ): 48 bytes
Benchmark: emmission (1 slot): OK
	per emission: 13.524ns
Benchmark: emmission (5 slots): OK
	per emission: 32.041ns, per slot: 6.4082ns
Benchmark: emmission with groups (2 groups, 4 slots): OK
	per emission: 34.826ns, per slot: 8.7065ns
Benchmark: plain callback loop: OK
	per round: 1.971ns
Benchmark: std::function callback loop: OK
	per round: 2.224ns
Program ended with exit code: 0
```

I think it's worth a nanosecond of our time.

richardeakin:
Thanks for the updates!

PR #782, opened 2015-04-02
added missing includes to opengl samples

gaborpapp:
I had to add these missing includes to build the samples in the _opengl sample directory.


richardeakin:
Thanks for opening this. For the samples, IMO it might be nicer to `#include "cinder/gl/gl.h"`, perhaps even removing the other explicit gl includes.  I think the main includers of things like gl/scoped.h would be library code.

gaborpapp:
Ok. I just wanted to make as little modifications as possible, but you are right that `#include "cinder/gl/gl.h"` without the other gl includes  is much nicer.

richardeakin:
The missing includes have been addressed with ccba1a2828705.

PR #785, opened 2015-04-02
Add MayaCamUI overloads that take MouseEvent

richardeakin:
This is in order to add support for pan and zoom to touchpads such as mac laptops, which otherwise require the user to add custom mappings.

The `MayaCamUI::mouseDown( MouseEvent )` overload isn't particularly needed, I added it for symmetry.


andrewfb:
I pushed this changeset and a couple of related commits to glNext_mayaCam

PR #790, opened 2015-04-02
ScopedVao const correctness.

sansumbrella:
Receive a `const VaoRef &` in constructor.

### DIFF
--- a/include/cinder/gl/scoped.h
+++ b/include/cinder/gl/scoped.h
@@ -38,7 +38,7 @@ typedef std::shared_ptr<class Renderbuffer>		RenderbufferRef;
 
 struct ScopedVao : private Noncopyable {
 	ScopedVao( Vao *vao );
-	ScopedVao( VaoRef &vao );
+	ScopedVao( const VaoRef &vao );
 	~ScopedVao();
 
   private:
--- a/src/cinder/gl/scoped.cpp
+++ b/src/cinder/gl/scoped.cpp
@@ -39,7 +39,7 @@ ScopedVao::ScopedVao( Vao *vao )
 	mCtx->pushVao( vao );
 }
 
-ScopedVao::ScopedVao( VaoRef &vao )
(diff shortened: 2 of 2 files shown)


andrewfb:
Hey - this is actually by design to prevent a user from accidentally taking a temporary by reference. Under the hood we take the raw pointer of the VAO (to reduce the overhead of atomics stored on the vector of VAOs on Context).

richardeakin:
I'm confused, how would passing by const& create a temporary, but non-const& wouldn't? Also, [Context::pushVao()](https://github.com/cinder/Cinder/blob/glNext/include/cinder/gl/Context.h#L108) is taking a const VaoRef & argument (there is also a variant that takes a raw pointer, but it is not used by ScopedVao and seems irrelevant to the argument of const correctness).

andrewfb:
Doesn't create a temporary - but prevents taking a temporary by reference. Without it, `ScopedVao scp( makeMyVao() )` would be legal, but is dangerous since the VaoRef returned by `makeMyVao()` will not outlive the construction of the ScopedVao, and will likely crash mysteriously. Such a use case should be written as `gl::VaoRef vao = makeMyVao(); ScopedVao scpVao( vao );` and this arrangement (VaoRef& but not const VaoRef&) forces that.

richardeakin:
I see. As an aside, and probably a moot point because it would be quite wasteful, we could add r-value reference constructors to support something like `ScopedVao( Vao::create() )`. Perhaps better to make using temporary Vao's a bit difficult.

sansumbrella:
Gotcha. A bit annoying when making `draw()` methods const, since we need to store `VaoRef`. Would be good to document this design choice inline for the future.

richardeakin:
@sansumbrella's point seems to push the favor towards adding an additional && constructor. But then, we're talking Vao here, are we really gaining some performance by not storing a shared_ptr for the life of the RAII object?

PR #796, opened 2015-04-03
Fixed shader preprocessor for es2 / iOS

wdlindmeier:
This fixes glsl shader support on iOS devices which don't have an explicit version macro. 

• Prior to this change, shaders would not compile on iOS because CINDER_GL_ES is defined in cinder/gl/Platform.h, not cinder/app/Platform.h, and would therefore default to version 150, which is not supported.

• Changed the default ES_2 version to 100, as 110 is not supported on iOS.

• Removed "es" from the version macro in ES_2 (I believe this is ES_3 syntax?)

### DIFF
--- a/src/cinder/gl/ShaderPreprocessor.cpp
+++ b/src/cinder/gl/ShaderPreprocessor.cpp
@@ -23,6 +23,7 @@
 
 #include "cinder/gl/ShaderPreprocessor.h"
 #include "cinder/app/Platform.h"
+#include "cinder/gl/Platform.h"
 #include "cinder/Utilities.h"
 #include "cinder/Log.h"
 
@@ -45,7 +46,7 @@ ShaderPreprocessor::ShaderPreprocessor()
 #if defined( CINDER_GL_ES_3 )
 	mVersion = 300;
 #elif defined( CINDER_GL_ES_2 )
-	mVersion = 110;
+	mVersion = 100;
 #else // desktop
 	mVersion = 150;
 #endif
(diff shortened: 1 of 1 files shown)


src/cinder/gl/ShaderPreprocessor.cpp
```diff
@@ -85,9 +86,9 @@ std::string ShaderPreprocessor::parseDirectives( const std::string &source )
 	
 	// if we don't have a version yet, add the default one
 	if( version.empty() ) {
-#if defined( CINDER_GL_ES )
-		version = "#version " + to_string( mVersion ) + "es\n";
-#else
+#if defined( CINDER_GL_ES_3 )
+		version = "#version " + to_string( mVersion ) + " es\n";
+#elif defined( CINDER_GL_ES_2 )
```
richardeakin: I think you're right on the point that es2 doesn't use " es" suffix, but this looks like it doesn't cover desktop anymore. I think the `#elif defined( CINDER_GL_ES_2 )` should just be and `#else`.

richardeakin:
Good points. I'd like to merge this pending the change above, except that it looks like you've committed the old cinder submodule references too. Can you make a new PR (or push force to this one) with a clean commit?  Else, if you don't mind, I'll just hand-cherry-pick your changes..

wdlindmeier:
@richardeakin Thanks. Made those changes.

richardeakin:
Awesome.  I did a squash merge and pushed to _glNext_.

PR #800, opened 2015-04-03
added case for closed polyLine in draw

### LINKED ISSUE #794: glnext, void draw( const PolyLine2 &polyLine )  does not draw closed
If I mark the polyLine as closed, the draw function still draws it as an open polyline

Ahbee:
fixes #794


richardeakin:
Thanks!

Issue #773, opened 2015-03-28
TriMesh::write without texture coordinates


gaborpapp:
`TriMesh::write` fires an assert in glNext if the `TriMesh` has no texture coords in https://github.com/cinder/Cinder/blob/glNext/src/cinder/TriMesh.cpp#L460
although texcoords should not be required.

richardeakin:
Looks like TriMesh read() / write() aren't as flexible as glNext's TriMesh yet. What if instead of storing only the number elements, we also store the dims? Ex, the first 8 numbers in the file would be:

numPos, sizePos, numNormals, sizeNormals, numTexCoords0, sizeTexCoord0.

I don't see why we shouldn't read and write texCoords1-3 as well..

Then we'd be able to remove those asserts.

gaborpapp:
Good suggestion. It would be nice to support at least as much as ObjLoader, so read obj's can be written with TriMesh::write() easily.

richardeakin:
In addition, I propose we add an optional argument to the write() method that will enable writing tangents / bitangents:

```
void	write( DataTargetRef out, bool writeTangents = false ) const;
```

It seems that if the usefulness of this method is to optimize load times, users could benefit from not having to calculate tangents / bitangents again a the cost of extra memory.

richardeakin:
I've begun work on this in the following branch: https://github.com/richardeakin/Cinder/tree/glNext_trimesh_io

Still need to add some test code and optional writeTangents.

gaborpapp:
Great, thanks!

richardeakin:
Fixed in #832.

PR #802, opened 2015-04-03
TriMesh::read() / write() improvements to support all dims

richardeakin:
Fixes issue #773.

Test added at test/TriMeshReadWriteTest, though obj file is missing as it is awaiting an update.


paulhoux:
Instead of throwing on version, couldn't the ```read()``` method support the old version, too?

richardeakin:
Thanks for taking a look. We could, but we would have no way to discern between meshes created with 0.8.6 or the version currently in glNext branch, since they both use version number 1 and they are composed differently. Also, the glNext version was pretty minimal and rigid, it was never production ready.

I'm open to the opinions of others, though personally I'd prefer to ditch it (keep the throw) and upgrade any existing binary files, to avoid having some legacy, development only code laying around.  Unfortunately this would mean re-exporting existing .msh files, the one in NormalMapping sample is the only one I can think of within cinder itself.

richardeakin:
Ah, actually I see the current glNext format is compatible with 0.8.6, so my first point is a moot one.  As I said, I'm open to what others think, I can re-add the legacy read code back if preferred. Though I think we should still throw if the first uint8_t isn't 1 or 2.

paulhoux:
I agree on throwing if version number is invalid. Personally, I'd love to stay backward-compatible if possible, it should be pretty easy to keep support for version 1 MSH files. No need adding support for writing them, though. Thanks for considering it, Rich.

richardeakin:
Paul, how was [leprechaun.msh](https://github.com/cinder/Cinder/blob/glNext/samples/_opengl/NormalMapping/assets/leprechaun.msh) created?  It seems to have the first int8 equal to 2, but its format is for version 1.  I've updated the TriMesh::read() code to support both versions, but leprechaun.msh can't be read because of this.

richardeakin:
This is pretty much ready, other than the above issue mentioned related to opengl NormalMapping sample - that won't work unless we update the .msh one way of another.

I suppose the only other thing to consider is whether we want to make normals optional and thereby produce smaller file sizes, since they can also be regenerated.  This wouldn't work with the current file format, we'd have to store the float arrays interleaved in order for both normal and tangent data to be optional.

richardeakin:
After talking with Paul online, looks like we'll need to generate a new .msh file for the NormalMapping sample, and I know how to do it from the existing one.

And we figure that we might as well make storing normals optional, and make the change to the format to be the following to allow both optional normals and tangents:

* numPositions
* sizePositions
* position array
* numNormals
* sizeNormals
* normal array
* ...

Can't make these changes tonight but I will soon.

paulhoux:
Ola,

I would suggest adding identifiers, so we can later change the order of the data blocks if we want to:

* 'identifier: positions'
* numPositions
* dimensionPositions
* position array (densely packed)
* 'identifier: normals'
* numNormals
* dimensionNormals
* normal array (densely packed)
* ...

richardeakin:
Pushed the proposed branch to the main repo so paul could work on this, as he kindly offered to improve the serialization format so that it is more flexible. We'll create a new PR for that when it is ready.

paulhoux:
I've pushed an update to the code, see: #c77b6ed783acde44dea3f1e19d2d3642675198d2

Issue #787, opened 2015-04-02
DeferredShading (basic) sample problems with Mac Intel Iris Pro card


richardeakin:
When forcing the mbo's integrated graphics card, I see the following:

![screen shot 2015-04-02 at 10 16 35 am](https://cloud.githubusercontent.com/assets/5132127/6965665/613f220c-d921-11e4-8d2f-c7cd5677a4c7.png)

And debug view:
![screen shot 2015-04-02 at 10 16 56 am](https://cloud.githubusercontent.com/assets/5132127/6965673/6cd3bd9e-d921-11e4-96f6-e7440ad27fa5.png)

Also reported in [this thread](https://forum.libcinder.org/topic/glnext-deferred-shading-samples#23286000002299087). @gaborpapp seems to be having problems with it on his dedicated NVidia GT 750M card, though I have that exact same model and it works for me. However one thing I did notice, was that if I forced the integrated card, and then force switch it to dedicated, things still look wrong. But if I fire up the sample with just dedicated (or using dynamic switching), it renders fine.

richardeakin:
minor update: [this](https://github.com/cinder/Cinder/blob/glNext/samples/_opengl/DeferredShading/src/DeferredShadingApp.cpp#L306) call to `mGlslProgLBuffer->uniformBlock( 0, 0 )` is causing an `GL_INVALID_VALUE` error with the Intel Iris Pro only. I didn't see a std140 uniform block in lbuffer.frag so I thought to comment it out - the error goes away but the rendering is still off..

BanTheRewind:
Started a branch for fixes here.
https://github.com/BanTheRewind/Cinder/tree/glNext_DeferredFixes

BanTheRewind:
This PR addresses this issue.
https://github.com/cinder/Cinder/pull/803

richardeakin:
Unfortunately I'm still seeing rendering problems, though it looks a little different:

![screen shot 2015-04-06 at 1 16 43 pm](https://cloud.githubusercontent.com/assets/5132127/7008311/4a96b394-dc5f-11e4-9e9c-9fbb30258a15.png)

![screen shot 2015-04-06 at 1 16 54 pm](https://cloud.githubusercontent.com/assets/5132127/7008316/52be905a-dc5f-11e4-8cbc-2221917bc014.png)

Let me know if there is anything I can do to help you debug, though I haven't had time to look at the example in detail..

BanTheRewind:
[...]

PR #803, opened 2015-04-03
DeferredShading sample optimizations

BanTheRewind:
Bug fixes and improved performance.


BanTheRewind:
Don't merge juuuuust yet. Working on instances drawing for the lights and shadow casters to eek out a couple or few more FPS.

richardeakin:
I think this is not ready to merge still? I just tried running DeferredShading sample and noticed #787 issue persists, also DeferredShadingAdvanced sample is throwing an error in the DOF fragment shader:

```
Depth of field: GLSL Error: FRAGMENT: ERROR: 0:31: Incompatible types (float and vec4) in assignment (and no available implicit conversion)
ERROR: 0:45: Incompatible types in initialization (and no available implicit conversion)
...
```

BanTheRewind:
I wonder if I just accidentally left the updated version of the DoF shader out of the commit. Will check when I get home tonight. 




This update doesn't fix the scrambled image issue (which I haven't been able to reproduce to debug, unfortunately), but the advanced sample runs about 20-25% faster with nicer looking shadows. It also removes any code from the basic sample left over from the advanced (ie, loading the "random" texture, setting a uniform block binding, etc).

BanTheRewind:
Shader issue on the advanced sample fixed. Just needed to commit files. Didn't realize they'd been left out.

BanTheRewind:
Fixed that bug. Just needed to set alpha to 1.0 in the point light fragment shader. That pass uses additive blending, so it makes sense that it was going wonky. The advanced sample doesn't have an alpha channel in the ping=pong FBO, which is why the problem doesn't happen there. I'm using the alpha channel the basic sample's G-buffer to store emission data to render the light sources.

BanTheRewind:
OK. Bug is fixed for the Iris integrated card now, too. I've also switched all the drawing routines to use batches. This makes the ::draw() method a little more compact and easier to read.

richardeakin:
Yep, iris bug fix confirmed. Also, I'm seeing 60fps on the advanced sample - nice work!

PR #806, opened 2015-04-04
glNext checkerboard

andrewfb:
Adding support for rendering checkerboards via ip::checkerboard(). Utilizing this in ObjLoader sample, along with new head .obj file with tex coords.


pizthewiz:
This is a great addition!

Issue #635, opened 2014-12-14
getWindowPos() not updating after resizing window from top left


richardeakin:
[reported here](https://forum.libcinder.org/topic/getwindowpos-does-not-update-after-resizing-window-mac-os#23286000002221028).

Confirmed on Mac OS X / master branch, unconfirmed on windows.

richardeakin:
Fixed with #807

PR #807, opened 2015-04-05
Update AppImplMac.mm

lithium-snepo:
addresses https://github.com/cinder/Cinder/issues/635 on osx where the window position isn't updated when resizing from the top left

### DIFF
--- a/src/cinder/app/cocoa/AppImplMac.mm
+++ b/src/cinder/app/cocoa/AppImplMac.mm
@@ -628,9 +628,21 @@ using namespace cinder::app;
 	NSSize nsSize = [mCinderView frame].size;
 	mSize = cinder::ivec2( nsSize.width, nsSize.height );
 
+	NSRect frame = [mWin frame];
+	NSRect content = [mWin contentRectForFrameRect:frame];
+	
+	ivec2 prevPos = mPos;	
+	mPos = ivec2( content.origin.x, [[[NSScreen screens] objectAtIndex:0] frame].size.height - frame.origin.y - content.size.height );
+
 	if( ! ((PlatformCocoa*)Platform::get())->isInsideModalLoop() ) {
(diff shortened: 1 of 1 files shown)


src/cinder/app/cocoa/AppImplMac.mm
```diff
@@ -628,9 +628,21 @@ - (void)resize
 	NSSize nsSize = [mCinderView frame].size;
 	mSize = cinder::ivec2( nsSize.width, nsSize.height );
 
+	NSRect frame = [mWin frame];
+	NSRect content = [mWin contentRectForFrameRect:frame];
+	
+	ivec2 prevPos = mPos;	
+	mPos = ivec2( content.origin.x, [[[NSScreen screens] objectAtIndex:0] frame].size.height - frame.origin.y - content.size.height );
```
pizthewiz: Hmm, doesn't this assume that the window is on the first screen? It might be a better idea to get the screen from the window's `screen` property and the `NSGeometry.h` functions might be helpful too:
```Objective-C
mPos = ivec2( NSMinX( content ), NSHeight( mWin.screen.frame ) - NSMinY( frame ) - NSHeight( content ) );
```

lithium-snepo: I pinched the mPos calculation code from the window movement handler, so if this is wrong, it's wrong there too.

pizthewiz: Ahh, interesting - it should be pretty easy to test and find out if it is an issue, just move the window to a secondary display and then hit the relevant method.

lithium-snepo: I don't have a secondary display available at the moment so i'll have to check later. I think you're right, though.

richardeakin:
Thanks for this!

PR #811, opened 2015-04-09
Avoid NSWindow warning on contentView assignment

### LINKED ISSUE #584: OS X 10.10 emits warning when switching to fullscreen at startup
When adding `settings->setFullScreen( true );` to prepareSettings() in OS X 10.10, the following warning / stacktrace is printed (tested on master branch):

```
2014-11-14 19:32:52.227 windowTest[85547:5829855] NSWindow warning: adding an unknown subview: <NSView: 0x10645c0>
2014-11-14 19:32:52.244 windowTest[85547:5829855] Call stack:
(
	0   AppKit                              0x917fa776 -[NSThemeFrame addSubview:] + 108
	1   AppKit                              0x910a1d77 -[NSView replaceSubview:with:] + 152
	2   AppKit                              0x9192b4d7 __60-[NSView(NSFullScreenMode) enterFullScreenMode:withOptions:]_block_invoke + 178
	3   AppKit                              0x91191bd7 NSPerformWithScreenUpdatesDisabled + 97
	4   AppKit                              0x9192b3cd -[NSView(NSFullScreenMode) enterFullScreenMode:withOptions:] + 3147
	5   windowTest                          0x0007d00c -[CinderView setFullScreen:options:] + 1036
	6   windowTest                          0x000430e7 -[WindowImplBasicCocoa setFullScreen:options:] + 135
	7   windowTest                          0x00040d1b -[AppImplCocoaBasic applicationDidFinishLaunching:] + 1403
```

Though the message is massively ugly, it does appear to still work. @andrewfb tried using the NSFullScreenModeSetting setting, which he found in Apple's [OpenCL N-Body](https://developer.apple.com/library/mac/samplecode/OpenCL_NBody_Simulation_Example/Listings/Sources_View_OpenGLView_mm.html) sample, but this seems to stop full screen from working altogether.

pizthewiz:
Resolves #584 (or at least works around it).

When a window's content view is directly assigned, an odd warning is spewed to the console about an unknown subview being added to `NSThemeFrame`, which is occurring in AppKit - this can be sidestepped by instead adding the CinderView as a subview of the window's content view.

### DIFF
--- a/src/cinder/app/cocoa/AppImplMac.mm
+++ b/src/cinder/app/cocoa/AppImplMac.mm
@@ -818,8 +818,9 @@ using namespace cinder::app;
 															highDensityDisplay:appImpl->mApp->isHighDensityDisplayEnabled()
 															enableMultiTouch:appImpl->mApp->isMultiTouchEnabled()];
 
-	[winImpl->mWin setDelegate:self];	
-	[winImpl->mWin setContentView:winImpl->mCinderView];
+	[winImpl->mWin setDelegate:self];
+	// add CinderView as subview of window's content view to avoid NSWindow warning: https://github.com/cinder/Cinder/issues/584
+	[winImpl->mWin.contentView addSubview:winImpl->mCinderView];
 
 	[winImpl->mWin makeKeyAndOrderFront:nil];
 	// after showing the window, the size may have changed (see NSWindow::constrainFrameRect) so we need to update our internal variable


richardeakin:
Thanks piz!

andrewfb:
It looks like this change has broken window resizing on 10.9. I don't have immediate access to a 10.10 machine to test there but I've narrowed it down to this commit.

pizthewiz:
Hmm, perhaps best to revert the PR while we investigate. I wonder if the subview of content view introduced an issue with the autoresize mask then.

andrewfb:
OK - I've reverted it on glNext for the time being. To elaborate, the behavior is that the GL view stays the same size as the window resizes, remaining in the lower left corner.

richardeakin:
Confirming that rese is jacked on 10.10, same as described by Andrew above. It looks like we'll need an even more kludgy kludge..

PR #813, opened 2015-04-10
Provided additional getter functions for Texture classes...

paulhoux:
... ( e.g. getFormat() ) and for the Format itself. Also added gl::Context::isTextureBound() and gl::Context::isFramebufferBound(), which use the stack to determine if a texture/fbo is currently in use. Note: just like Fbo::getFormat(), the Texture::getFormat() function may not return a fully updated format structure; if the user calls functions like setMinFilter() on a texture, the format will not reflect the new setting.

### DIFF
--- a/include/cinder/gl/Texture.h
+++ b/include/cinder/gl/Texture.h
@@ -278,7 +278,6 @@ class TextureBase {
 		bool				mImmutableStorage;
 		GLfloat				mMaxAnisotropy;
 		GLint				mInternalFormat, mDataType;
-		//GLint				mDataFormat;
 		bool				mSwizzleSpecified;
 		std::array<GLint,4>	mSwizzleMask;
 		bool				mBorderSpecified;
@@ -290,9 +289,6 @@ class TextureBase {
 #endif
 		friend class TextureBase;
 	};
-
-	//! Returns the Texture::Format of this texture
-	virtual const Format& getFormat() const = 0;
 
   protected:
 	TextureBase();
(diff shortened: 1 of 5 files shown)


include/cinder/gl/Texture.h
```diff
@@ -269,7 +278,7 @@ class TextureBase {
 		bool				mImmutableStorage;
 		GLfloat				mMaxAnisotropy;
 		GLint				mInternalFormat, mDataType;
-		GLint				mDataFormat;
+		//GLint				mDataFormat;
```
richardeakin: Is this commented out because it is no longer used? If so should remove entirely.

richardeakin:
Should we store the Format in either case (including `Fbo`)?  We debated this same topic around storing App::Settings and came to the conclusion that it is cleaner to store whatever data we need from it and ditch the load-time temporary, in order to avoid duplication of data and the problems that come from that.

andrewfb:
Paul and I just discussed this for a while and I think we came to a similar conclusion Rich. We will stop storing the Format used for construction on both Texture and Fbo. However to accommodate the use-case that's most common, which is cloning a Texture/Fbo with the same or similar Format, we'll provide a getFormat() method that constructs a Format on the fly that reflects the current state of the Texture/Fbo. This should address probably the biggest flaw with the way it's done now, which is that state that was changed after construction is not reflected in getFormat().

paulhoux:
Ok, I've removed the ```Texture*::getFormat()``` methods for now.

include/cinder/gl/Texture.h
paulhoux: Yes, it appeared it was no longer used, but I didn't want to remove it altogether because I didn't know why it was there in the first place.

richardeakin:
I've moved the getFormat() to a new issue to be resolved separately, this looks good to me to merge.

PR #814, opened 2015-04-10
Ignore non-core CinderBlocks

pizthewiz:
My local Cinder repo state is usually a bit polluted with statuses of a bunch of unknown non-core CinderBlocks and this tweak will ignore them. One catch is that if a new block were added or renamed, it would need to be reflected in the ignore file as well - initially I thought this was going to be a big deal, but it doesn't appear as though a block has been added in quite some time.

While I'm not sure others will find this as useful as I do, I thought I'd at least propose it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,8 @@ ipch/
 # doxygen generated files
 docs/html
 docs/doxygen/cinder.tag
+
+# CinderBlocks
+blocks/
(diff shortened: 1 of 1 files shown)


ryanbartley:
I feel like this would be a great edition. I often have quite a bit of red in my git status because of blocks that I have in the repo.

andrewfb:
This is a nice solution that hadn't occurred to me. I'd be open to it myself.

richardeakin:
+1, I think the drawback you mentioned is minor and easy enough for maintainers to deal with.

BanTheRewind:
+1.

PR #816, opened 2015-04-12
sample updates for those who use MayaCamUI

richardeakin:
This is mainly to take advantage of the features introduced in #808.  Also added `MayaCamUI::setAspectRatio( float )`, which is many times easier to use than getting a copy of the `Camera`, updating its aspect ratio, then re-setting the new `Camera`. Some sample-specific notes:

* `rayMarcher`: this sample has been updated and seems to be working, but the way that it constructs the `RayMarcher` class (from RayMarcherApp's constructor) doesn't work so well anymore. It was causing an exc_bad_access, which I fixed by later resetting the Camera pointer. Perhaps this sample should be improved for the latest glNext. 
* `_opengl/ObjLoader`: I removed the optional `ArcBall` usage, instead only `MayaCamUI` is in use. For iOS, disabled multitouch to make this possible. Main reason was that the `ArcBall` vertical rotation was flipped, but otherwise did the exact same thing as mayacam left-click drag.

### DIFF
--- a/include/cinder/MayaCamUI.h
+++ b/include/cinder/MayaCamUI.h
@@ -162,6 +162,9 @@ class MayaCamUI {
 	
 	const CameraPersp& getCamera() const				{ return *mCamera; }
 	void setCurrentCam( const CameraPersp &currentCam ) { *mCamera = currentCam; }
+
+	//! Convenience routine to set the internal Camera's aspect ratio
+	void setAspectRatio( float aspectRatio )	{ mCamera->setAspectRatio( aspectRatio ); }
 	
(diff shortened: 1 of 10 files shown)


richardeakin:
@andrewfb I noticed your changes in [glNext_camChanges](https://github.com/cinder/Cinder/compare/glNext...glNext_camChanges) branch - after you merge that I can rebase this if you like. There are some additional sample updates that I'd like to get in here, though much of it is overlapping.

andrewfb:
Addressed with recent merge.

Issue #444, opened 2014-06-04
InterfaceGl enumerated parameters with Options missing


gaborpapp:
the new Options-style version of `void addParam( const std::string &name, const std::vector<std::string> &enumNames, int *param, const std::string &optionsStr = "", bool readOnly = false );` is missing. Most of the Options functions are not applicable here, but updateFn would be really useful.

richardeakin:
Added with #820

PR #820, opened 2015-04-15
enumerated parameters with callbacks

gaborpapp:
Added enumerated parameters with accessors for partly solving #444. Please revise.

### DIFF
--- a/include/cinder/params/Params.h
+++ b/include/cinder/params/Params.h
@@ -158,6 +158,9 @@ class InterfaceGl {
 	template <typename T>
 	Options<T>	addParam( const std::string &name, const std::function<void ( T )> &setterFn, const std::function<T ()> &getterFn );
 
+	//! Adds an enumerated param to the interface with no target, but is instead accessed with \a setterFn and \a getterFn. The input parameter of \a setterFn and the return value of \a getterFn correspond to the indices of \a enumNames. \return Options<T> for chaining options to the param.
+	Options<int> addParam( const std::string &name, const std::vector<std::string> &enumNames, const std::function<void ( int )> &setterFn, const std::function<int ()> &getterFn );
+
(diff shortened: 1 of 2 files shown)


richardeakin:
Thanks for this work.  As it currently stands, the `Options<int>` passed back is going to cause problems, since it maps to TW_TYPE_INT32 ([here](https://github.com/cinder/Cinder/blob/master/src/cinder/params/Params.cpp#L623)). The easiest solution would be to disallow the chained options for enum param /w callbacks, users would then have to use the optionsStr to do things like specify keyUp, keyDown, or group. This seems OK to me since its where we are currently at anyway.

The other, messier, solution would be to somehow store on the Options that it needs to be added with TwDefineEnum. We'd need to store the vector<string> on Options, too..

richardeakin:
Well I'll be tickled pink, this does work and works well.  I don't actually see any limitations here.

I've pushed a branch that adds an overload for use with updateFn(), as well as some sample code to ParamsBasic. Sweet!

gaborpapp:
Great! Thanks! And what is the situation with the TW_TYPE_INT32 mapping? Does this really get mapped and it works because int is 32-bit in OS X and Windows in 32 and 64-bit as well?

richardeakin:
On Mac it is built as 64-bit and working there, but yes we should test on both windows 32-bit and 64-bit before merging that branch.  Tell you the truth I'm not entirely sure how it works, I guess once you TwDefineEnum then ATB remembers that even when it is removed and re-added later from an `Options<T>`.

richardeakin:
Merged after testing here and working on both 32-bit and 64-bit windows. Now the new callback stuff should be pretty complete, which is great.

gaborpapp:
Awesome! Thanks Rich!

PR #821, opened 2015-04-15
avoid deprecated Gestalt function to fetch OS X version when possible

pizthewiz:
The `Gestalt` functions have been deprecated since 10.8 (and are said to incorrectly report 10.10 as 10.9, though I was unable to reproduce on 10.10.3) and API was added to `NSProcessInfo` to replace it. Though the API wasn't made public until OS X 10.10 / iOS 8, it is available on earlier releases (possibly backported when using newer base SDKs due to `Gestalt` being deprecated without a replacement) so the Objective-C runtime is used to divine availability and execute, falling back to `Gestalt` with the deprecation warning hushed.

The `NSProcessInfo` API is also available on iOS and is a bit cleaner than the `-[UIDevice systemVersion]` string parsing, so it is use when available, falling back to the `UIDevice` method.

So far, I have only tested on 10.10.3 / Xcode 6.3 building the OS X and iOS static libs, I'll fire up a VM and make sure things are good in 10.9 as well.

### DIFF
--- a/src/cinder/System.cpp
+++ b/src/cinder/System.cpp
@@ -32,8 +32,8 @@
 		#import <CFNetwork/CFNetwork.h>
 		#import <UIKit/UIDevice.h>
 	#endif
-    #import <Foundation/Foundation.h>
-    #import <objc/objc-runtime.h>
+	#import <Foundation/Foundation.h>
+	#import <objc/objc-runtime.h>
 	#import <netinet/in.h>
 	#import <netdb.h>
 	#import <ifaddrs.h>
@@ -450,9 +450,9 @@ int System::getNumCores()
 
 #if defined( CINDER_COCOA )
 typedef struct {
-    NSInteger majorVersion;
-    NSInteger minorVersion;
-    NSInteger patchVersion;
+	NSInteger majorVersion;
+	NSInteger minorVersion;
+	NSInteger patchVersion;
 } ShadowOSVersion;
 #endif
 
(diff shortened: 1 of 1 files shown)


pizthewiz:
I'll push a space to tab fix as well.

pizthewiz:
Building the OS X target on OS X 10.9.5 / Xcode 5.1.1 works, building the iOS device target does not:
```sh
/Users/pizthewiz/sandbox/Cinder/Cinder/src/cinder/System.cpp:36:10: 'objc/objc-runtime.h' file not found
```

The same thing occurs on 10.10.3 / Xcode 6.3, my previous test had a simulator selected instead of `iOS Device`. I'll get that sorted shortly.

pizthewiz:
`arm64` only exposes `objc_msgSend`, more footwork required.

paulhoux:
Unrelated, but it appears that cinder::asio is using ```::GetVersionExW()``` on MSW, which is also deprecated.

pizthewiz:
Ok, I've simplified things and limited the changes to only OS X as the situation on iOS is a bit strange at present.

:heavy_check_mark: OS X 10.10.3 / Xcode 6.3 OS X target
:heavy_check_mark: OS X 10.9.5 / Xcode 5.1.1 OS X target
:heavy_check_mark: OS X 10.8.5 / Xcode 5.1.1 OS X target
~~iOS Targets~~

---

I haven't done anything with Cinder and iOS but I was going to file an issue for this:
```sh
$  lipo -info lib/libcinder-iphone* 
Architectures in the fat file: lib/libcinder-iphone-sim.a are: armv7 arm64 
Architectures in the fat file: lib/libcinder-iphone-sim_d.a are: armv7 arm64 
Architectures in the fat file: lib/libcinder-iphone.a are: armv7 arm64 
Architectures in the fat file: lib/libcinder-iphone_d.a are: armv7 arm64 
```

The simulator binaries are `armv7` and `arm64`?!

pizthewiz:
@paulhoux - I am all for fixing deprecated API warnings, but I'm not super Windows savvy to recommend a change.

richardeakin:
@pizthewiz if you build cinder-iphone-sim with your phone still selected as the test target (even though you can't test building a static lib on a device), it will infer you need binaries for that device, instead of building for sim (x86 32bit)

pizthewiz:
Spun the iOS Simulator target bit out as #825.

pizthewiz:
Apparently the `Gestalt` functions can return 10.9.3 on 10.10.3 still: https://twitter.com/_vade/status/590892319169191936

andrewfb:
Thanks for this - tested on 10.10 and 10.9.

Issue #829, opened 2015-04-17
Drawing geom::Rect or geom::Circle causes crash


wdlindmeier:
Drawing a basic geom::Rect or geom::Circle crashes my app (OS X, Xcode 6.3). Drawing a geom::Sphere doesn't cause a crash. Here's a simple test case:

```c++
void TestApp::draw()
{
    gl::setMatricesWindowPersp( getWindowSize() );
    gl::GlslProgRef shader = gl::getStockShader( gl::ShaderDef().color() );

    // Crashes
    // geom::Rect source = geom::Rect( getWindowBounds() );

    // Crashes
    // geom::Circle source = geom::Circle().center( vec2( getWindowCenter() ) ).radius(100.f);

    // Doesn't crash
    geom::Sphere source = geom::Sphere().center( vec3( 0.f ) ).radius( 100.f );

    TriMesh::Format geomFmt = TriMesh::Format().positions().colors();
    TriMesh mesh(source, geomFmt);
    gl::BatchRef batchRect = gl::Batch::create( mesh, shader );
    gl::color(0, 1, 0);
    batchRect->draw();
}
```

richardeakin:
I can verify the crash, and you get a nice stacktrace by enabling guard malloc in Xcode:

```
* thread #1: tid = 0x36c663, 0x000000010007936a BasicApp`void cinder::geom::(anonymous namespace)::copyIndexDataForceTrianglesImpl<unsigned int>(primitive=TRIANGLE_STRIP, source=0x000012344c37dff0, numIndices=4, target=0x000012344c37fff0) + 266 at GeomIo.cpp:318, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x12344c380000)
    frame #0: 0x000000010007936a BasicApp`void cinder::geom::(anonymous namespace)::copyIndexDataForceTrianglesImpl<unsigned int>(primitive=TRIANGLE_STRIP, source=0x000012344c37dff0, numIndices=4, target=0x000012344c37fff0) + 266 at GeomIo.cpp:318
    frame #1: 0x000000010007924f BasicApp`cinder::geom::Target::copyIndexDataForceTriangles(this=0x00007fff5fbfb888, primitive=TRIANGLE_STRIP, source=0x000012344c37dff0, numIndices=4, target=0x000012344c37fff0) + 47 at GeomIo.cpp:417
[...]

PR #830, opened 2015-04-19
TriMesh from TRIANGLE_STRIP/FAN primitives

lithium-snepo:
Addresses https://github.com/cinder/Cinder/issues/829


andrewfb:
Nice fix - thanks for this.

PR #831, opened 2015-04-19
GlslProg/VboMesh Update

ryanbartley:
This is a somewhat large update to the inner working of GlslProg and a smaller update to the inner workings of Batch and VboMesh. Much of the public facing interface hasn't changed at all.

The highlights in GlslProg are:
- Clean up of caching variables. We now store Uniforms as a unified struct, same with attributes, uniform blocks, and transform feedback varyings.
- Uniforms now have a CPU-side cache so as to not re-buffer uniform values that haven't changed.
- We don't lazily initialize cache variables anymore. All of the above are cached during initialization so as to mitigate OpenGL get* operations.  
- Using templated type checking for uniform functions.
- Added multiple uniform type overloads.
- Added public facing getters for all of the new structs of info. Added public facing const pointer find methods for the new structs. 

The highlights in Batch and VboMesh are:
- Batch - removed initVao logic to centralize Buffer/GlslProg Vao creation inside vboMesh, in buildVao function.
- attribs that have a size greater than 4 now initialize a vao correctly. For instance, sending a mat4 up to the gpu. 

Also changed:
- Cloth Simulation sample updated. Found a bug with AMD GPUs that wasn't allowing two TFO's, which is in progress. However, I made the sample work on all platforms including AMD until by using gl::bindBufferBase until AMD fixes the problem.
- Changed ConstantStrings.h/.cpp to ConstantConversions.h/.cpp to include other conversions such as size of conversions.


include/cinder/gl/GlslProg.h
```diff
@@ -45,11 +44,49 @@
...
+	
+class UniformValueCache;
 
 class ShaderPreprocessor;
 
 class GlslProg : public std::enable_shared_from_this<GlslProg> {
   public:
+	
+	//! Resembles all the information Queryable of an attribute
+	struct Attribute {
```
richardeakin: If these structs are going to be public, we should provide accessors instead of exposing public ivars with m prefix. Also, they look quite redendant, seems that it could be minimized by combining most of this into a base class (not sure what you'd call it..).

richardeakin:
Looks like some nice features. I haven't yet gotten through the entire changeset but I did notice that there are icon imageas added for  `test/_opengl/IntegerAttribTest` and `samples/_opengl/ClothSimulation/` - could you remove those and reference the ones from samples/data instead?

src/cinder/gl/GlslProg.cpp
```diff
@@ -137,13 +181,49 @@ void GlslProg::Format::setShaderSource( const std::string &source, std::string *
 
 GlslProg::Format& GlslProg::Format::attrib( geom::Attrib semantic, const std::string &attribName )
 {
-	mAttribSemanticMap[attribName] = semantic;
+    bool exists = false;
```
richardeakin: Looks like this was formatted with spaces instead of tabs (you can see in xcode by going to Editor -> Show Invisibles).

src/cinder/gl/GlslProg.cpp
```diff
@@ -22,20 +22,65 @@
...
+static GlslProg::AttribSemanticMap	sDefaultAttribNameToSemanticMap;
+	
+class UniformValueCache : cinder::Noncopyable {
+  public:
+	UniformValueCache( uint32_t bufferSize )
+		: mBuffer( new uint8_t[bufferSize] ), mBufferSize( bufferSize )
+	{
+	}
+	
+	void insertBeginningByte( uint32_t value ) {
```
richardeakin: opening function bracket should be on new line

src/cinder/gl/GlslProg.cpp
```diff
@@ -426,6 +566,219 @@ void GlslProg::link()
...
+			// So they've told us at setup what to look for. Find the varying, they told us
+			// about so that we can finish the setup.
+			auto varying = findTransformFeedbackVaryings( name );
+			if( varying ) {
+				varying->mCount = count;
+				varying->mType = type;
+			}
+		}
+		
+		// TODO: This needs to be created correctly and checked, this is where the magic will happen.
```
richardeakin: Is this still a necessary TODO?

src/cinder/gl/GlslProg.cpp
```diff
@@ -457,573 +810,870 @@ std::string GlslProg::getShaderLog( GLuint handle ) const
...
+		return checkUniformValue( uniform, &val, 1 );
 	}
-	else
-		glUniform2fv( loc, count, &data[0].x );
 }
 
-// vec3*, count
-void GlslProg::uniform( int location, const vec3 *data, int count ) const
+template<typename T>
+string GlslProg::cppTypeToGlslTypeName()
```
richardeakin: I think this would be faster if a precomputed map were created and cached. This can be done by using something from `typeid()` as the key, and the glsl equivelant as the value. Maybe something like (unchecked):

```cpp
template<typename T>
string GlslProg::getGlslTypeName()
{
	static unordered_map<type_info, const char*>  sTypeNameMap;
	if( sTypeNameMap.empty() ) {
		sTypeNameMap[typeid( bool )] = "bool";
		sTypeNameMap[typeid( vec2 )] = "vec2";
		...
	}

	auto it = sTypeNameMap.find( typeid( T ) );
	CI_ASSERT( it != sTypeNameMap.end() );

	return it->second;
}
```

Even better would be to pass in the `type_info` as an argument variable to the method, since this would mean we didn't need extra templated versions of this method for each glsl type (there would just be one).

Concerning the case of failure, I think we should either assert or throw, since this should never happen unless we've made an oversight as to what types GlslProg can handle.

Lastly, I think the if should be indented the same as the following else if's..

richardeakin:
Finally to throw this out there, what is the rationale behind caching uniforms?  It seems sort of against other decisions made in cinder code base to introduce complexity of this nature to fix user mistakes, such as updating uniforms when they haven't changed locally.

src/cinder/gl/GlslProg.cpp
richardeakin: As an alternative, [here's a way](https://github.com/facebook/folly/blob/master/folly/dynamic.cpp) to do it with macros (from facebook's `folly::dynamic` object).

src/cinder/gl/GlslProg.cpp
ryanbartley: I'll fix that.

include/cinder/gl/GlslProg.h
ryanbartley: I can give accessors. I was thinking about a base class as well. Maybe GlslProgInfo that holds the redundant stuff?

src/cinder/gl/GlslProg.cpp
ryanbartley: I'll change that.

src/cinder/gl/GlslProg.cpp
ryanbartley: It isn't going to be tackled with this PR. I could probably remove some of the psuedo code. However, this would be where Cinder could make connections between output (transform feedback) buffers and input (attribute) buffers. There's quite a bit of checking we can do, similar to the checks on Uniform or the AttributeMap check inside of buildVao, that could give feedback to the user if they've set things up incorrectly. I don't have a problem removing it though.

src/cinder/gl/GlslProg.cpp
ryanbartley: I personally like the compile time checking that the templated version provides, in the case that more functions and more types are added in the future. It's symmetrical with the other function "checkUniformType", but I'm not opposed to the change.

src/cinder/gl/GlslProg.cpp
richardeakin: I see. I'd say removing the TODO / psuedo code from GlslProg.cpp will make the implementation easier to read (and also won't suffer from going out of sync). Could be a nice feature addition later on, though.

src/cinder/gl/GlslProg.cpp
richardeakin: This isn't a compile time check; although there are versions of `cppTypeToGlslTypeName<T>` created at compile time for each type, the check goes through (worse case) all of the if's at runtime  whenever this is called. You *can* make it compile time safe with template specialization but the way to handle the 'unknown' case (the non-specialized one) is pretty messy, I can't remember how to do it at the moment. IMO caching a static map works okay here.

ryanbartley: Actually, you're totally right! We started with a compile time check system. It was definitely messy. I think doing this would be fine. And it may be better passing type_info in for future proofing.

ryanbartley:
@BanTheRewind, I actually made some changes to the way it populates the vao, to allow for arbitrary types as well as arrays of arbitrary types. Do you think you could pull and test again? I just want to make sure.

BanTheRewind:
Will do.

BanTheRewind:
Still works. :) You can try it in my branch here, if interested:
https://github.com/bantherewind/Cinder/tree/GlslProgRebase

I have a closed source prototype going which uses TF to update per-instance attribute data also using this branch. Still working there, as well.

ryanbartley:
Great. Thanks so much for the tests!

BanTheRewind:
np

ryanbartley:
With the last commit, we should be ready for merge. Let me know if there's anything else needed.

BanTheRewind:
Woohoo! Thanks for the work here. Uniform caching and arbitrary dim attributes are awesome features that the user might not notice. These are huge. I noticed. :)

BanTheRewind:
Can you think of any reason why the last update would cause ciColor to only contain a value for the green channel on OSX? Windows and iOS are fine. I merged this update into my repo, and suddenly in both deferred samples, only the green channel in ciColor is coming through on OSX.

I can confirm that it isn't a case of a mismatched attribute, because I can use ci::color() to control the brightness of green.

ryanbartley:
Hmm. Not sure. I'm headed home now. Let me check when I get there.

On Monday, April 27, 2015, Stephen Schieberl <notifications@github.com>
wrote:

> Can you think of any reason why the last update would cause ciColor to
> only contain a value for the green channel on OSX? Windows and iOS are
> fine. I merged this update into my repo, and suddenly in both deferred
> samples, only the green channel in ciColor is coming through on OSX.
>
> I can confirm that it isn't a case of a mismatched attribute, because I
> can use ci::color() to control the brightness of green.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cinder/Cinder/pull/831#issuecomment-96814169>.
>


-- 
Ryan Bartley
214-641-7788

BanTheRewind:
Hey, sorry, took the comment down for a sec to sanity check. So, yeah. Only on OSX. I'm not getting a red or blue channel values from ciColor. Windows and iOS are both fine. If I set the color to gray( 0.1 ), I get a very dark green, so it's the right attribute. Weird. It's possible it's something in my sample, but it worked fine until the latest merge, and that it only happens in OSX.

frame397
frame800
frame659

BanTheRewind:
Actually, I'm getting inconsistent result from setting color, so it looks the attribute data coming in to ciColor could be from somewhere else. Hmmmm...

ryanbartley:
then I'd say it's probably an offset problem. What branch are you working from, because it isn't happening on the one currently in glNext?

BanTheRewind:
I've been merging from your GlslProgRebase branch so I can use mat4 per-instance attributes in the advanced sample.




The green is now black. If it's an offset problem, it would appear on other OSes, right? It's only happening on OSX.

ryanbartley:
hmm.. Yeah it should be happening on other OSes but I have seen times where there's an error on one card that is checked on another card. Can you by chance test it on another OSX machine or I could download the branch and test on my machine if you let me know where?

ryanbartley:
Are you sure it's a problem with ciColor and not with Material or Light?

BanTheRewind:
Sure: https://github.com/bantherewind/Cinder/tree/GlslProgRebase

BanTheRewind:
I'm pretty sure. The basic sample doesn't use any UBO. The entire G-buffer is green or black (as it reads ciColor). If use white in the GBuffer instead of ciColor, everything appears correctly.

ryanbartley:
cool. I'm cloning now.

BanTheRewind:
I know OSX is extra picky about clearing. I just can't find where I'm not following the rules. The fact that it happened to both apps when I hadn't done any work on the apps in a bit is telling.

ryanbartley:
Just as I thought, I'm not getting green. UGGGGHHHH!!!
screen shot 2015-04-27 at 5 44 18 pm

BanTheRewind:
Hey, that looks great! ;) Frame rate isn't so great.  :/  This machine has a GeForce GTX 680M. What are you on?

BanTheRewind:
Wait, are you on the GlslProgRebase branch for sure? Fog should be disabled by default. Or did you just turn it on?

ryanbartley:
AMD Radeon HD 6770M

ryanbartley:
Whoops. Switching over.

BanTheRewind:
Ah. It's occurring to me that in the advanced sample, only the cylinder (room) is green (or black). Though everything is green in the basic sample.

ryanbartley:
Works on GlslProgRebase as well. however, I'm getting two warnings that I'm checking on now.

BanTheRewind:
I believe those are warnings about ciColor being in the shader, but not in the geometry. Both are spheres used for rendering lights.

BanTheRewind:
No warnings in the basic sample. I work with that one first since it's far less involved. You can also test it on iOS to see it working correctly.

ryanbartley:
Yeah. I'm not seeing green here. I tried both. Also, this is super weird but the warning was for "gl_InstanceID" as an attribute. I need to look into that and make sure that we're not pulling that for some reason when querying the attribs.

BanTheRewind:
Ah, weird. I couldn't get the attribute name to come up for me in the warnings. I assumed ciColor since sphere wasn't providing it. Well, instanced drawing is working and the basic sample doesn't use instancing.

It might help if I set this computer on fire.

ryanbartley:
HAHA. It's weird that we'd be pulling down gl_InstanceID. I just want to make sure that that's happening. Also, it may not be a bad idea to puke your glsl to console. I've updated the ostream operator and it should tell you everything that you have. I should probably provide types if I'm not already doing that. I have to run out really quick but I'll be back in a bit.

BanTheRewind:
Will give that a shot real quick.

BanTheRewind:
Yeah, everything is in there. Weird.

BanTheRewind:
I did notice that there is no semantic for gl_InstanceID, which would explain why it leaves the attribute name blank in the warning.

The G-buffer shaders look fine in the <<. But that makes sense because they totally work fine on everything but this machine. Ugh...

BanTheRewind:
Thinking about that issue more... in the cases of those warnings, I'm not using the instance ID in the shader at all. I have a VBO of mat4 per-instance attributes. No need for the ID.

ryanbartley:
I'm not sure if you have it undefined but gl_InstanceID is in gbuffer.vert and ibuffer_light.vert

BanTheRewind:
Oh, wait, you're right. I am using in there. But yeah... no semantic comes up for it. Not sure if that means anything.

BanTheRewind:
So if I tell the cylinder in the advanced app to use colors...

```
geom::Cylinder().subdivisionsAxis( 5 ).subdivisionsHeight( 1 ).colors()
```

The colors show up. So it appears that on OSX, for me, anyway... the color on the stack is ignored... ?

BanTheRewind:
Furthermore... I'm realizing color on the stack is actually being ignored across the board. Colors from lights and materials are OK. There were enough reflections on the spheres and icosahedrons to make it look like they were being colored, but they aren't.

ryanbartley:
Wow, that's a problem that I didn't expect. Also, gl_InstanceID doesn't have a defined semantic within Cinder. I'll change the way that stuff is logged as I've never seen gl_InstanceID pulled down in that way before.

BanTheRewind:
Hmmm... may have spoken too soon. Looking at debug mode, sometimes somethings are colored. Sometimes they aren't. Sorry if this turns out to be a weird runaround. Going to double check my driver situation.

ryanbartley:
I wonder if there's some error where calling gl::vertexAttrib4f in setDefaultShaderVars on the color attribute is mucking up the way that the vertexAttribPointer stuff is working.

ryanbartley:
are you getting any errors at run time using CI_CHECK_GL()?

BanTheRewind:
Checking everywhere. No GL errors. Keeping at it..

BanTheRewind:
Tell you what. Let's call this one good. I can't reproduce the error on my other Mac. No GL errors anywhere (I checked after every line of every GL operation in the basic sample). Nothing is leading to a problem in Cinder or the app (especially since it works great everywhere else). If this branch is merged, I can PR these samples and that will get it in front of a few more eyes to confirm or deny this issue.

At this point I have to believe it's a driver or other bug specific to this machine.

ryanbartley:
I'm going to be in front of a nvidia machine in a bit. I'll run it on that
one too.

On Tuesday, April 28, 2015, Stephen Schieberl <notifications@github.com>
wrote:

> Tell you what. Let's call this one good. I can't reproduce the error on my
> other Mac. No GL errors anywhere (I checked after every line of every GL
> operation in the basic sample). Nothing is leading to a problem in Cinder
> or the app (especially since it works great everywhere else). If this
> branch is merged, I can PR these samples and that will get it in front of a
> few more eyes to confirm or deny this issue.
>
> At this point I have to believe it's a driver or other bug specific to
> this machine.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cinder/Cinder/pull/831#issuecomment-97143129>.
>


-- 
Ryan Bartley
214-641-7788

ryanbartley:
Found the problem. Fixing it. Should be ready soon.

BanTheRewind:
Oh, awesome! Good work!

ryanbartley:
Alright. So here's a screen shot from my computer AMD GPU, and the app looks this way on Andrew's Nvidia GPU and Intel GPU for both glNext and GlslProgRebase.
screen shot 2015-04-29 at 1 09 00 am
it looks to be something weird with the lights.

ryanbartley:
Can you detect this on your machine?

BanTheRewind:
Weird. Looks like discard isn't getting called for pixels outside the light cube radius.

ryanbartley:
Any reason you think it would be disregarding the discard? Do you see this same behavior on your machines?

BanTheRewind:
I'll be on the "problem Mac" in about an hour and will check then. Not seeing this issue on Windows or iOS.

BanTheRewind:
Either that or culling or isn't working? It basically looks like it's rendering out to the corners of the cube when it shouldn't be.

ryanbartley:
Also just for reference I saw this on my AMD on glNext and GlslProgRebase

BanTheRewind:
I merged in your changes and all is well here. Do you want to pull the latest from my GlslProgRebase branch and give it a shot? Try both the basic and advanced samples and see if this problem persists. Discard is being called OK. Culling should be OK, too. Looks like something weird with alpha?

ryanbartley:
Looks good from here. 
screen shot 2015-04-29 at 3 21 50 pm
screen shot 2015-04-29 at 3 23 30 pm
@andrewfb do you want to try it on your mac?

BanTheRewind:
Woohoo! Oddly enough, I can sort of get the effect on iOS if I zoom way, way out, but I think that is related to depth precision and not actually a bug (something that would be optimized in a *real* project -- or the camera distance would be clamped).

BanTheRewind:
I just realized that I need to calculate the normal matrix per instance. I did it in the shader and everything looks right. But to be more efficient, I made the normal matrix a per instance attribute which is calculated with the model matrix in update. However, it appears that it isn't taking. Model matrix still works fine.

Here's how I'm setting it up:
https://github.com/BanTheRewind/Cinder/blob/GlslProgRebase/samples/_opengl/DeferredShadingAdvanced/src/DeferredShadingAdvancedApp.cpp#L417-L444

I tried just passing the normal matrix off the stack to sanity check, and it just looks like that attribute isn't working at all.

Have you tried setting up interleaved attributes that each had dims greater than 4? Maybe interleaved isn't working here?

frame2005

ryanbartley:
I'll take a look. Is there a reason you're using the AttribMap? Is the Glsl not finding the attributes correctly?

ryanbartley:
I'm getting build errors that it's missing Model constructor and other Model functions.

BanTheRewind:
Because this is how I know to set up batches with custom attributes. Recommendations on a better way?

BanTheRewind:
Doh, let me get that in the repo. Thought I added it.

ryanbartley:
no problem. I found it. Also, I understand why you used it now. And lastly, it looks correct to me...
screen shot 2015-04-30 at 5 23 43 pm

ryanbartley:
Actually, I'm getting warnings and errors, the warnings, I'll fix. Let me look into it.

|warning| void cinder::gl::VboMesh::buildVao(const cinder::gl::GlslProg *, const AttribGlslMap &)[474] Batch GlslProg expected an Attrib of USER_DEFINED but vertex data doesn't provide it.
|warning| void cinder::gl::VboMesh::buildVao(const cinder::gl::GlslProg *, const AttribGlslMap &)[474] Batch GlslProg expected an Attrib of USER_DEFINED but vertex data doesn't provide it.
|error  | void cinder::gl::VboMesh::buildVao(const cinder::gl::GlslProg *, const AttribGlslMap &)[422] CUSTOM_1's BufferLayout defined dimensions(12) does not match shader expected dimensions(9) for vInstanceNormalMatrix, skipping attribute
|error  | void cinder::gl::VboMesh::buildVao(const cinder::gl::GlslProg *, const AttribGlslMap &)[422] CUSTOM_1's BufferLayout defined dimensions(12) does not match shader expected dimensions(9) for vInstanceNormalMatrix, skipping attribute

BanTheRewind:
Aw, snap! You're right. Duh!

frame171

BanTheRewind:
It's been a long few days.

ryanbartley:
Did you get those warnings?

BanTheRewind:
I switched to debug mode and got them. I feel like something happened recently where just nothing dumps to the console anymore in release. Also, i didn't notice I was running in release.

Just pushed the update. Check it out. Having the correct normals per instance makes a huge difference. Not just in light reflecting, but it corrects some subtle AO issues.

ryanbartley:
Yeah, it looks amazing on my side.

BanTheRewind:
OK, let's merge this thing already so I can finally PR my sample updates.  :D

ryanbartley:
I'm good with it now.

ryanbartley:
Now this should be ready.

BanTheRewind:
After merging in your latest...  :/

frame161

ryanbartley:
I rebased. Can you switch over to glNext and see if the problem exists there? We were seeing that on both glNext and GlslProgRebase.

ryanbartley:
I don't see it.
screen shot 2015-04-30 at 7 50 27 pm
I'm going to clean everything and run it again, just to make sure.

BanTheRewind:
OK. I just rebuilt Cinder after the merge and that's when it showed up.

ryanbartley:
After cleaning and rebuilding both Cinder and the app, I get this.
screen shot 2015-04-30 at 7 54 49 pm
Looks fine to me.

BanTheRewind:
OK, will clean and rebuild both here.

BanTheRewind:
Nope, still seeing it. Hmmm...

BanTheRewind:
Advanced does it now, too.

Anything change with alpha or additive blending?

ryanbartley:
No, nothing of those have changed. The only change was a rebase. Can you push so I can see it here?

BanTheRewind:
Just pushed. I backed up my sample projects, deleted my local GlslProgRebase branch, updated glNext from upstream, then created a new GlslProgRebase branch into which I merged your repo. Then I copied in my projects. So this should be all up to date and clean now.  :)

ryanbartley:
I can see it here too. What do the volumes represent? Which shader is the culprit?

BanTheRewind:
lbuffer_frag in basic and lbuffer_light.frag in advanced. The light volumes are drawn using cubes. No changes to this shader in forever and I've been using the math for the falloff successfully forever. It looks like a blending issue.

BanTheRewind:
Stepping out for a few.

Yeah, there's nothing new in the project other than the CamUI. Hmmm...

ryanbartley:
Was there anything that you changed when we saw this behavior a couple of days ago?

BanTheRewind:
No, it just kind of automagically fixed itself it seemed.

ryanbartley:
do you see this behavior on OSX only?

BanTheRewind:
No, those screens are on Windows.

ryanbartley:
@paulhoux Can you think of anything that might be causing the discard not to take place?

ryanbartley:
So I un-rebased and I'm back to normal. 
screen shot 2015-04-30 at 9 41 31 pm
This is especially hard for me to track down as I have no idea where these volumes are being calculated. If you can tell me how these volumes are being calculated I may be able to help but it seems to be a value off. So the most important thing is to figure out what that value is that isn't working.

BanTheRewind:
It's in lbuffer.frag. The condition which calls discard and further down, the falloff equation (the longest line of code) are important. But note that these both started breaking after the rebase, and both worked fine before, without any sample code changes. At dinner, will be on later.

num3ric:
@BanTheRewind I'm seeing these new artifacts in the latest glNext too fyi, I don't think it's related to this PR.

ryanbartley:
that is correct. I'm narrowing down the commit that breaks it in glNext now.

BanTheRewind:
Does your glNext include cam changes? Switching to MayaCam fixed it.

ryanbartley:
This is the commit that breaks it. https://github.com/cinder/Cinder/commit/afab295c7f744a630784822db736e7c6459fa0d5
the one before works.

BanTheRewind:
Ah! The cube change totally makes sense!

ryanbartley:
So what should i do? Can it be fixed in the math or does it need to be fixed at the commit?

BanTheRewind:
So I actually ran into this a long time ago (well before this sample) with Cinder and had to write a cube specific to point lights. I was pleased when glNext cubes just worked. It's been awhile, but I believe the winding order was CW so I had to make a CCW cube so culling worked properly. Not sure if that's the issue. Will be home in a bit and can investigate further.

BanTheRewind:
Also, disregard the camera comments. I made the mistake of anti-vaxxers and implied correlation with causation. ;)

BanTheRewind:
OK, I see what's happening. The cubes are being draw at half of the light volume (radius). I can fix this in the samples by drawing cubes at 2x Light::mVolume.




Super easy. Will update when I get home. Try it yourself (scale cubes by 2x volume) and see if it works.

flight404:
That seems like the culprit. My cube ended up 1/2 size after pulling this
commit. Scale( 2.0f ) did the trick.

On Fri, May 1, 2015 at 12:02 AM, Stephen Schieberl <notifications@github.com

richardeakin:
We should probably make a note on the forum that Cube's size is now actually size, so by default 1 = 1 unit (before 1 = 2 units, see [this issue](https://github.com/cinder/Cinder/issues/833)

richardeakin:
Oh, and by the way don't forget you can do: `geom::Cube().size( vec2( 2 ) )` instead of scaling it later somehow.

flight404:
I went with geom::Cube() >> geom::Scale( vec3( 2.0f ) ) because im
different.

BanTheRewind:
True. 




If a sphere of one and a cube of one are drawn next to each other, should they appear to be the same size?




Correct me if I'm wrong, but the recent update would make the sphere twice as big? Is that to be expected? Does one mean radius for a sphere, but the length of a side for a cube?

num3ric:
I declare this PR officially hijacked ;)

ryanbartley:
hahaha.

BanTheRewind:
Sorry. :P




Let's get this merged and then we can have a meandering conversation on my deferred sample updates PR.

ryanbartley:
haha. This has been a funny one. I feel like we've gone on a journey of Cinder.

BanTheRewind:
But I guess we should probably check if any samples look weird now with half size cubes.




For the record, I believe 1 should represent a sphere's diameter and a cube's side's length. The defaults from geom should all look the same size, IMO.

BanTheRewind:
Totally.

BanTheRewind:
Deferred samples updated with bug fixes, Xcode project and camera updates.

src/cinder/gl/GlslProg.cpp
```diff
@@ -329,6 +516,57 @@ GlslProg::GlslProg( const Format &format )
...
+				// check if the user defined attribute has cached what type
+				// of semantic this is
+				if( userAttrib.mSemantic != geom::Attrib::USER_DEFINED ) {
+					activeAttrib.mSemantic = userAttrib.mSemantic;
+				}
+				break;
+			}
+		}
+		if( !active ) {
+			CI_LOG_E( "Unknown attribute: \"" << userAttrib.mName << "\"" );
```
num3ric: Hey @ryanbartley, I hit this line because my attribute had been automatically optimized out since I wasn't using it. Would it be possible to make this message more explicit, or add a comment? Or are there more cases where this would be triggered?

richardeakin:
#1028 is open, which proposes to turn this and related error logs into warnings. I do think they help diagnose serious problems, but yea not always an error.

PR #835, opened 2015-04-21
init ivars in Sphere’s default constructor to zero

richardeakin:
Also make sure calculateBoundsSphere() doesn't crash when pased an empty array.


richardeakin:
After discussing offline, I've updated this PR to remove the zero initialization. To avoid the crash on empty array, we instead return `Sphere( vec3( 0 ), 0 )`.

PR #842, opened 2015-04-22
Upgrade Xcode project warnings with Xcode 6.3.1 suggestions

pizthewiz:
These changes should be backward compatible, though older Xcode builds may ignore some settings if they are unrecognized.


andrewfb:
Thanks for this - manually merged 93688e37cbf0c1cab00d05fc2f3caea7a2cdebf6

Issue #824, opened 2015-04-16
Add gl::ScopedDepthBuffer( enableRead, enableWrite )?


paulhoux:
This is more a reminder that I want to add this to our RAII objects.

richardeakin:
I'd like this too, and as convenience an overload that only takes one bool that enables or disables both read and write.

simongeilfus:
Haha well, that's a funny coincidence... I just pushed [that branch](https://github.com/simongeilfus/Cinder/tree/glNext_ScopedDepthTesting)!
Didn't had time to fully test it yet though

simongeilfus:
I just pushed a commit that changed the class to the name Paul was suggesting, I guess ScopedDepthTesting was a bit confusing because of GL_DEPTH_TEST.

What about adding also a gl::ScopedDepthFunc? All the stack/push/pop needed are already in there, and unfortunately as if we have a gl::ScopedDepthBuffer( bool ) it's impossible to have a gl::ScopedDepthBuffer( GLenum func ) without introducing ambiguity. So we'll need another class if we want to have this other convenience. I personally change a lot the depth function without touching at the other states.

simongeilfus:
the current interface is like this:
```c++
gl::ScopedDepthBuffer enableAll;
gl::ScopedDepthBuffer disableDepth( false );
gl::ScopedDepthBuffer disableDepthWrite( true, false );
gl::ScopedDepthBuffer enableReadWriteWithGreater( true, true, GL_GREATER );
```

And could look like this for the depth func:

```c++
gl::ScopedDepthFunc depthFunc( GL_LEQUAL );
```

simongeilfus:
Would that be better?

~~gl::ScopedDepthBuffer enableAll;~~
```c++
gl::ScopedDepthBuffer disableBothReadAndWrite( false );
gl::ScopedDepthBuffer enableBothReadAndWritel( true );
gl::ScopedDepthBuffer disableDepthWrite( true, false );
gl::ScopedDepthBuffer enableReadWriteWithGreater( true, true, GL_GREATER );
```

richardeakin:
I like your current proposed version, handling all three cases with constructor overloads. Yea not sure if we need a default enable = true, the following seems like it would read quite nicely to me:

```cpp
gl::ScopedDepthBuffer scopedDepth( true );
```

Not that I'm against a default bool = true, either.

simongeilfus:
[...]

PR #845, opened 2015-04-23
glNext gl::ScopedDepthBuffer

simongeilfus:
This PR adds a new RAII object for depth testing... see the discussion going [here](https://github.com/cinder/Cinder/issues/824).


richardeakin:
Thanks!

andrewfb:
After talking to Rich for a bit about this, can I propose we rename ScopedDepthBuffer to ScopedDepth? Besides a little advantage in terms of brevity, I think "Buffer" evokes Framebuffers and the like, whereas what we're really changing is state.

richardeakin:
I made the name change to `ScopedDepth` in 75bc9de7145bbc6b.

PR #848, opened 2015-04-26
Texture::setData now respects internal format

m1keall1son:
If i wanted to create a one channel 16u texture (often the case for depth textures) and access it in the shader as a usampler2D, cinder automatically defaulted to GL_RED in the Texture::update method, breaking the ability to sampler the texture at all if it was not initialized with type GL_RED

Now the internal format is checked and the correct dataFormat is used.


m1keall1son:
oh also theres the test case in the test/_opengl dir :)

richardeakin:
Along with that test, I see a [500k icon checked in](https://github.com/cinder/Cinder/pull/848/files#diff-9288b54be0475a2c8360eafa9ce4a7cd). Can you remove this? The rule of thumb is to reference the icons in samples/data, though for tests they don't need icons since they are only for, well, testing.

m1keall1son:
oops!  its removed.

andrewfb:
Rich just pointed out this PR includes test/_opengl/VolumeKinect - was that intentional? I will remove it if not.

m1keall1son:
Ah fuck that's an accident plz remove

m1keall1son:
But could be a cool sample at some point haha

richardeakin:
Yea I was thinking perhaps add it to the kinect block's repo if if it seems usefult for others.

m1keall1son:
its not far enough along enough for that unfortunately

Issue #723, opened 2015-02-24
Renaming .size() to .capacity() on ci::ConcurrentCircularBuffer<T>


num3ric:
https://github.com/cinder/Cinder/blob/master/include/cinder/ConcurrentCircularBuffer.h#L110

To follow the standards, and because at least twice I've made the mistake of trying to using .size() instead of say .isNotFull(), assuming it was dynamic. *(And yes I realize it's a stupid mistake to make given that doing a comparison would not be atomic.)*

Thoughts?

richardeakin:
Makes sense to me. Also, should probably be `getCapacity()`, to indicate that it isn't some typedef'ed standard container.

Does it work to change the size implementation to something like the following?:

```
size_t getSize() const
{
    std::lock_guard<std::mutex> lock( mMutex );
    return mNumUnread;
}
```

If we were to make these changes, now seems like a good time (break it good!).

andrewfb:
Addressed in #852

PR #852, opened 2015-04-29
Clear up concurrent circular buffer's size & capacity interface.

num3ric:


num3ric:
Refers to issue #723

andrewfb:
I'm good with this - anyone object to this (slightly) breaking change?

include/cinder/ConcurrentCircularBuffer.h
```diff
@@ -107,7 +107,13 @@ class ConcurrentCircularBuffer : private Noncopyable {
 	}
 	
 	//! Returns the number of items the buffer can hold
-	size_t size() const { return (size_t)mContainer.capacity(); }
+	size_t getCapacity() const { return (size_t)mContainer.capacity(); }
+	
+	//! Returns the number of items the buffer is currently holding
+	size_t getSize() {
```
richardeakin: I think this should be const, and as such I don't believe it needs the protection of the mutex (see: [you don't know const](http://herbsutter.com/2013/01/01/video-you-dont-know-const-and-mutable/)).

richardeakin: Hah, I realize in the linked issue I suggested implementing `getSize()` with a mutex, but then now that I'm thinking about it doesn't seem necessary.

richardeakin:
Sorry, I had a bit of a brain fart last night when discussing this. To me what you have now is correct, after making getSize() const and the mutex mutable.

PR #853, opened 2015-04-29
Disabled msaa by default

num3ric:
Follows from [this forum post](https://forum.libcinder.org/topic/cinder-multisampled-anti-aliasing-and-large-high-resolution-displays), and the email discussion. Are we still going for this? Should I go through more samples?


richardeakin:
I believe we're going forward with this change. We agreed on updating all samples to use MSAA unless they specifically don't need it (such as PostProcessingAA or the DeferredShading samples), also to modify the tinderbox templates to enable MSAA 16 by default.

Issue #856, opened 2015-04-29
gl::ScopedDepth not pushing first depth func state


simongeilfus:
Edit: No point in reading this, see pull request below.

Hey guys,

I've realised that my [commit](https://github.com/cinder/Cinder/pull/845) for this feature has some major issue. I've kind of reproduced the exact same pattern as the other states when implementing this and didn't really went deep into understanding the mechanism behind ```Context::push/popStackTrace```.

At the moment there's no way to tell what was the initial depth function when popping the last value from the stack. ```gl::ScopedDeph``` basically doesn't work when setting the depth function. I've double and triple check and my push/popDepthFunc seem to be doing exactly the same as the others, and I'm having trouble understanding how it can be working in other cases.

Here's a example, let say that you have 3 ScopedDepth like this :

```c++
{gl::ScopedDepth depth( true );}
{gl::ScopedDepth depth( true, false, GL_GREATER );}
{gl::ScopedDepth depth( true );}
```

The first and last one will not touch the depthFunc stack has it it's not modified by those calls. I'm expecting the depth function after those two calls to be set to GL_LESS.

The second one on the other hand will start by calling the function ```pushDepthFunc( GLenum func )```, pushing ```GL_GREATER``` as the first value of the stack. It will quickly be followed by a call to the destructor, calling ``popDepthFunc()`` as it is supposed to. This one will indeed pop the last ```GL_GREATER``` enum and directly after call ```getDepthFunc()``` that will indeed find an empty stack and glGet the current value. Obviously the current value will be the same, which is ```GL_GREATER```, it will be put twice on the stack and be set with the ```glDepthFunc``` instead of setting the default value GL_LESS that was before.

I really don't understand how this mechanism can work (here or everywhere else) without checking for an empty stack at the very first call to push*. Maybe I introduced an error while trying to replicate those functions but I honestly don't understand. 

Obviously it works perfectly if you call ```Context::getDepthFunc()``` before doing anything else. As this function do check for an empty stack before doing anything else.

I'm really sorry for having pushed this without testing more and would love if someone could shine some light to the general pushStackState/popStackState mechanism that is pretty much everywhere... Am I missing something obvious?

Thanks a lot.

simongeilfus:
Alright, .... sometimes writing an issue makes you realise the obvious!!!
Working on a pull request now.! ; )

PR #857, opened 2015-04-29
Added default/initial Depth test and func stack values

simongeilfus:
Fix my previous bad commit and solves my [shameful monologue here.](https://github.com/cinder/Cinder/issues/856)

Sorry about that! : )


andrewfb:
Glad to hear all of glNext is not fundamentally broken :).

Issue #864, opened 2015-05-01
AxisAlignedBoxes constructed by .include() don't update vertices


lithium-snepo:
When calling `::include ( const vec3 &point )` on an `AxisAlignedBox3<T>`, only the extents are recalculated, not the individual vertices. This manifests itself in a zero sized box after using `::transformed()`, and presumably erroneously failing intersection tests.

```cpp
AxisAlignedBox3f box;
box.include(vec3(-1));
box.include(vec3(1));
    
std::cout << box.getSize() << std::endl;
box = box.transformed( mat4() );
std::cout << box.getSize() << std::endl;
```

richardeakin:
This makes sense to me, though perhaps an implementation worth considering would be to mark the vertices as dirty and update them the next time that `box.getSize()` is called. This would mean if you add a list of vec3 points then you're only doing the `recalculateVertices()` once. This would be an optimization, so it could also be done at a later time than this bug fix as well..

richardeakin:
I believe this was addressed in #924.

PR #865, opened 2015-05-01
AxisAlignedBox vertex recalculation

lithium-snepo:
Addresses #864


pizthewiz:
Should this be targeting glNext instead of master?

lithium-snepo:
Yes, I'm a dummy. Is there an easy way to change it or should i just delete it an submit a new PR?

PR #866, opened 2015-05-01
AxisAlignedBox vertex recalculation

lithium-snepo:
Let's try this again ;)
Addresses #864

### DIFF
--- a/include/cinder/AxisAlignedBox.h
+++ b/include/cinder/AxisAlignedBox.h
@@ -30,7 +30,7 @@ namespace cinder {
 
 class AxisAlignedBox3f {
  public:
-	AxisAlignedBox3f() {}
+	AxisAlignedBox3f() { mNeedsRecalculate = false; }
 	AxisAlignedBox3f( const vec3 &aMin, const vec3 &aMax );
 
 	bool	intersects( const Ray &ray );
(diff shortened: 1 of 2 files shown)


richardeakin:
This makes sense to me, though perhaps an implementation worth considering would be to mark the vertices as dirty and update them the next time that box.getSize() is called. This would mean if you add a list of vec3 points then you're only doing the recalculateVertices() once. This would be an optimization, so it could also be done at a later time than this bug fix as well..

lithium-snepo:
This did occur to me but i thought given what I assumed to be a typical use case that it was probably unnecessary. If you think it's something that will end up getting added later, I might as well just add the optimization to this PR.

lithium-snepo:
Ugh, i'm *terrible* at git. The fix has been made, but i need to work out this rebasing business.

pizthewiz:
I've found the edX team's [How to Rebase a Pull Request](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) guide to be super helpful.

lithium-snepo:
Thanks for that. I thought i had it working but I ended up just piling more and more shit into the branch, so I just bailed on it. I'll play around on some local repositories before I go committing willy-nilly to public ones in the future.

PR #868, opened 2015-05-03
Remove hardcoded QTKit and CoreVideo framework paths

pizthewiz:
While Xcode seemed to sort out how to properly handle this at compile-time, the frameworks were highlighted red in the file list indicating it couldn't find them which was a bit annoying. I suspect this is just a setting carried over from the original Xcode project file.


richardeakin:
Nice one, thanks for the cleanup.

PR #870, opened 2015-05-04
fixed and optimized AABB recalculation

lithium-snepo:
Addresses #864 and includes @richardeakin's optimization notes. Sorry about all the PR noise, i'm a complete git n00b.

### DIFF
--- a/include/cinder/AxisAlignedBox.h
+++ b/include/cinder/AxisAlignedBox.h
@@ -30,7 +30,7 @@ namespace cinder {
 
 class AxisAlignedBox3f {
  public:
-	AxisAlignedBox3f() {}
+	AxisAlignedBox3f() { mNeedsRecalculate = false; }
 	AxisAlignedBox3f( const vec3 &aMin, const vec3 &aMax );
 
 	bool	intersects( const Ray &ray );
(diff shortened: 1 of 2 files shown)


paulhoux:
Hi Lithium,

thanks for the pull request. I have a few remarks:

* the ```transformed()``` method should stay const, as it does not alter the existing AxisAlignedBox3f, but just returns a new one. To do this while still being able to cache/update ```mVerts```, both the boolean and ```mVerts``` should be made mutable.
* while lazy updating the vertices is faster, I am not sure ```include``` is called often enough to add lazy updating to the class. Can you name a use case where you really need the performance boost that lazy updating gives you?
* not really related to this pull request, but I noticed that the code can become a lot cleaner by simply using ```glm::min( vec3 a, vec3 b )``` throughout. Here's what it would look like:
https://github.com/paulhoux/Cinder/commit/97641b1d27a5c77699816b92ca1019367011891f

lithium-snepo:
1. Sounds fine to me
2. I originally didn't have the optimization for `include()`, but added it at @richardeakin's suggestion. I don't have strong feelings about it one way or another
3. I agree with this too, though my general move when committing is to leave whatever isn't directly affected as it was, so i left the min/max stuff as it was. 

Thanks for tidying up the PR. I'll get better at it as time wears on.

A.

richardeakin:
Concerning the lazy recalculate, I think it is more efficient for things like calculating the bounding box of a particle system (where you might call `include()`, say, 10,000 times but only want the size once). That said it was a suggestion in the dark, I didn't profile at all and isn't a great idea to add without profiling the code with and without. In any event it falls into the category of optimization, so good to keep separate from the core of this PR, which is a bug fix.

lithium-snepo:
Fair enough. Should i close this pull request and you blokes can use paul's much cleaner version?

paulhoux:
Hi,

since you did most of the work, @lithium-snepo , I want your name on the pull request :) What you might do is merge my version into your branch and push it to your repo once you're happy with it. That way it will become part of this pull request.

-Paul

paulhoux:
Hi Lithium,

since the changes I made were done on top of your pull request, your contribution will also be be recorded to the repo if I merge my code with glNext. So if it is the same to you, I could close this pull request and merge my code. Let me know if that works for you.

-Paul

lithium-snepo:
That's fine mate. Sorry i fell behind on this. The thing that I was working on to find the bug in the first place popped up on me. Thanks.

paulhoux:
See #924. Thanks, Lithium.

PR #883, opened 2015-05-06
Added solid and stroked rounded rect, based of 0.8.6 implementation

ryanbartley:


ryanbartley:
Hold on this. I'm updating the implementation of Solid to allow for texCoords and Normals.

src/cinder/gl/draw.cpp
```diff
@@ -880,6 +1052,7 @@ void drawSolidCircle( const vec2 &center, float radius, int numSegments )
 		if( normals )
 			normals[s+1] = vec3( 0, 0, 1 );
 		t += tDelta;
+		cout << texCoords[s+1] << endl;
```
andrewfb: Looks like a stray cout snuck in here...

ryanbartley: whoops! I'll take care of that. Also, I'll update the Stroked Rounded Rect as well.

src/cinder/gl/draw.cpp
```diff
@@ -718,6 +833,63 @@ void drawStrokedRect( const Rectf &rect, float lineWidth )
...
+	int posLoc = curGlslProg->getAttribSemanticLocation( geom::Attrib::POSITION );
+	if( posLoc >= 0 ) {
+		enableVertexAttribArray( posLoc );
+		vertexAttribPointer( posLoc, 2, GL_FLOAT, GL_FALSE, 0, (void*)0 );
+	}
+	ctx->getDefaultVao()->replacementBindEnd();
+	ctx->setDefaultShaderVars();
+	ctx->drawArrays( GL_LINE_LOOP, 0, tri );
+	ctx->popVao();
+	delete [] verts;
```
richardeakin: i know this is the from the version in 0.8.6, but could you use unique_ptr, or better yet std::vector, to manage the memory without an explicit delete? You can [reserve()](http://en.cppreference.com/w/cpp/container/vector/reserve) the memory it needs up front to make it the same in efficiency.

Thanks for adding these though, they're one of the last missing ones remaining.

ryanbartley: Yeah. That's what I meant about updating this method like the drawsolid version.

Issue #899, opened 2015-05-11
ip::EdgeDetect not working with Surface32f and Channel32f


simongeilfus:
It seems that ip::edgeDetect is only supporting Surface8u and Channel8u.
From what I can tell, the [reinterpret_cast](https://github.com/cinder/Cinder/blob/master/src/cinder/ip/EdgeDetect.cpp#L51) and [static_cast](https://github.com/cinder/Cinder/blob/master/src/cinder/ip/EdgeDetect.cpp#L59) are not necessary and make the function only compatible with *8u objects...

You can try with this app:

```c++
#include "cinder/app/App.h"
#include "cinder/app/RendererGl.h"
#include "cinder/gl/gl.h"
#include "cinder/ip/EdgeDetect.h"

using namespace ci;
using namespace ci::app;
using namespace std;

class SobelApp : public App {
  public:
	void setup() override;
	void draw() override;
	
	gl::Texture2dRef mTexture;
};

void SobelApp::setup()
{
	using SourceType = Channel;
	//using SourceType = Surface;
	//using SourceType = Channel32f;
	//using SourceType = Surface32f;
	SourceType source = loadImage( loadAsset( "lena.png" ) );
	SourceType destination = source.clone();
	ip::edgeDetectSobel( source, &destination );
	mTexture = gl::Texture2d::create( destination );
}

void SobelApp::draw()
{
	gl::clear( Color( 0, 0, 0 ) );
	gl::draw( mTexture );
}

CINDER_APP( SobelApp, RendererGl, []( App::Settings* settings ){ settings->setWindowSize( 512, 512 ); } )
```

Here's a potential [fix](https://github.com/cinder/Cinder/pull/900).

andrewfb:
Should be fixed by above...

PR #900, opened 2015-05-11
Potential Fix for ip::EdgeDetect and Surface32f/Channel32f

simongeilfus:
Potential fix for https://github.com/cinder/Cinder/issues/899


andrewfb:
Thanks for this - made a few tweaks to getting it working on 16u but should be good now.

PR #903, opened 2015-05-12
Fixing outside of screen crash

MattijsKneppers:
This change should fix the crash when the output window is positioned fully outside of visible screen space and then moved again.

findFromCgDirectDisplayId() returns null when the window is outside of the monitor space. This is fine until the next time the event is triggered, when trying to compare the new display id to the one of the current (non-existing) display.

I chose to emitDisplayChange() only when the window re-appears in visible screen space.

### DIFF
--- a/src/cinder/app/AppImplCocoaBasic.mm
+++ b/src/cinder/app/AppImplCocoaBasic.mm
@@ -555,9 +555,15 @@
 
 	[mAppImpl setActiveWindow:self];
 
-	if( displayID != mDisplay->getCgDirectDisplayId() ) {
+	if( mDisplay ) {
+		if( displayID != mDisplay->getCgDirectDisplayId() ) {
+			mDisplay = cinder::Display::findFromCgDirectDisplayId( displayID );
+			mWindowRef->emitDisplayChange();
+		}
+	}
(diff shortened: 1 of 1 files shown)


richardeakin:
Hey there, the fix is very much appreciated. However we are looking to merge the glNext branch into master soon, wherein the files within the app namespace have been heavily rearranged. Would you mind creating a PR against that branch so that it will cleanly merge into the to-be 0.9.0 codebase?

Also, do you know of steps where we can repro the bug?  For instance I'm not entirely sure how to get the window off of the visible screen, and then how I should move it in order to cause crash.

MattijsKneppers:
Aah, it looks like this has already been taken care of in the glNext branch. Good to know that this is the most recent development branch.

Is there an easy way to transition an app from the master to glNext other than generating a new app with the new Tinderbox and adding old source step by step?

richardeakin:
Ok cool. I've been updating any existing projects by hand, but then I'm familiar with the changes. We're planning to do a short 'transitioning guide' before releasing, but nothing automated is planned. Most of the changes are relatively straightforward, but let us know if you have any questions about the process.

PR #907, opened 2015-05-13
Added support for missing int and unsigned int samplers

gaborpapp:
Without these type mismatch warnings and unknown uniform errors are displayed when a uniform of `GL_UNSIGNED_INT_SAMPLER_2D`, `GL_INT_SAMPLER_2D_RECT`, or `GL_UNSIGNED_INT_SAMPLER_2D_RECT` is set.


andrewfb:
Thanks for flagging these; a few are actually defined in ES 3 but I think 65299b417c4d550ffb38e2621a8bf62536507450 should address these and some additional constants.

gaborpapp:
Thanks!

PR #910, opened 2015-05-14
Add newline after user-provided shader version when using defines

### LINKED ISSUE #909: GLSL preprocessor appends define onto #version value causing exception?
Say I have a passthrough shaders:
```C++
const std::string EffectorNode::sVertexShaderPassThrough = R"(
    #version 150
    uniform mat4 ciModelViewProjection;

    in vec4 ciPosition;
    in vec2	ciTexCoord0;

    out vec2 vTexCoord0;

    void main() {
        vTexCoord0 = ciTexCoord0;
        gl_Position = ciModelViewProjection * ciPosition;
    }
)";
const std::string EffectorNode::sFragmentShaderPassThrough = R"(
    #version 150
    uniform sampler2D image;

    in vec2 vTexCoord0;

    out vec4 oFragColor;

    void main() {
        oFragColor = texture(image, vTexCoord0);
    }
)";
```

if any defines are added:
```C++
auto format = gl::GlslProg::Format().vertex(sVertexShaderPassThrough).fragment(FragmentShaderGaussianBlur);
format.define("ACK", toString(1));
mShader = gl::GlslProg::create(format);
```
the resulting `preprocessedSource` from [`GlslProg::LoadShader`](https://github.com/cinder/Cinder/blob/glNext/src/cinder/gl/GlslProg.cpp#L628) has the define tacked right onto the version number:
```gdb
(lldb) p preprocessedSource
(std::__1::string) $1 = "    #version 150#define ACK 1\n\n\n    uniform mat4 ciModelViewProjection;\n\n    in vec4 ciPosition;\n    in vec2\tciTexCoord0;\n\n    out vec2 vTexCoord0;\n\n    void main() {\n        vTexCoord0 = ciTexCoord0;\n        gl_Position = ciModelViewProjection * ciPosition;\n    }\n"
```
which I think is causing  a compile error:
```
|error  | void cinder::app::AppBase::executeLaunch()[140] Uncaught exception, type: cinder::gl::GlslProgExc, what : VERTEX: ERROR: 0:1: '' : syntax error: #version

libc++abi.dylib: terminating with uncaught exception of type cinder::gl::GlslProgExc: VERTEX: ERROR: 0:1: '' : syntax error: #version
```

This seems to be a regression in glNext over the last week or so, #895 seems related but I think the code path that was changed there is when the preprocessor inserts a version, so perhaps not.

pizthewiz:
fixes #909 

Adding the newline all the time would change the line numbers regardless of if the preprocessor was actually changing the shader source and without a newline, defines will append to the version line like:
```glsl
#version 150#define VARIABLE
…
```

### DIFF
--- a/src/cinder/gl/ShaderPreprocessor.cpp
+++ b/src/cinder/gl/ShaderPreprocessor.cpp
@@ -104,6 +104,9 @@ std::string ShaderPreprocessor::parseDirectives( const std::string &source )
 		version = "#version " + to_string( mVersion ) + "\n";
 #endif
 	}
+	else if( ! mDefineDirectives.empty() ) {
+		version += "\n";
+	}
 
 	// copy the preprocessor directives to a string starting with the version
 	std::string directivesString = version;


richardeakin:
Thanks for the fix!

PR #913, opened 2015-05-15
Added a drawStrokedCircle that supports a lineWidth setting.

paulhoux:
In response to https://forum.libcinder.org/#Topic/23286000002349013 , I have created a version of ```drawStrokedCircle``` that supports an arbitrary line width. It creates a triangle strip instead of a line loop. 

I should not have to mention that for line width values smaller than 1.0, you need proper (msaa) sub-sampling to make it look good.

### DIFF
--- a/include/cinder/GeomIo.h
+++ b/include/cinder/GeomIo.h
@@ -408,6 +408,32 @@ class Circle : public Source {
 	size_t		mNumVertices;
 };
 
+class Ring : public Source {
+public:
+	Ring();
+
+	Ring&		center( const vec2 &center ) { mCenter = center; return *this; }
+	Ring&		radius( float radius );
+	Ring&		width( float width );
+	Ring&		subdivisions( int subdivs );
+
+	size_t		getNumVertices() const override;
+	size_t		getNumIndices() const override { return 0; }
+	Primitive	getPrimitive() const override { return Primitive::TRIANGLE_STRIP; }
+	uint8_t		getAttribDims( Attrib attr ) const override;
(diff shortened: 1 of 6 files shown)


ryanbartley:
Is there a way this could go into the geom::WireCircle?

richardeakin:
I think the design choice is that Wire* objects create line geometry, or in other words geometry suitable for GL_LINES. We were thinking that perhaps a `geom::Disk` or `geom::Ring` would be more suitable for that task. My vote for namig is Ring, but either would work.

I think if we were to go that route, I still like the idea of using gl::drawStrokedCircle() as the convenient way to use it, since it will be easy to find once you are used to the naming pattern of those methods (whereas gl::drawRing() would be harder to understand).

andrewfb:
I agree with the geom::Ring idea as a nice additional piece of functionality, and then keeping drawStrokedCircle() variant that takes a `lineWidth` as the convenience entry point. Ryan just did something similar w/ geom::RoundedRect. It also helps us maintain the "only use convenience methods for convenience - upgrade to Batch / geom:: variants when perf matters" philosophy.

BanTheRewind:
Nice addition. In my old MeshHelper block I had this as "ring", so that gets my vote, as well.  :)

paulhoux:
OK, ```geom::Ring``` added and its convenience method is using it, too. Also fixed an issue with ```WireCircle``` (didn't draw at all, due to missing pointer increment) and ```geom::Rect``` (incorrect polygon winding).

PR #920, opened 2015-05-17
To avoid a race condition, I created a ScopedClipboard class that all…

paulhoux:
…ows exclusive access to the clipboard on Windows. Let me know if you have any objections or remarks.

### DIFF
--- a/include/cinder/Clipboard.h
+++ b/include/cinder/Clipboard.h
@@ -6,9 +6,9 @@
  Redistribution and use in source and binary forms, with or without modification, are permitted provided that
  the following conditions are met:
 
-    * Redistributions of source code must retain the above copyright notice, this list of conditions and
+	* Redistributions of source code must retain the above copyright notice, this list of conditions and
 	the following disclaimer.
-    * Redistributions in binary form must reproduce the above copyright notice, this list of conditions and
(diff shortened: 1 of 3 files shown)


richardeakin:
It's difficult to tell what race condition this change is solving because the PR is modifying quite a lot in just one commit. Can you explain the race condition?

Also, I see the RAII scoped structs we're using as a way to conveniently and safely push / pop, or lock / unlock something, but there should always be a way to do this sort of operation manually as well. It looks to me like with the changes proposed in this PR, Clipboard would require using ScopedClipboard?

paulhoux:
The race condition can occur between a call to "hasImage" and "getImage"... another application might replace the contents of the clipboard between these calls. This happened to me while running the demo, resulting in a (rare) crash.

The ScopedClipboard will make sure no other application can modify the clipboard (on Windows). But you can still use the default Clipboard class instead.

I have discussed this with Andrew and we're probably not going forward with this change. It might be better to simply return an empty pointer from getImage and let the demo handle this situation.

richardeakin:
Yea in that case, I'd vote for using `Clipboard::getImage()`'s returned value (possibly null) to determine if there is a an image vailable or not.

PR #921, opened 2015-05-18
Added windows EventLog support for system logging

lucasvickers:
Built out windows support for EventLog logging.  Now when SystemLog is enabled it will use EventLog on windows and syslog on OS X.

I don't love how I need to convert log messages to wchar strings every log, but I see no way around it.


richardeakin:
Looks great, Lucas and I talked in person today so we know how to proceed here but I'll note a couple minor style things in line as well. I think the big points were:

* lets try not to write to console() from a Logger::write() method (on error) since that could lead to log spew.
* We discussed comping LoggerEventLog and LoggerSysLog into one LoggerSystem with different implementations, I think we decided it was a good move to reduce one more public facing Logger class.

src/cinder/Log.cpp
```diff
@@ -268,9 +281,12 @@ void LogManager::enableSystemLogging()
 
 #if defined( CINDER_COCOA )
 	addLogger( new LoggerSysLog );
+#elif defined( CINDER_MSW )
+	addLogger(new LoggerEventLog );
```
richardeakin: space between opening '(' and 'new'

src/cinder/Log.cpp
```diff
@@ -502,6 +521,64 @@ void LoggerSysLog::write( const Metadata &meta, const string &text )
...
+
+LoggerEventLog::LoggerEventLog()
+{
+	char filename[MAX_PATH];
+	wchar_t wFilename[MAX_PATH];
+	string stem;
+
+	DWORD size = GetModuleFileNameA( NULL, filename, MAX_PATH );
+	if( size ) { 
+		boost::filesystem::path exePath(filename);
```
richardeakin: Here we probably want to use `fs::path`, since that'll eventually turn into the C++17 std::filesystem::path. That is already used for Visual Studio 2015 (or vc140), see [here](https://github.com/cinder/Cinder/blob/glNext/include/cinder/Filesystem.h#L38-L44) for details.

src/cinder/Log.cpp
```diff
@@ -502,6 +521,64 @@ void LoggerSysLog::write( const Metadata &meta, const string &text )
...
+// MARK: - LoggerEventLog
+// ----------------------------------------------------------------------------------------------------
+
+LoggerEventLog::LoggerEventLog()
+{
+	char filename[MAX_PATH];
+	wchar_t wFilename[MAX_PATH];
+	string stem;
+
+	DWORD size = GetModuleFileNameA( NULL, filename, MAX_PATH );
```
richardeakin: We haven't documented this much, but when calling platform specific C API's, especially for windows or mac, we prepend a `::` to ensure the name comes from the global namespace, also documenting that fact.

src/cinder/Log.cpp
```diff
@@ -502,6 +521,64 @@ void LoggerSysLog::write( const Metadata &meta, const string &text )
...
+
+	if( ! mHLog ) {
+		app::Platform::get()->console() << "RegisterEventSourceW() failed with " << GetLastError() << endl;
+	}
+}
+
+LoggerEventLog::~LoggerEventLog()
+{
+	if( mHLog ) {
+		CloseEventLog(mHLog);
```
richardeakin: spacing (and same `::` prefix treatment), should look like `::CloseEventLog( mHLog );`

lucasvickers:
I'll implement the wrapping LoggerSystem shortly.  We can leave this PR open until then.

lucasvickers:
Okay Rich please take a look.  I'd like your opinion on how I hid the logging implementation classes inside the LoggerSystem's private declarations.  Since setting the system logging level is different on each platform I couldn't think of a good reason why I would want to expose LoggerSysLog and LoggerEventLog, rather than only exposing LoggerSystem.

include/cinder/Log.h
```diff
@@ -35,6 +35,11 @@
 #include <memory>
 #include <mutex>
 
+#if defined( CINDER_MSW )
+#include <Windows.h>
```
richardeakin: Can these windows-specific headers be pushed to the .cpp? [CinderMswFwd.h](https://github.com/cinder/Cinder/blob/glNext/include/cinder/msw/CinderWindowsFwd.h) might help, but if not hopefully we can forward declare whatever we need so that Windows.h doesn't get pulled into user app files.

richardeakin:
In light of the necessity for platform specific headers, I'd argue that they should be pimpls, forward declared in the header and declared / implemented in the cpp:

```cpp
class LoggerSystem :  public Logger {
  public:
  private:
#if defined( CINDER_COCOA )
    struct ImplSysLog;
    std::unique_ptr<ImplSysLog> mImpl;
#elif defined( CINDER_MSW )
    struct ImplEventLog;
    std::unique_ptr<ImplEventLog> mImpl;
#endif
};
```
This means no other information about the impl needs to be in the header.  I also like unique_ptr here because it means that we avoid needing to a naked call to delete (though keep in mind that with unique_ptrs and forward declared types, class will force you to still add a destructor for `LoggerSystem` to the cpp, even if it is with `= default`).

include/cinder/Log.h
```diff
@@ -135,18 +140,46 @@ class LoggerBreakpoint : public Logger {
...
-
-	void write( const Metadata &meta, const std::string &text ) override;
-};
-
+	LoggerSystem();
+	virtual ~LoggerSystem();
+	
+	void write( const Metadata &meta, const std::string &text ) override { mImpl->write( meta, text ); }
+	// SysLog and EventLog logging levels inherently work differently (see specific comments).  Currently
+	// this doesn't matter since logging implimentation only allows once instance of a logger type.
```
richardeakin: How do they work differently? Not sure what the user should take away from this message, or if there are functionality differences in practice.

lucasvickers: The github diff may be confusing since it doesn't show that my logging level comments pertain to the function below
```
// SysLog and EventLog logging levels inherently work differently (see specific comments).  Currently
// this doesn't matter since logging implimentation only allows once instance of a logger type.
void setLoggingLevel( Level minLevel ) { mImpl->setLoggingLevel( minLevel ); }
```

To elaborate on how they are different:

In **SysLog** changing the logging level is done through a call to ```setlogmask```.  (I need to make it clear this is a system level call ie ```::setlogmask```).  This causes the OS X system logger to mask the logging call.  This would be applied to all SysLog calls from the application.

In **EventLog** we keep an internal log level and use an if statement to decide if we should log.

If, in future versions of the logger we allowed multiple ```LoggerEventLogs``` or multiple ```LoggerSysLogs```, then changing the logging level on one single SysLog would affect all SysLogs.  This is not the case for EventLog.  Also, if for some reason the user is calling syslog elsewhere in the app (outside of our CI_LOG), then the mask would also be applied to their syslog calls.

I could not use the setlogmask call, and do the same if statement check when logging either type.  It may actually be more efficient than passing our log message to SysLog and letting SysLog filter.  In that case both implementations would be functioning the same.

After writing all of this I'm leaning towards doing the loglevel check in ```LoggerSystem``` rather than in either ```LoggerSysLog``` or ```LoggerEventLog``` implementation.

richardeakin: Ah, yes I see, they are quite different. Essentially what you've done for EventLog is add runtime level filtering support. In the future I'd like to add this at a more generic level, perhaps the LoggerMulti, so we get the functionality for even things like LoggerFile, but I'd like to do it so it is minimal overhead when turned off - that is the tricky part.

It is a tough decision, what to do around ::setlogmask(). Part of me thinks it should be a static LoggerSystem::setLogMask(), that is only available on platforms where syslog exists. If that were done, it would make sense to either move the EventLog's runtime level filtering to LoggerSystem, or remove it altogether, since it will likely come back in a more generic form. 

Any thoughts around the pimpl move? I can get this merged and do the change locally if you prefer, or let me know if you have other opinions. I'd like to get this PR closed soon since it is against glNext, which is going to be removed soon since it was merged into master branch.

lucasvickers:
Not yet ready, I'm having issues with the windows forward declaration of ImplEventLog.  Will look into it further shortly.

lucasvickers:
Okay ready for review/merge.

richardeakin:
Cool, looks great to me.

richardeakin:
Oops, this merged into glNext, will manually get it into master branch now..

include/cinder/Log.h
lucasvickers: Sorry completely missed this comment somehow.




I think the pimpl makes a lot of sense and I implemented it. 




I gave a lot of thought to the LoggerSystem filtering. I think we need to explore optimized methods for setting logging filters down the road, but for now I like consistency between platforms and I doubt setlogmask can be any more efficient than a bool check. I also didn't love how it has a global application effect if you call it but use syslog elsewhere - but I'm not sure that's a real use case anyway. 




I think we need secondary control over system log levels because I see them used for reporting warnings or errors not general debug logging. There are tons of tools that scan event logs or syslogs and report on errors and they are what I think sys logging is good for. Other logs should just append to the (rotating) log file.

PR #924, opened 2015-05-19
Rewrite of the AxisAlignedBox3f class.

paulhoux:
Complete rewrite of the AxisAlignedBox3f class. 
* renamed to ```AxisAlignedBox```
* its internal representation has been reduced to 2 ```vec3```'s.
* most intersection tests now only use a few lines of code, thanks to the new internal structure and the use of ```glm::``` functions where possible.
* I have tried to keep backward compatibility, tested all samples that use an AABB and wrote a few unit tests.
* Added new functions, like ```contains(vec3)``` and ```project(vec3, float *min, float *max)```.
* implementation is header-only, to allow the compiler to inline the code where possible.
* Added ```transform()``` and ```transformed()``` methods to ```Area``` and ```Ray``` as well. The ```Rect``` class already had a ```transformCopy()``` method. After discussing this with Andrew, we have chosen to rename these to ```transformed()``` and change them in ```Path2d``` and ```Shape2d``` as well.


paulhoux:
Ok, I think it is ready to go. Let me know if you have any questions or concerns.

ryanbartley:
Should this not be AxisAlignedBox, instead of AxisAlignedBox3f? Is it not implied that an aabb is within 3 dimensions and float? Not to nitpick, but since everything's been moving to glm and losing the 'f', I feel like this should as well. In fact, should it not just be call aabb?

paulhoux:
I agree on removing the '3f' suffix. And while AABB is a lot shorter, I still prefer AxisAlignedBox, since it is more descriptive. We can later add the OrientedBox, etc.

richardeakin:
I like the AxisAlignedBox naming, and the prospects of a related OrientedBox, rather than acronyms.

The implementations look fantastic.

include/cinder/AxisAlignedBox.h
```diff
@@ -28,37 +28,178 @@
...
+	//! Returns \c true if the axis-aligned boxes intersect.
+	bool intersects( const AxisAlignedBox &box ) const
+	{
+		return glm::all( glm::lessThanEqual( glm::abs( box.mCenter - mCenter ), box.mExtents + mExtents ) );
+	}
+
+	//! Returns \c true if the ray intersects the axis-aligned box.
+	bool intersects( const Ray &ray ) const
+	{
+		vec3 min = ( getMin() - ray.getOrigin() ) / ray.getDirection();
```
richardeakin: Should we protect against divide by zeros here? Ex. if the direction is vec3( 1, 0, 0 ) then min.y and .z will be NaNs.

ryanbartley:
Yeah amazing job Paul!!!

include/cinder/AxisAlignedBox.h
paulhoux: No, it's fine, but thanks for pointing it out. The #INF's will be ignored by the min() and max() functions.

PR #927, opened 2015-05-20
Quit fix

BanTheRewind:
Somewhere along the line of various merges, the bug where the app would crash because it was quitting from a params button was re-introduced. I've re-implemented the work around.


andrewfb:
Hey - are you seeing this crashing again? I changed the samples in light of #925, which work correctly on my machine on the latest glNext.

BanTheRewind:
Ah, gotcha. Yes, it crashes on exit with this. Same situation. VS2013 debug mode.

andrewfb:
Apologies - I used this app as my test case on two different machines so I just want to eliminate other possibilities. You definitely have the very latest glNext commits (<24 hours old)? If you step into the call to quit(), the code looks like this ecf583fe51a9b8ec3756616394a9c5c053a045c9 at AppImplMswBasic::quit()?

BanTheRewind:
Shoot, apologies. My branch's libs weren't updated. Your fix works.

BanTheRewind:
I would like to PR these tiny light scattering updates. I'll put the quit button back in and PR separately.

andrewfb:
No prob - way rather that. In light of this fix though you should be able to do what you had wanted to do originally. A call to quit() now takes effect once we return to the event loop. The bug you ran into was due to us closing the windows out from under you from inside quit().

PR #928, opened 2015-05-20
Light scattering tweaks

BanTheRewind:
One appearance tweak and fixed a uniform value.


richardeakin:
Sorry for the delay on this, just tested after reading the latest on this [forum post](https://forum.libcinder.org/topic/glnext-deferred-shading-samples#23286000002367099). I'm also seeing that the floor is missing when looking down on it, even after the changes in this PR. See this [video clip](https://dl.dropboxusercontent.com/u/3905723/cinder/screenrecordings/deferred_shading_sample_ground_gone.mov).

Also on the DeferredShaderAdvanced sample, it appears the framerate is pretty low again (both before and after the changes in this PR), down to ~11fps on a 2013 mpb, 30fps (in release) on my work PC with a Radeon R9 290. When trying to turn off some of the layers, I'm seeing strange artifacts where the ambient occlusion is.

screenshot 2015-06-07 14 18 46

While I like the feel of the new cylinder room, I sort of think the other additions like rays, blur, etc., are less important as making sure the app stays fast.

BanTheRewind:
That's what unfiltered SAO looks like. If you crank the sampling numbers
way up, it can look good without filtering, but it generally makes sense to
keep the "AO blur" turned on.

I just ran it on my 2010 MBP and am getting 30. What video card do you have
on the 2011?

I'll fix the floor in the morning. Probably a geom change?

richardeakin:
NVIDIA GeForce GT 750M. It was running really good when I last tested, but this was a couple iterations back for you.

BanTheRewind:
Weird. Will hunt it down. The last version was also the best performing for me. What FPS do you get with high quality off? This is closer to the previous "medium" mode. Maybe leave HQ off by default?

PR #937, opened 2015-05-26
iOS sample fixes for latest glNext, post app_refactor

richardeakin:
For iosKeyboard and iosNativeControls. The latter solution is not so elegant, the NativeViewController point had to be made global but I don't yet see any problem with that in a real world scenario.

There's something I'd like to verify with other obj-c gurus (pinging @pizthewiz), in that I had to declare the Window's UIViewController as `__unsafe_unretained` because it was causing the sample's NativeViewController object to be deallocated when the Window::Format went out of scope (the sample is using ARC, while cinder is still retain / release).  I believe this is kosher, that you can use `__unsafe_unretained` from within a non-ARC project, but it would be nice if someone else could verify as it is a bit hard to find docs related to this.

### DIFF
--- a/include/cinder/app/Window.h
+++ b/include/cinder/app/Window.h
@@ -254,7 +254,7 @@ class Window : public std::enable_shared_from_this<Window> {
 		bool					mTitleSpecified;
 
 #if defined( CINDER_COCOA_TOUCH )
-		UIViewController *mRootViewController;
+		__unsafe_unretained UIViewController *mRootViewController;
 #endif
 	};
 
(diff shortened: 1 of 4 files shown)


meshula:
Resolving issues like that is what it's for.

richardeakin:
Great, I'm off the hook then. :)

pizthewiz:
Sorry, just back from vacation and quickly agreeing that `__unsafe_unretained` is what you'd want here as it can be used in a non-ARC project. `__weak` additionally zeros the reference when deallocated but I believe requires ARC.

PR #942, opened 2015-05-28
Added static void Voice::clearBufferCache()

rockolo:
This function adds the ability to clear the audio buffer cache that the MixerImpl retains.

I am aware of audio::Voice::Options().maxFramesForBufferPlayback to avoid buffering all together, but for my current application I load scenes with dozens of audio files that get played multiple times, so I want them to be cached, but I also want to be able to clear the cache once I unload that scene and load a new scene. This commit adds the ability to do this.


richardeakin:
Looks good to me, thanks!

Issue #948, opened 2015-05-29
NormalMapping sample broken


gaborpapp:
The output is black only showing the emissive parts of the model. It was working a while ago, some changes probably broke it recently. OS X 10.9, NVidia 750M/Intel HD4000.

![screen shot 2015-05-29 at 12 00 12](https://cloud.githubusercontent.com/assets/203532/7880609/75311c5a-05fb-11e5-964a-b25b7e3ae897.png)

paulhoux:
Thanks for reporting, will look at it this weekend. If you find a solution yourself, let me know.

paulhoux:
Caused by the changes to glslProg. The sample sets the following uniform: 
```mShaderNormalMapping->uniform( "uLights[0].diffuse", mLightLantern.diffuse );```

, but apparently this is no longer supported. Will try to find a fix.

paulhoux:
@andrewfb  and @notlion  : the support for array brackets in uniform names does not allow accessing the ```diffuse` member of my struct. I've pushed a solution that fixes the sample, but I'm not sure what you guys were trying to fix. Could you go over it? You can find the pull request here: #951

gaborpapp:
Thanks for hunting this down. I checked the console for uniform errors, but nothing was printed, so I assumed they were set properly.

andrewfb:
This should be fixed and merged.

PR #951, opened 2015-05-31
Addressing GlslProg uniform issue with array of struct member access

paulhoux:
…where ```uUniformName[0].member``` is not supported because it only compares the name before the bracket.

See #948


paulhoux:
Will add a use case to the ```GlslProgAttribTestApp```.

PR #952, opened 2015-05-31
Build iOS simulator target with simulator SDK to avoid bad archs

### LINKED ISSUE #825: cinder_iphone_simulator scheme builds arm7/arm64 NOT i386/x86_64 via fullbuild.sh
glNext seems to build the simulator target binaries for `armv7` and `arm64`. I haven't done any iOS work with Cinder but it should be `i386` and `x86_64` no?
```sh
$  lipo -info lib/libcinder-iphone* 
Architectures in the fat file: lib/libcinder-iphone-sim.a are: armv7 arm64 
Architectures in the fat file: lib/libcinder-iphone-sim_d.a are: armv7 arm64 
Architectures in the fat file: lib/libcinder-iphone.a are: armv7 arm64 
Architectures in the fat file: lib/libcinder-iphone_d.a are: armv7 arm64 
```

I'm not familiar with the use case or specifics, but is a separate simulator target needed, would it not be enough for the iOS target to build fat?

pizthewiz:
Fixes #825

The `cinder_iphone_sim` target seems to build defaulting to the `iphone` SDK when built via `fullbuild.sh`, yielding bad simulator binaries. The tweak here just specifies the SDK for both the `cinder_iphone` and `cinder_iphone_sim` targets to make sure they build as expected - it is still possible to get oneself into trouble building these targets from Xcode with an unintended device vs simulator selection.

### DIFF
--- a/xcode/fullbuild.sh
+++ b/xcode/fullbuild.sh
@@ -3,5 +3,12 @@
 CURRENT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 CINDER_XCODEPROJ="${CURRENT_DIR}/cinder.xcodeproj"
 
-xcrun xcodebuild -project ${CINDER_XCODEPROJ} -alltargets -configuration Release $@
-xcrun xcodebuild -project ${CINDER_XCODEPROJ} -alltargets -configuration Debug $@
+# OS X
+xcrun xcodebuild -project ${CINDER_XCODEPROJ} -target cinder -configuration Release $@
+xcrun xcodebuild -project ${CINDER_XCODEPROJ} -target cinder -configuration Debug $@
+
+# iOS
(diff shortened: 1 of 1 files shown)


pizthewiz:
Some quick verification:

```sh
 ~/s/C/Cinder 🍔  ./xcode/fullbuild.sh; and lipo -info lib/*.a
[…]
Architectures in the fat file: lib/libcinder-iphone-sim.a are: i386 x86_64 
Architectures in the fat file: lib/libcinder-iphone-sim_d.a are: i386 x86_64 
Architectures in the fat file: lib/libcinder-iphone.a are: armv7 arm64 
Architectures in the fat file: lib/libcinder-iphone_d.a are: armv7 arm64 
input file lib/libcinder.a is not a fat file
input file lib/libcinder_d.a is not a fat file
Non-fat file: lib/libcinder.a is architecture: x86_64
Non-fat file: lib/libcinder_d.a is architecture: x86_64
```

richardeakin:
Yea this seems like the simplest solution to this problem for now, thanks for making the changes.

PR #955, opened 2015-06-01
Add no optimization option to Objloader

araid:
I added the option to not optimize vertices in ``ObjLoader``.
Instead of using a separate function, like it was in 0.8.6, I'm forcing the existing variable ``forceUnique`` to ``true`` for every vertex, when the user chooses to not optimize.
It's a pretty minimal change, although it might not be the fastest way, not sure.

I also fixed a namespace bug in the ObjLoader sample.

### DIFF
--- a/samples/_opengl/ObjLoader/src/ObjLoaderApp.cpp
+++ b/samples/_opengl/ObjLoader/src/ObjLoaderApp.cpp
@@ -93,7 +93,7 @@ void ObjLoaderApp::writeObj()
 	fs::path filePath = getSaveFilePath();
 	if( ! filePath.empty() ) {
 		console() << "writing mesh to file path: " << filePath << std::endl;
-		writeObj( writeFile( filePath ), mMesh );
+		ci::writeObj( writeFile( filePath ), mMesh );
 	}
 }
 
--- a/src/cinder/ObjLoader.cpp
+++ b/src/cinder/ObjLoader.cpp
@@ -402,7 +402,7 @@ void ObjLoader::loadGroupNormalsTextures( const Group &group, map<VertexTriple,i
     bool hasColors = mMaterials.size() > 0;
 	for( size_t f = 0; f < group.mFaces.size(); ++f ) {
 		vec3 inferredNormal;
-		bool forceUnique = !mOptimizeVertices;
+		bool forceUnique = ! mOptimizeVertices;
 		Color rgb;
 		if( hasColors ) {
 			const Material *m = group.mFaces[f].mMaterial;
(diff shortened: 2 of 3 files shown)


samples/_opengl/ObjLoader/src/ObjLoaderApp.cpp
```diff
@@ -93,7 +93,7 @@ void ObjLoaderApp::writeObj()
 	fs::path filePath = getSaveFilePath();
 	if( ! filePath.empty() ) {
 		console() << "writing mesh to file path: " << filePath << std::endl;
-		writeObj( writeFile( filePath ), mMesh );
+		ci::writeObj( writeFile( filePath ), mMesh );
```
richardeakin: Some minor nits following: `ci::` prefix isn't necessary, we generally don't use these in samples and instead put a `using namespace ci` at the top.

src/cinder/ObjLoader.cpp
```diff
@@ -402,7 +402,7 @@ void ObjLoader::loadGroupNormalsTextures( const Group &group, map<VertexTriple,i
     bool hasColors = mMaterials.size() > 0;
 	for( size_t f = 0; f < group.mFaces.size(); ++f ) {
 		vec3 inferredNormal;
-		bool forceUnique = false;
+		bool forceUnique = !mOptimizeVertices;
```
richardeakin: Space goes between `!` and variables, in the two cases below as well (I'm going to add this to the CONTRIBUTING.md guide, since Andrew keeps reminding us that this is the preferred way).

samples/_opengl/ObjLoader/src/ObjLoaderApp.cpp
araid: I know, but this one is necessary or the sample doesn't compile, since writeObj() is already defined as a function of the app.

src/cinder/ObjLoader.cpp
araid: Cool, didn't know about that specific guideline.

richardeakin:
Thanks for PR'ing this @araid. other than the style nits I mentioned, it looks good from my end and works as expected.

As an aside, and certainly not something that needs to be included in this PR, it seems beneficial to at an ObjLoader::Options() now, the following is hardly readable at the callsite:

```cpp
ObjLoader loader( dataSource, true, true, false );
```

samples/_opengl/ObjLoader/src/ObjLoaderApp.cpp
richardeakin: Ah, of course.. missed that.

Issue #949, opened 2015-05-29
Make scoped objects movable


notlion:
After spending way too long tracking down a bug that amounted to 2 characters (`ScopedTextureBind(tex);` vs `ScopedTextureBind t(tex);`) I decided to use the `auto a = A()` form for constructing scoped objects. Using auto makes it impossible to have that type of bug.

```cpp
auto t = ScopedTextureBind(tex);
```

Thing is, ScopedTextureBind is not move-constructible and neither are the other ScopedXXX objects.

Would you guys would be open to adding default move constructors to these? It would require also adding one to Noncopyable.

I'd be willing to make a PR for this if people are open to it.

andrewfb:
Sorry for the delay on this. I don't see a disadvantage to this other than some maintenance. I think we'd be up for a PR for it.

notlion:
Awesome. I just made the changes in PR #965.

notlion:
Closed the PR because of lack of default constructors in VC2013. Implementation would have been too ugly. I'll be careful to add variable names for now ;)

PR #965, opened 2015-06-05
Add default move constructors to gl::Scoped objects

notlion:
This is in response to issue #949. It allows initialization via the `auto scoped = ScopedXxx()` form which can be less error prone. I've tested auto-constructing every scoped object, and converted every use in my own code to that form.

Whitespace edits are separated into their own commit, so feel free to cherry pick that out if you want.

### DIFF
--- a/include/cinder/Noncopyable.h
+++ b/include/cinder/Noncopyable.h
@@ -28,6 +28,7 @@ namespace cinder {
 struct Noncopyable {
   protected:
 	Noncopyable() = default;
+	Noncopyable( Noncopyable&& ) = default;
 	~Noncopyable() = default;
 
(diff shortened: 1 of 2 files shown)


andrewfb:
This looks good except for some sad news - Visual C++ 2013, which is still our baseline on the PC, doesn't support default move constructors.

notlion:
Oh crumbs. I guess if it was this simple you guys might have done it already. I'll look into fixing this today.

notlion:
After partially implementing this and then thinking about it for a while I've decided to wait. Adding manual implementations of the default move constructors would bloat the Scoped classes with a lot of semantic noise. I'm going to close this until VC2014 is the baseline.

PR #966, opened 2015-06-05
Adds deleter to Texture*::Format; changes clean bounds semantics

andrewfb:
This address 2 issues; the first is my approach for #542, adding an optional custom deleter to the Format of the various Texture types. Second, this makes "clean bounds" on Texture2d the default mode of thinking, and the 'actual' size specialized. For the average use case none of this matters, but it makes having a clean bounds distinct from the full Texture2d behave more naturally in more situations.


paulhoux:
Without having tested it, the design looks good to me.

PR #967, opened 2015-06-05
Expose display name

pizthewiz:
Propose adding a method to fetch the display name:
```C++
//! Returns the display's name or an empty string if unavailable.
std::string		getName() const { return mName; }
```

### OS X
Now that Cinder links to the IOKit framework on the OS X side, we can fetch and expose the display name without any additional dependencies. Implementation tested on:
:heavy_check_mark: OS X 10.10.4 / Xcode 6.4 / `x86_64`
:heavy_check_mark: OS X 10.8.5 / Xcode 5.1.1 / `x86_64` & `i386` (both in a VM)
and verified by adding the following to `setup()` and checking the output:
```C++
for( auto &display : Display::getDisplays() ) {
    console() << display->getName() << std::endl;
}
```

It is worth noting that an "AirPlay Display", the additional display made available by an extended desktop to an Apple TV, does not provide a display name.

### iOS
Probably not relevant as I don't think iOS has named displays.

### MSW
:question: Windows 8.1 / VS 2013 Express

An MSW implementation is provided but hasn't been tested. Though I did get Cinder building in a VM, VMWare Fusion unfortunately only supports OpenGL 2.1 (as does Parallels), so while Cinder and my sample app compile, it dies on launch failing to create a renderer. Could someone with a more reasonable setup give it a look; I'm also not 100% sure if my `TCHAR` to `std::string` conversion is legit either.

### WinRT
Probably relevant though I'm not providing an implementation here (yet).

### DIFF
--- a/include/cinder/Display.h
+++ b/include/cinder/Display.h
@@ -39,7 +39,7 @@ typedef std::shared_ptr<class Display> 	DisplayRef;
 
 class Display {
   public:
-	Display() : mArea( Area::zero() ), mBitsPerPixel( 0 ), mContentScale( 1.0f ), mName( "" ) {}
+	Display() : mArea( Area::zero() ), mBitsPerPixel( 0 ), mContentScale( 1.0f ), mName( "" ), mNameDirty ( true ) {}
 	virtual ~Display() {}
 
 	//! Returns the width of the screen measured in points
@@ -65,7 +65,7 @@ class Display {
 	bool	contains( const ivec2 &pt ) const { return mArea.contains( pt ); }
 
 	//! Returns the display's name or an empty string if unavailable.
-	std::string		getName() const { return mName; }
+	virtual std::string		getName() const { return mName; }
 
 	//! Returns the system's primary display
 	static DisplayRef						getMainDisplay();
@@ -82,10 +82,11 @@ class Display {
 	}	
 
   protected:
-	Area			mArea;
-	int				mBitsPerPixel;
-	float			mContentScale;
-	std::string		mName;
+	Area				mArea;
+	int					mBitsPerPixel;
+	float				mContentScale;
+	mutable std::string	mName;
+	mutable bool		mNameDirty;
 };
 
 } // namespace cinder
--- a/include/cinder/Display.h
+++ b/include/cinder/Display.h
@@ -39,6 +39,7 @@ typedef std::shared_ptr<class Display> 	DisplayRef;
 
 class Display {
   public:
+	Display() : mArea( Area::zero() ), mBitsPerPixel( 0 ), mContentScale( 1.0f ), mName( "" ) {}
 	virtual ~Display() {}
 
 	//! Returns the width of the screen measured in points
@@ -63,6 +64,9 @@ class Display {
 	//! Returns whether the Display's coordinates contain \a pt.
 	bool	contains( const ivec2 &pt ) const { return mArea.contains( pt ); }
 
+	//! Returns the display's name or an empty string if unavailable.
+	std::string		getName() const { return mName; }
+
 	//! Returns the system's primary display
 	static DisplayRef						getMainDisplay();
 	//! Returns a vector of all displays connected to the system
@@ -78,9 +82,10 @@ class Display {
 	}	
 
   protected:
-	Area		mArea;
-	int			mBitsPerPixel;
-	float		mContentScale;
+	Area			mArea;
+	int				mBitsPerPixel;
+	float			mContentScale;
+	std::string		mName;
 };
 
 } // namespace cinder
--- a/include/cinder/app/cocoa/PlatformCocoa.h
+++ b/include/cinder/app/cocoa/PlatformCocoa.h
@@ -149,6 +149,8 @@ class DisplayMac : public Display {
 	NSScreen*			getNsScreen() const;
 	CGDirectDisplayID	getCgDirectDisplayId() const { return mDirectDisplayId; }
 
+	std::string			getName() const override;
+
   protected:	
 	static void	displayReconfiguredCallback( CGDirectDisplayID displayId, CGDisplayChangeSummaryFlags flags, void *userInfo );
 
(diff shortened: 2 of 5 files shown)


richardeakin:
Thanks for this work, Piz.  Couple things to note:

* I think it would be ideal if things like display names, which are mainly useful only for debugging / logging / etc, were opt-in performance wise so we can keep the init process as slim as possible. The approach I took with `ci::audio::Device` is to [lazy load the name string](https://github.com/cinder/Cinder/blob/master/src/cinder/audio/Device.cpp#L95-L100), so the info that we absolutely need (or is readily available) is loaded right when the devices are first iterated, but additional information comes later. I also took the same approach with an experimental [DisplayMac::getName()](https://github.com/richardeakin/Cinder/commit/9a6e0ac1532aca30c8440c523d4b4d9457871a23#diff-aa16a3ba0a4e3e61205cb3feedd5700aR565) implementation I did during the app_refactor phase (wasn't cross platform yet).
* I see you're using `CGDisplayIOServicePort()` on the mac, which has unfortunately been deprecated since OS X 10.9. I found a somewhat long-winded workaround, which can also found in [this commit](https://github.com/richardeakin/Cinder/commit/9a6e0ac1532aca30c8440c523d4b4d9457871a23).

Haven't tried at the windows impl yet, but will in the near future. Looks much more straight forward there. In any event thanks for taking a stab on that platform.

richardeakin:
Also I just [updated windowTest](https://github.com/cinder/Cinder/commit/a6bc83034514151a9ca7487c26d94e2b80932a2f), which is what we've been using to test functionality related to `app::Window`. I propose we change change the log in the setup to:

```cpp
	for( auto display : Display::getDisplays() )
		CI_LOG_V( "display name: '" << display->getName() << "', bounds: " << display->getBounds();
```

pizthewiz:
Lazy load sounds like a good idea, I'll take a quick pass at that soon. Weird, `CGDisplayIOServicePort()` was deprecated in 10.9 but I could have sworn the deprecated attribute was removed from the SDKs in Xcode 6.0.0 and is now back:
```sh
 ~/s/C/Cinder 🍔  
cat /Applications/Xcode-beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk/System/Library/Frameworks/CoreGraphics.framework/Headers/CGDisplayConfiguration.h | grep -n2 'CGDisplayIOServicePort'
353-/* Return the IOKit service port of a display. */
354-
355:CG_EXTERN io_service_t CGDisplayIOServicePort(CGDirectDisplayID display)
356-  CG_AVAILABLE_BUT_DEPRECATED(__MAC_10_2, __MAC_10_9,
357-    __IPHONE_NA, __IPHONE_NA);
 ~/s/C/Cinder 🍔  
cat /Applications/Xcode-beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/System/Library/Frameworks/CoreGraphics.framework/Headers/CGDisplayConfiguration.h | grep -n2 'CGDisplayIOServicePort'
370-/* Return the IOKit service port of a display. */
371-
372:CG_EXTERN io_service_t CGDisplayIOServicePort(CGDirectDisplayID display)
373-    CG_AVAILABLE_BUT_DEPRECATED(__MAC_10_2, __MAC_10_9,
374-                                __IPHONE_NA, __IPHONE_NA);
```

Particularly annoying that a replacement API still hasn't been recommended and I filed a (likely duplicate) bug report rdar://problem/21376104 asking for replacement API.

It looks like GLFW [continues to use `CGDisplayIOServicePort()`](https://github.com/glfw/glfw/blob/6d5753c54863877020a4f3d12960e417fb04824b/src/cocoa_monitor.m#L50-53) and just takes note, we could do something similar and suppress the clang deprecation warning, or we could jump on a reimplementation - if we did choose to reimplement we should probably jump on your branch [display_mac_logging](https://github.com/richardeakin/Cinder/tree/display_mac_logging) since it has lazy loading and a `CGDisplayIOServicePort()` no?

pizthewiz:
To add to the strangeness, neither Xcode 6.3.2 nor Xcode 7 beta 1 displays a deprecation warning for `CGDisplayIOServicePort()` use when compiling…

pizthewiz:
Just rebased the branch to pick up the windowTest updates and added a commit to log the display name in windowTest.

richardeakin:
I noticed there wasn't a warning for the deprecation either, which is strange since it is clearly marked as such. Well, I wouldn't mind going with the more straightforward but deprecated approach for now anyway, especially if it isn't adding a warning. I'll discuss with Andrew tomorrow in the office.

richardeakin:
We talked about it and it doesn't seem worth jumping through the hoops to avoid using CGDisplayIOServicePort, if it ends up missing in a future sdk version then either we can use the workaround then or hopefully they'll provide a replacement. So the way you're doing it now looks good.

pizthewiz:
Great, I'll make the change to defer display name fetch tonight.

pizthewiz:
Hmm, one catch with the lazy display name fetching is that an empty string is a valid display name value, so if the lazy fetching occurred based on the empty string, some displays would constantly re-fetch. We could create a static hash of ID -=> name, but that's definitely a lot more machinery.

richardeakin:
Good point. I'd just add a bool mNameDirty and init to to false in Device's constructor.

andrewfb:
Just a quick note, I tested this on MSW; the current version returns something that's not especially descriptive "..//DISPLAY1". However I tracked down a different technique that I can easily append to this once the lazy fetch is implemented.

pizthewiz:
@richardeakin - I might be reversing the idiom, but wouldn't the dirty member var  be initialized to `true` to force fetch on the first access? Also, since the `getName` method would fetch and assign `mName`, should the method then use the _Conceptual Constness_ principal and be `const` with a `mutable` `mName` and `mNameDirty`?

richardeakin:
Yes and yes (we use mutable for cached / lazyloading vars throughout).

pizthewiz:
Quickly tested windowTest on OS X 10.11 / Xcode 6.3.2, I hope I captured the deferred name fetching intent.

@andrewfb - If you could patch up the MSW version that would be much appreciated, sorry I left it neglected.

andrewfb:
Merged - thanks for all your work on this @pizthewiz...

pizthewiz:
Happy to help out - I'll do a quick PR to add the lazy fetch for the MSW side.

PR #973, opened 2015-06-08
Deferred shading sample updates

BanTheRewind:
Fixed floor in basic sample.

High quality is off by default in the advanced sample. This should deliver equivalent performance to the old "normal" mode. High quality is a bit too heavy for older machines. Fortunately the visual difference is only subtle.

This PR also includes some light scattering optimizations from a previous PR.


richardeakin:
... and the floor returns.  Advanced sample is also showing a nice 42fps here when I fire it up, thanks for the updates!

BanTheRewind:
np

PR #980, opened 2015-06-11
Use override specificer to quiet warnings

pizthewiz:
Xcode 7 beta 1 warns on functions missing an explicit override specifier. Compilation tested on all 3 targets with Xcode 6.3.2 and Xcode 7 beta 1 on OS X 10.11.

### DIFF
--- a/include/cinder/app/Renderer.h
+++ b/include/cinder/app/Renderer.h
@@ -78,13 +78,13 @@ class Renderer {
 #if defined( CINDER_COCOA )
 	#if defined( CINDER_MAC )
 		virtual void	setup( CGRect frame, NSView *cinderView, RendererRef sharedRenderer, bool retinaEnabled ) = 0;
-		virtual CGContextRef			getCgContext() { throw; } // the default behavior is failure
 		virtual CGLContextObj			getCglContext() { throw; } // the default behavior is failure
 		virtual CGLPixelFormatObj		getCglPixelFormat() { throw; } // the default behavior is failure
 	#elif defined( CINDER_COCOA_TOUCH )
 		virtual void		setup( const Area &frame, UIView *cinderView, RendererRef sharedRenderer ) = 0;
 		virtual bool		isEaglLayer() const { return false; }
 	#endif
+	virtual CGContextRef	getCgContext() { throw; } // the default behavior is failure
 
 	virtual void	setFrameSize( int width, int height ) {}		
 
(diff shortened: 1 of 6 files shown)


include/cinder/app/Renderer.h
```diff
@@ -78,13 +78,13 @@ class Renderer {
...
 	#if defined( CINDER_MAC )
 		virtual void	setup( CGRect frame, NSView *cinderView, RendererRef sharedRenderer, bool retinaEnabled ) = 0;
-		virtual CGContextRef			getCgContext() { throw; } // the default behavior is failure
 		virtual CGLContextObj			getCglContext() { throw; } // the default behavior is failure
 		virtual CGLPixelFormatObj		getCglPixelFormat() { throw; } // the default behavior is failure
 	#elif defined( CINDER_COCOA_TOUCH )
 		virtual void		setup( const Area &frame, UIView *cinderView, RendererRef sharedRenderer ) = 0;
 		virtual bool		isEaglLayer() const { return false; }
 	#endif
+	virtual CGContextRef	getCgContext() { throw; } // the default behavior is failure
```
pizthewiz: This function is overridden in `Renderer2d` when building `CINDER_COCOA_TOUCH` so it needed to be moved out of the `CINDER_MAC` exclusive block.

richardeakin:
The general consensus is that we move to using `override` throughout so I'm sure this is good here, however we've developed a bit of a policy to drop the extra `virtual` specifier since it is redundant in the wake of marking as overridden. We should probably add a note to the style guide for this, but the policy has been as follows:

* `virtual` is used where you first declare the method.
* `override` with no `virtual` is used for anything that overrides a virtual method (including overriding an abstract virtual method).

I'm actually liking that the new Xcode warns on this, because it will help us follow this policy.

pizthewiz:
Should the removal of `virtual` from any function with `override` be handled separately? I looked at just removing it from the functions I added to `override` to and it seems like it might stick out a bit in cases like `VaoImplEs` where it probably needs to be added more evenly?

richardeakin:
Hm good point, in those cases they need the style change made overall. A PR around that would also be very much appreciated. :) Personally I've been trying to remove the 'virtual's as I add 'override' so I can do it more incrementally, but yea not so fitting in this case. I'll hit the green button as is.

Of course, the reason why this wasn't always the policy was because much of the codebase was written when the `override` specifier wasn't available on all supported compilers.

pizthewiz:
We might be able to craft a regex in Xcode that could find all the `virtual` use with `override`, I'll see what I can do ;0)

richardeakin:
Or if you use something like sublime text, you could get the MSW stuff too. I'd be happy to test the changes there if you can't get on a windows machine.

PR #981, opened 2015-06-12
GlslProg::findUniform() v3

richardeakin:
I ran into a few issues with the existing `GlslProg::findUniform()` implementation, this is a redux that aims to fix those problems:

* Avoids exc_bad_access on the result uniform when the requested name has a left square bracket, but a match isn't found in `mUniforms` (bad access was on the `*resultLocation = ret->mLoc + stoi( ... )` line.
* Handles the case when an array uniform is requested, but a location offset isn't necessary because we found an exact match in mUniforms
* Handles bad struct member names, ex. "uTestStruct[2].blah" in the test app.
* Overall, tried to document and make the string matching logic easier to follow / more robust.

This has been tested on Mac only so far, MSW testing still needed.  Some suggestions / discussion around possible future improvements:

* Much of the logic is based on what `GL_ACTIVE_ATTRIBUTES` is returning, to me at least it feels like we are reverse engineering each hardware implementation, trying to cover each case that we run into. I was wondering if it isn't possible to have a fallback, where if findUniform() returns null, we still try to ask OpenGL if it knows about a uniform by that name, the old school way. This way we'd still get a result even if our parsing logic doesn't cover a specific case, and there would be no extra cost when we are covering the case.
* Currently the string parsing logic in findUniform() happens each time the user tries to set a uniform, yet the bulk of the logic is for array uniforms, which are much less often used.  I'm wondering if we can't add an optimization to avoid that, like flag the uniform as `UNIFORM_USER_DEFINED_ARRAY` or something, so we only do the extra string parsing on those types. Or think of a clever way to cache the results of the findUniform() lookup so it is done as few times as possible.

Please let me know if I've missed any other edge cases or mucked something up, I've got a good understanding of that portion of the code now so I can still add further work if necessary.

### DIFF
--- a/src/cinder/gl/GlslProg.cpp
+++ b/src/cinder/gl/GlslProg.cpp
@@ -987,10 +987,10 @@ const GlslProg::Attribute* GlslProg::findAttrib( geom::Attrib semantic ) const
 
 const GlslProg::Uniform* GlslProg::findUniform( const std::string &name, int *resultLocation ) const
 {
-	const Uniform* ret = nullptr;
+	const Uniform* resultUniform = nullptr;
 	for( const auto & uniform : mUniforms ) {
 		if( uniform.mName == name ) {
-			ret = &uniform;
+			resultUniform = &uniform;
 			break;
 		}
 	}
@@ -1001,7 +1001,7 @@ const GlslProg::Uniform* GlslProg::findUniform( const std::string &name, int *re
 
 	// if we didn't find an exact match, look for the array version in the active uniforms list
 	bool needsLocationOffset = false;
-	if( ! ret ) {
+	if( ! resultUniform ) {
 		for( const auto &uniform : mUniforms ) {
 			size_t activeUniformLeftSquareBracket = uniform.mName.find( '[' );
 			// skip match detection if this active uniform isn't an array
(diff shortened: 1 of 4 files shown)


ryanbartley:
These seem like good changes. 

As far as the first suggestion goes, can you be a little more specific of what you're looking for in that functionality? 

As far as the second suggestion goes, I agree. It could be a simple change. Since there are two ways to buffer a uniform, an impl to buffer one uniform and an impl to buffer an array of uniforms in the interface, maybe we should leave findUniform as the light find function and add a findUniformArray which contains the parsing logic.

richardeakin:
What I mean by the first suggestion is that if our custom `findUniform()`  function fails, still try to get the location using glGetUniformLocation(), like [we did in 0.8.6](https://github.com/cinder/Cinder/blob/release_v0.8.6/src/cinder/gl/GlslProg.cpp#L264) only now it would be a fallback when caching is enabled. We could also make it the immediate route when `DISABLE_UNIFORM_CACHING` is defined.

The second suggestion seems a bit involved to me without thinking about it in too much depth, because we need to cache each search, or at least whether it falls into the array category or 'regular' uniform. Cases like `uniform( "uColors", &mColors.r, 3 )`, where the requested name is non-array, but active uniform is called "uColors[0]", really complicate this.  In any event I suggest we get it right before adding a different layer of caching, or making the existing one any more complex.

andrewfb:
Thanks Rich; this looks good to me. I also added an early-out test 3956f3a9cbb3145868a6c815179afdcb1bc60e7f (smaller change than it appears to be due to a tab-outdent).

PR #982, opened 2015-06-12
Support HUSL color space

heisters:
HUSL is a useful color space for generating perceptually uniform colors. Its interface ends up being similar to the already-supported HSV, but it outputs a more uniform distribution.

````
ColorAf( CM_HUSL, randFloat(), 1.f, 0.5f, 1.f )
````

### DIFF
--- a/include/cinder/Color.h
+++ b/include/cinder/Color.h
@@ -34,7 +34,8 @@ namespace cinder {
 
 typedef	enum {
 	CM_RGB, // Red[0 - 1.0] Green[0 - 1.0] Blue[0 - 1.0]
-	CM_HSV  // Hue[0 - 1.0] Saturation[0 - 1.0] Value[0 - 1.0]
+	CM_HSV, // Hue[0 - 1.0] Saturation[0 - 1.0] Value[0 - 1.0]
+    CM_HUSL // Hue[0 - 1.0] Saturation[0 - 1.0] Lightness[0 - 1.0]
 } ColorModel;
 
 template<typename T>
--- a/include/cinder/Color.h
+++ b/include/cinder/Color.h
@@ -35,7 +35,7 @@ namespace cinder {
 typedef	enum {
 	CM_RGB, // Red[0 - 1.0] Green[0 - 1.0] Blue[0 - 1.0]
 	CM_HSV, // Hue[0 - 1.0] Saturation[0 - 1.0] Value[0 - 1.0]
-    CM_HUSL // Hue[0 - 1.0] Saturation[0 - 1.0] Lightness[0 - 1.0]
+	CM_HUSL // Hue[0 - 1.0] Saturation[0 - 1.0] Lightness[0 - 1.0]
 } ColorModel;
 
 template<typename T>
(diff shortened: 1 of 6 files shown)


heisters:
sorry, the commit history got a little wonky with a rebase against glNext.... let me know if that needs to be fixed.

andrewfb:
Interesting - I'll take a look at this in the coming days. In the meantime, for anyone else evaluating this, http://www.husl-colors.org/ is helpful. I haven't gotten too deep with this but it seems to be a human-friendly remapping of CIELUV. Essentially this leads to a perceptually uniform HSL-like color space. An easy way to see why this is important is to scroll down in the previous link to the random colors section of the page. In the comparison you can see that HUSL gives random colors of comparable brightness, while HSL (and HSV, not depicted) gives variable brightness, even though the Saturation and the Lightness values are held constant.

I'm open to this, but we need to weigh it against the fact that it's pretty far from standard, and may not have a place in Cinder core as a result. The bit I've parse seems a little ad hoc to me, though the results are useful. Additional points of comparison:
https://vis4.net/blog/posts/mastering-multi-hued-color-scales/
http://eeecon.uibk.ac.at/~zeileis/papers/Zeileis+Hornik+Murrell-2009.pdf

heisters:
While it may be non-standard, I think the draw is definitely its interface simplicity.

If it is not suitable for core, it would be nice if there were a way to register custom color spaces or something. The main reason I added it to core was so that it would be easy to switch back and forth between color spaces, and writing an adaption layer seemed like overkill.

andrewfb:
To be honest, right now I am betting that by the 0.9.1 timeframe we'll be removing Color in favor of just using vec3 / vec4 in combination with custom construction functions. In retrospect I regret not doing this in 0.9.0, but there's probably already enough breaking changes in one release, and I'm not quite positive it's the right choice yet.

To my mind, in general the primary moment I want to interact with a Color as such (rather than as a  vec3/vec4) is at its moment of creation. To that end, I think we'll end up with functions like `vec3 colorHsv()`, `vec4 colorHsvA()` and then possibly methods like `float getHsvHue( const vec3 &colorHsv )`, etc. In light of that, perhaps 2 methods `Color colorHusl()` and `ColorA colorHuslA()` would allow this to exist as a CinderBlock? I do think that HUSL as a color space may be too young for us to be comfortable including it in the Cinder core, and it's also a reasonably large amount of code as well. However I do see the value, and if it were to become more of a standard it's something we could revisit in the future.

paulhoux:
Hi, 

thanks for your work on this, I had never heard of HUSL before and it made for an interesting read. Looks like it can be useful when creating color schemes and palettes where all colors have a similar bightness and saturation. 

I think this would make a nice little Cinder block, especially if conversion from RGB/HSL to and from HUSL is done with free functions. That way, it will be compatible with the changes to the color support that Andrew is mentioning. I am not convinced it should be added to Cinder's core at this moment, as it will undoubtedly need to be changed in the near future and HUSL as a system does not seem to be adopted by a large number of industry professionals. 

-Paul

andrewfb:
I think this may be the direction we head with this - not no, but not yet. For the time being I'd like to wait and see if HUSL takes off in the wild. Additionally, if you do take the time to turn this into a CinderBlock and it ultimately becomes something the Cinder community embraces, even if the larger graphics community does not, that would be good cause to revisit this as well. I'm going to go ahead and close this now though as it's the last PR against *glNext*, which I'll now remove too now that it's been merged with *master*.

PR #983, opened 2015-06-13
WIP: Remove virtual specifier from functions with override specifier

pizthewiz:
Based on some discussion in #980, the Cinder style now suggests that functions with the `override` specifier should also not use the `virtual` specifier as it is redundant.

To find and replace offending lines of code a regular expression was used in BBEdit and it was applied to all files in the `include`, `src`, `samples`, `test` and `blocks` subdirectories.
  Find: `virtual(\s+)(.*)(\s+)override(.*)`
  Replace: `\2\3override\4`

I quickly compiled the OS X and iOS targets and those seemed fine but Windows should be verified. Also it is worth noting that this likely messed up custom block indenting and I'm really not sure how we'd deal with that - it really seems as though it would have to be done by hand for each changed line which might be out of the scope of this particular change - maybe [ClangFormat](http://clang.llvm.org/docs/ClangFormat.html) could save the day? :sob:.

Should CONTRIBUTING.md also be updated with the new style info?

### DIFF
--- a/include/cinder/Exception.h
+++ b/include/cinder/Exception.h
@@ -34,7 +34,7 @@ class Exception : public std::exception {
 	Exception( const std::string &description );
 	virtual ~Exception() throw() {}
 
-	virtual const char* what() const throw() override	{ return mDescription.c_str(); }
+	const char* what() const throw() override	{ return mDescription.c_str(); }
 
   protected:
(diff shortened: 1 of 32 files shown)


richardeakin:
Thanks, fixed tab realigment locally and will merge into master shortly..

PR #985, opened 2015-06-14
Compute Shaders

clinthidinger:
I updated the pull request.

I did a less destructive version this time that followed the example of geometry shaders in GlslProg::Format.  Alternatively, we could have made a shader stage struct to wrap the shader string and path.  Then we could have stored that in a map keyed on the shader type GLint.  

Compute shaders can be created by doing:

mUpdateProg = gl::GlslProg::create( gl::GlslProg::Format().compute( loadAsset( "particleUpdate.cs.glsl" ) ) );

There are two samples for testing.  One is a conversion of the ParticleSphereGPU sample.  The other is an NVidia sample I did to learn more about compute shaders:

http://docs.nvidia.com/gameworks/content/gameworkslibrary/graphicssamples/opengl_samples/computeparticlessample.htm

https://github.com/NVIDIAGameWorks/OpenGLSamples/tree/master/samples/es3aep-kepler/ComputeParticles

I also did a transform feedback version of this one here: 
https://github.com/clinthidinger/Cinder-Sketches/blob/master/NVidiaComputeParticles/src/NVidiaComputeParticlesApp.cpp

There are also two functions that I originally had in gl/Context.  getMaxComputeWorkGroupSize() and getMaxComputeWorkGroupCount() currently exists in the ParticleSphereCS app source.

Please let me know if I need to do any more work on this for it to be accepted.


andrewfb:
This looks really great - thanks for this contribution. I've created a merge PR #989 with a few modifications. Overall this looks excellent though, and barring any objections I'll merge it soon.

clinthidinger:
Thanks!

PR #992, opened 2015-06-19
disabling LoggerSystem on winrt.

richardeakin:
as EventLog API doesn't appear to be available there. This just gets the project building again, while LoggerSystem::write() will do nothing for now. @lucasvickers Can you think of a solution we can use on that platform?  I haven't spent too much time looking, but I did see a [suggestion here](https://social.msdn.microsoft.com/Forums/windowsapps/en-US/73c7fb3a-2183-4d60-b2b1-ff4aa9286700/systemdiagnosticseventlog-replacement) to use COM and EventWrite(), not sure what the difference is between that method and the one we're using.

### DIFF
--- a/include/cinder/Log.h
+++ b/include/cinder/Log.h
@@ -135,7 +135,7 @@ class LoggerBreakpoint : public Logger {
 	Level	mTriggerLevel;
 };
 
-//! Provides 'system' logging support. Uses syslog on platforms that have it, on MSW uses Windoes Event Logging
+//! Provides 'system' logging support. Uses syslog on platforms that have it, on MSW uses Windows Event Logging. \note Does nothing on WinRT.
 class LoggerSystem : public Logger {
 public:
 	LoggerSystem();
--- a/src/cinder/Log.cpp
+++ b/src/cinder/Log.cpp
@@ -593,9 +593,11 @@ LoggerSystem::~LoggerSystem()
 
 void LoggerSystem::write( const Metadata &meta, const std::string &text )
 {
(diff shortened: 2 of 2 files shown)


lucasvickers:
[EventWrite](https://msdn.microsoft.com/en-us/library/windows/apps/aa363752.aspx) looks straight forward enough, but you'll have to show me how to build/test WinRT stuff (haven't touched it).

Side note, does the syslogging work in iOS and OS X, or do we have the same type of desktop/mobile logging issue?

richardeakin:
Okay, I'm going to merge this as it fixes the winrt build, and open a new issue for trying to use `EventWrite` on winrt.  The only difference for building winrt stuff is that you currently have to use the [vc2013_winrt](https://github.com/cinder/Cinder/tree/master/vc2013_winrt) project, since we haven't figured out how to combine the mode it needs into the main Visual Studio solution we have (in vc2013 folder).  Of course you have to be on a Windows 8+ machine, and then I'd either run one of the existing winrt samples ([BasicApp](https://github.com/cinder/Cinder/tree/master/samples/BasicApp/vc2013_winrt) has one), or we should add a winrt configuration to test/DebugTest.

Re LoggerSystem on iOS /w syslog: yes it works, if you run an app on an iPhone, open up 'Devices' in Xcode, and click the near-hidden upwards triangle at the bottom of the window, you see some logs like the following:

screenshot 2015-06-21 17 59 03

lucasvickers:
Sounds good.  I'll be doing some windows work this week so I'll look into it.  I'll try to get to it first.

Issue #879, opened 2015-05-05
Cinder Rotation of Camera when WorldUp() changed


bradrothenberg:
Hi,

You could potentially argue that this is not an issue, but when you change the worldUp() of a mayaCam (now Camera), the axes for rotation do not switch to adjust for the new world up. It would seem that if you change the worldUp() of the camera, the axes should update so the mouse inputs seem correct to a user.

This also has to do with the best practice to switch between Y-up & Z-up...

Bradley

paulhoux:
I'll take a look. Thanks for reporting, Bradley.

paulhoux:
Hi Bradley,

so... when I call ```CameraPersp::setWorldUp()```, the orientation of the camera is changed accordingly. If I then call ```CameraPersp::getBillboardVectors()```, it correctly gives me the up and right vectors. So that all seems to work.

Could you show me in code what you're trying to do, what you'd expect and what you get instead?

bradrothenberg:
Hi Paul,

The issue I was having was with how the CameraUI responds to mouse inputs once you change the SetWorldUp of the camera -- So if I set the SetWorldUp to {0,0,1}, Z should now be up. The Billboard vector is correct, but when passing the MouseEvent info to the CameraUI, the Camera still wants to rotate as if Y is up.

Thanks!

Bradley

paulhoux:
Ah, OK, thanks for your reply. Will look at it again tonight.

bradrothenberg:
Paul,

Did you get a chance to look at this??

it looks like the issue is in line 124 of CameraUi.cpp :

```c++
		bool invertMotion = ( mInitialCam.getOrientation() * glm::vec3( 0, 1, 0 ) ).y < 0.0f;
```

I think glm::vec3(0,1,0) should be replaced with whatever the camera's z-axis is set to...I'm going to try this now...

Bradley

andrewfb:
Addressed by #995

PR #995, opened 2015-06-21
CameraUi Tumble Orientation

bradrothenberg:
Issue #879:

I updated the tumbling operation to take the camera's worldUp value, rather than always glm::vec3{0,1,0).

This allows for the tumble functionality to remain consistent after setWorldUp() is called, changing a camera's worldUp -- WorldUp can now be switched to change between Y-up, or Z-up, & the mouse controls do not invert.


andrewfb:
Thanks for this improvement. Minor formatting consistency issues addressed in e906728 but otherwise merged as-is.

Issue #994, opened 2015-06-20
Move GL preprocessor checks to GL_HAS_XXX system


richardeakin:
Currently we're using things like `#if ! defined( CINDER_GL_ES_ 2 )` for selectively adding gl features on different platforms, but the line for what is available on OpenGL ES and OpenGL desktop is becoming more and more gray, especially with the advent of android support.  Right now the discussed solution is to add feature macros to `cinder/gl/platform.h`, like `CINDER_GL_HAS_UNIFORM_BUFFERS`, and then the individual classes like `GlslProg` or `Context` will use those to determine what API to enable.

andrewfb:
Addressed in #1006

PR #1001, opened 2015-06-28
Implement GL_HAS_XXX feature to simplify new ports

richardeakin:
This is an initial pass at addressing #994, look at the bottom of `cinder/gl/platform.h` for what I've added so far. Mostly `gl::GlslProg` and `gl::Context` have gotten updates from this, so their features are enabled (or disabled, depending on how you look at it) by what is declared in gl/platform.h.

Many things just aren't available on any of the gl es versions but are on desktop, for example `glBindFragDataLocation`. In those cases I left it as `#if ! defined( CINDER_GL_ES )`, but I'm open to other solutions as well, such as defining that as a symbol too, or defining a `CINDER_GL_DESKTOP` macro. 

I'm looking for feedback on what else I should cover during this initial pass, (@chaoticbob) specifically what could help for android.


andrewfb:
Superseded by #1006

PR #1002, opened 2015-06-29
Updated glload to 052

FictionIO:
I updated glload to 052 to fix linker errors i got with some opengl extensions. Related forum thread can be found at : https://forum.libcinder.org/#Topic/23286000002380009


andrewfb:
Thanks for this - which glload distribution is this code from?

FictionIO:
Hi Andrew,

Its from http://glsdk.sourceforge.net/docs/html/index.html

FictionIO:
Hi Andrew,

I found another annoying bug in this version. Its reported here https://bitbucket.org/alfonse/glloadgen/issue/68/gltexturepagecommitmentext-wrong-signature

I will try to find fixed more recent version. Will push another pull request hopefully today.

FictionIO:
Hi Andrew,

I found working with glloadgen very difficult and its only supports up to 4.4 OpenGL. Many recent hardware now supports 4.5. For this need i had to switch Cinder/GLNext to Glew which you can find it here https://github.com/FictionIO/Cinder/tree/glew. I don't know if its something you would consider changing but thought worth for mentioning.

paulhoux:
What are the 4.5 features that you need at the moment?

FictionIO:
Hi Paul,

Thanks the info. At the moment im using many recent extensions like Sparse Texture Arb 2.0. Many of them are working fine on 4.4 OpenGL. 

The reason i (personally) switched to glew is to bypass some bugs i reported yesterday and today. Also to support 4.5 to make sure i can utilize a more fancy extension if needed during the project. 

I found (personally) glew has more support and less bug for recent futures of OpenGL.

andrewfb:
We went through a pretty involved debate on the various loaders (forum thread here: https://forum.libcinder.org/topic/glew-instead-of-glee-in-cinder) and ultimately chose glLoad, in part because it didn't require external libraries.

We certainly don't want to discourage you from using GLEW for your own projects, but probably the best thing for Cinder would be to update to a more recent commit of glLoad: https://bitbucket.org/alfonse/glloadgen/commits/all

Judging from the commit messages it looks like 4.5 is supported now (and the bug you cited has been fixed).

FictionIO:
That would be great Andrew! Looking forward to use main repo!

andrewfb:
In 649ca77380fc33848df985a43c9b805cc4416e40 I pushed changes that parallel these (but don't add the GLX files and preserve some of the modifications we had to make to glload to make it work efficiently with Cinder). The glload build process is sufficiently complex though that I'd like to put off a full upgrade until after 0.9.0, especially since we may need to switch libraries to support ES 2/3. If there's a specific bug that's biting you though we can likely make modifications to glload manually, or hopefully GLEW addresses your immediate needs.

FictionIO:
Hi Andrew,

Thank you very much for spending your time on this. 
I'll give it a try as soon as i can. In the mean time i will stick GLEW until full glload upgrade. Looking forward to 0.9!

PR #1003, opened 2015-06-30
OS X warnings, leaks and tweaks

pizthewiz:
A few bits and bobs found via the static analyzer and from creating a new app via TinderBox - if you'd prefer they were split up I can do that too, but they all seemed so small that it might be a bit of overkill.

### DIFF
--- a/src/cinder/Clipboard.cpp
+++ b/src/cinder/Clipboard.cpp
@@ -152,7 +152,7 @@ ImageSourceRef Clipboard::getImage()
 {
 #if defined( CINDER_MAC )
 	NSPasteboard *pasteboard = [NSPasteboard generalPasteboard];
-    NSBitmapImageRep *rep = [NSBitmapImageRep imageRepWithPasteboard:pasteboard];
+    NSBitmapImageRep *rep = (NSBitmapImageRep*)[NSBitmapImageRep imageRepWithPasteboard:pasteboard];
 	if( rep )
 		return cocoa::ImageSourceCgImage::createRef( [rep CGImage] );
 	else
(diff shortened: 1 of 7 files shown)


src/cinder/qtime/QuickTimeImplAvf.mm
```diff
@@ -205,6 +205,10 @@ - (void)outputSequenceWasFlushed:(AVPlayerItemOutput *)output
 		[mAsset cancelLoading];
 		[mAsset release];
 	}
+
+	if( mPlayerVideoOutput ) {
```
pizthewiz: The conditional is unnecessary since messaging `nil` is fine, but in keeping with the proximate style.

src/cinder/Clipboard.cpp
```diff
@@ -152,7 +152,7 @@ ImageSourceRef Clipboard::getImage()
 {
 #if defined( CINDER_MAC )
 	NSPasteboard *pasteboard = [NSPasteboard generalPasteboard];
-    NSBitmapImageRep *rep = [NSBitmapImageRep imageRepWithPasteboard:pasteboard];
+    NSBitmapImageRep *rep = (NSBitmapImageRep*)[NSBitmapImageRep imageRepWithPasteboard:pasteboard];
```
pizthewiz: Hmm, missed an opportunity to fix a tab/space issue.

pizthewiz:
A couple more tweaks to update the OS X project template with Xcode 7's suggestions (double checking it is all kosher back to OS X 10.8.5/Xcode 5.1.1) and to resolve another warning coming…

andrewfb:
All of this looks good to me except the removal of `$CINDER_PATH/include` for header search paths. At least for the time being, as part of deprecating but not removing boost, users can still use boost directly. This needs to be at Header Search Paths, but not User Header Search Paths.

pizthewiz:
Ahh very good point, I'll back that rev out.

pizthewiz:
That should do it, there are 3 themes:
- Update the default OS X project template - one should no longer be prompted to upgrade settings on first open 
- Fix a couple of memory leaks called out by the static analyzer
- Quiet a warning with a cast

andrewfb:
Thanks for these...

PR #1004, opened 2015-06-30
Add an output stream operation to PolyLine.

drewish:
Makes debugging it much nicer. I'm totally open to suggestions on the formatting.


andrewfb:
I'm up to include this but I'd propose we leave out the "empty" bit. I feel like seeing `()` would be clear enough on its own and a little more consistent. I can make this change and address the minor C++ formatting issues if there are no objections.

drewish:
Feel free to make what ever changes you see fit. It was a barf pass against the 0.8.6 branch that I'd tried to get updated.

andrewfb:
Thanks for this - merged with db44dc4c25dd9df8581b57ec097487498fa1033e

drewish:
Awesome! I'll dig out some other PolyLine additions for another PR.

PR #1007, opened 2015-07-01
Add PolyLine reverse() and reversed()

drewish:
Makes it easier switch between from clockwise to counter-clockwise winding order.

Returning a copy makes for a nicer API:
```c++
intersection = PolyLine2f::calcIntersection( shapeA, shapeB );
if (intersection.size()) {
    // Reverse the winding order:
    PolyLine2f outline = intersection.front().reverse();
    // Use the outline
}
```

### DIFF
--- a/include/cinder/PolyLine.h
+++ b/include/cinder/PolyLine.h
@@ -56,6 +56,9 @@ class PolyLineT {
 
 	void		scale( const T &scaleFactor, T scaleCenter = T() );
 	void		offset( const T &offsetBy );
+	void		reverse();
+	
+	PolyLineT<T>	reversed() const;
 
 	//! Returns whether the point \a pt is contained within the boundaries of the PolyLine
 	bool	contains( const vec2 &pt ) const;
--- a/src/cinder/PolyLine.cpp
+++ b/src/cinder/PolyLine.cpp
@@ -71,6 +71,20 @@ void PolyLineT<T>::offset( const T &offsetBy )
 }
 
 template<typename T>
+void PolyLineT<T>::reverse()
+{
+	std::reverse( mPoints.begin(), mPoints.end() );
+}
+
+template<typename T>
+PolyLineT<T> PolyLineT<T>::reversed() const
+{
+	PolyLineT result( *this );
+	std::reverse( result.mPoints.begin(), result.mPoints.end() );
+	return result;
+}
+
+template<typename T>
 T linearYatX( const glm::tvec2<T, glm::defaultp> p[2], T x )
 {
 	if( p[0].x == p[1].x ) 	return p[0].y;


src/cinder/PolyLine.cpp
```diff
@@ -71,6 +71,12 @@ void PolyLineT<T>::offset( const T &offsetBy )
 }
 
 template<typename T>
+void PolyLineT<T>::reverse()
+{
+    std::reverse( mPoints.begin(), mPoints.end() );
```
drewish: My preference would be to return a copy but this matches what the `scale` and `offset` methods do.

drewish:
Some what related to this, I'd wanted to add a function to compute the bounds of the polyline, but I'm not quite sure how to handle the return type. I'm not sure how to get the types to go from a `T` of say `vec2` to `Rectf`, and much less what `dvec3` should do. What would be the 3-dimensional equivalent of `Rectd`?

richardeakin:
The 3d analog of Rect is AxisAlignedBox, although obviously it is different in some ways (not templated on the floating point type the moment, and well being an entirely different class).  It would be very tricky adding a method to PolyLine to do calculate the bounds.  I personally wouldn't mind moving to a more functional approach, so you have methods like:

```cpp
RectT calcBounds( const PolyLineT<vec2> &polyLine );
AxisAlignedBox calcBounds( const PolyLineT<vec3> &polyLine );
```

That doesn't cover dvec2, but I think we could manage that, at least it is much easier.  Also, this type of functional approach is more in line with GLM, which is the core of our math functionality at this point.

drewish:
@richardeakin interesting... yeah if i get a little time i'll try to play around with that approach. i'm stuck half way through upgrading my current app to master.

src/cinder/PolyLine.cpp
drewish: I changed this to return a copy.

include/cinder/PolyLine.h
```diff
@@ -57,6 +57,8 @@ class PolyLineT {
 	void		scale( const T &scaleFactor, T scaleCenter = T() );
 	void		offset( const T &offsetBy );
```
drewish: Curious if there's any interest in having these return copies.

drewish:
@richardeakin I'm definitely seeing the merits of having them as top level functions since it avoids a huge headache getting the types to cast from a `PolyLineT`'s `T` to the `vec3` that `AxisAlignedBox` expects. If you can suggest what file they should go in I'm happy to open up a new PR for that.

drewish:
Realized for `PolyLine2f` you can easily get the bounding box using: 
```c++
ci::Rectf calcBounds( const ci::PolyLine2f poly )
{
    return ci::Rectf( poly.getPoints() );
}

ci::Rectd calcBounds( const ci::PolyLine2d poly )
{
    return ci::Rectd( poly.getPoints() );
}
````
Perhaps we should just add a `AxisAlignedBox( const std::vector<vec3> &points )` constructor so `AxisAlignedBox calcBounds( const ci::PolyLine3f )` could be implemented the same way.

**Comment by richardeakin** [comment]
The AxisAlignedBox constructor would be one way.  Andrew has also suggested (offline) that we could add a method to PolyLineT like calcBounds( T *min, T *max ), that way it could be defined to use either vec2 or vec3 min / max.  For now I'd wait until all agree on what api we should choose, and you should be able to come up with a freestanding function for your own project fairly easily. Though thanks for trying it out, it definitally helps in hashing out what the best approach is, both for now and in the long run.

**Comment by drewish** [comment]
Yeah no rush on my part. Discovering that `RectT` could be initialized using a vector of vertices was all I needed. 

Any feedback on the `reverse()` function though?

**Comment by paulhoux** [comment]
Personally, I am against having an ```AxisAlignedBox( const std::vector<vec3> &points )``` constructor. It would be confusing as to what those points actually mean. I think the best approach is a free function to calculate the bounds of a set of points.

**Comment by drewish** [comment]
I've learned my lesson to not introduce tangents in a PR. I should have just opened a separate issue for the bounding box question ;)

**Comment by richardeakin** [comment]
Concerning whether to return copies or not, the general pattern is that the verb form (`reverse()`) will do the action in place, where as the adjective form (`reversed() const`) returns a copy, leaving the original untouched. You can see this in RectT, such as [here](https://github.com/cinder/Cinder/blob/master/include/cinder/Rect.h#L73-L76).

**Comment by drewish** [comment]
Ah, that makes sense. I'll update this PR to follow that pattern.

**Comment by drewish** [comment]
Any thoughts? Happy to make changes if they're needed.

**Comment by richardeakin** [comment]
This looks good to me. The only thing I think would be nice to change was that if you name the copied variable that will be returned `result` for consistency with the rest of the class. For the most part, that is the name that returned variables are given throughout the codebase.

File: src/cinder/PolyLine.cpp [hunk 3]
```diff
@@ -71,6 +71,20 @@ void PolyLineT<T>::offset( const T &offsetBy )
...
 template<typename T>
+void PolyLineT<T>::reverse()
+{
+	std::reverse( mPoints.begin(), mPoints.end() );
+}
+
+template<typename T>
+PolyLineT<T> PolyLineT<T>::reversed() const
+{
+	PolyLineT ret(*this);
```
richardeakin: Also, spacing:

```cpp
PolyLineT ret( *this );
```

The rest looks spot on, though.

drewish: Ah yeah good call. Fixing that and the naming of `ret` now.

richardeakin:
Thanks for the addition.

PR #1012, opened 2015-07-06
rename AppBase::isPrimaryThread() -> AppBase::isMainThread()

richardeakin:
As proposed [here](https://forum.libcinder.org/topic/rfc-minor-changing-app-isprimarythread-to-ismainthread).

Also added a freestanding method so you can use `app::isMainThread()` instead, and some minor tab realignments that probably got unaligned after switching to glm.


num3ric:
*thumbs up*

PR #1015, opened 2015-07-07
Add a signed distance function to RectT

notlion:
Sometimes it's useful to know how far a point lies inside a rectangle. Alternatively, I'd like to replace the plain `distance` function. It currently returns 0 for points inside which could be achieved just by min(x,0)'ing the signed version. Although, I can understand this might be unexpected for people used to the current behavior.

A 3d version should probably be added to AxisAlignedBox. Happy to do that as well.

Demo Sketch:
```cpp
#include "cinder/app/App.h"
#include "cinder/gl/gl.h"
#include "cinder/Vector.h"

using namespace ci;
using namespace ci::app;
using namespace std;

class TestRectSignedDistApp : public App {
public:
  void draw() override;
};

void TestRectSignedDistApp::draw() {
  gl::clear(Color(0, 0, 0));

  float t = getElapsedSeconds();
  float s = glm::length(vec2(getWindowSize())) / 4.0f;
  Rectf rect = { glm::rotate(vec2(s, 0), t * 0.8234f),
                 glm::rotate(vec2(0, s * 0.75f), t * 0.3345f) };
  rect.offset(getWindowCenter());
  rect.canonicalize();

  vec2 p = getMousePos() - getWindowPos();
  auto d = rect.distance(p);
  auto sd = rect.distanceSigned(p);

  gl::color(1, 1, 1);
  gl::drawStrokedRect(rect);

  gl::color(0, 1, 1);
  gl::drawStrokedCircle(p, glm::abs(sd), 64);
  gl::drawString("sd: " + std::to_string(sd), p + vec2(0, 40));

  gl::color(1, 0, 0);
  gl::drawStrokedCircle(p, d, 64);
  gl::drawString("d: " + std::to_string(d), p + vec2(0, 20));
}

CINDER_APP(TestRectSignedDistApp, RendererGl)
```

### DIFF
--- a/include/cinder/Rect.h
+++ b/include/cinder/Rect.h
@@ -86,6 +86,8 @@ class RectT {
 	T		distance( const Vec2T &pt ) const;
 	//! Returns the squared distance between the point \a pt and the rectangle. Points inside the rectangle return \c 0.
 	T		distanceSquared( const Vec2T &pt ) const;
+	//! Returns the signed distance from the rectangle to the point \a pt. Distances inside the rectangle are negative.
+	T		distanceSigned( const Vec2T &pt ) const;
 
 	//! Returns the nearest point on the Rect \a rect. Points inside the rectangle return \a pt.
 	Vec2T		closestPoint( const Vec2T &pt ) const;
--- a/src/cinder/Rect.cpp
+++ b/src/cinder/Rect.cpp
@@ -262,8 +262,17 @@ T RectT<T>::distanceSquared( const Vec2T &pt ) const
 	else if( pt.x > x2 ) squaredDistance += ( pt.x - x2 ) * ( pt.x - x2 );
 	if( pt.y < y1 ) squaredDistance += ( y1 - pt.y ) * ( y1 - pt.y );
 	else if( pt.y > y2 ) squaredDistance += ( pt.y - y2 ) * ( pt.y - y2 );
-	
+
 	return squaredDistance;
+}
+
+template<typename T>
+T RectT<T>::distanceSigned( const Vec2T &pt ) const
+{
+	Vec2T halfSize = getSize() * static_cast<T>(0.5);
+	Vec2T center = getUpperLeft() + halfSize;
+	Vec2T d = glm::abs(pt - center) - halfSize;
+	return glm::min<T>(glm::max(d.x, d.y), 0) + length(glm::max<T>(d, 0));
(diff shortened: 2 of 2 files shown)


notlion:
In my benchmarks on x86_64/release `distanceSigned` is pretty comparable speed wise, taking about 0.9 to 1.1 times as long as `distance`. It's slightly slower for points outside of the rectangle and slightly faster for points inside. Bench sketch is [here](https://gist.github.com/notlion/620ceaea62d47e50df0c).

richardeakin:
It does indeed seem useful to know the distance of a point to the closest edge of a Rect. To open a discussion about what api choice is best, I'm trying to think how you'd use this in practice - would you check for the result being negative, then branch and negate the returned distance?

One thing that comes to mind is that when evaluating the return of RectT::distance(), checking `== 0` isn't very good, seems `> 0` gives more of an indication that the point is outside (if Ryan's proposed changes were to happen).

Looking at the impls, both seem to be doing about the same math, most importantly they are both use a `sqrt`.

notlion:
To get the unsigned distance to the closest edge you'd simply do a `glm::abs(rect.distanceSigned)`.

My current use case is the implementation of Starcraft-style edge panning where the camera moves when the user moves the mouse close to the edge of the view. In this case, it's useful to know the signed distance for detecting when the cursor is close, as well as how much to move the camera. You want signed distance here, not just unsigned distance to edge, because it allows dragging beyond the edge to work seamlessly, as well as the ability to offset the rectangle in either direction (distance wise).

notlion:
Also I agree that `distance > 0` is a better indicator that the point lies inside.

richardeakin:
I suppose the current use case for distance returning 0 for inside is when you are using it as a scalar, in which case 0 would mean no scaling. This would break any apps that relied on that, whereas now they would be scaling by a negative number. Again, just discussing the pros and cons. :)

paulhoux:
I like the implementation of ```distanceSigned()```, but please not that it does not return the shortest distance for points that lie outside the rectangle. It only yields the proper results for points inside. 

Consider a rectangle with coords topleft = (10, 10) and bottomright = (30, 20). The distance to the origin as returned by the ```distance()``` function is ```d = sqrt( 10 * 10 + 10 * 10 ) =  14.14```, which is correct, but the distance returned by ```distanceSigned()``` is 10.

PR #1025, opened 2015-07-11
Add D3D11 feature level 9.3 support to ANGLE renderer

pizthewiz:
The existing display attributes require feature level 10.0+ and following [template code from the MSOpenTech/angle repo](https://github.com/MSOpenTech/angle/blob/ff68ed35a5fe07283e804849d56ae3993756faa4/templates/10/Windows/Windows%20Universal/XamlUniversal/OpenGLES.cpp#L107-131), we can test for and support 9.3. A few proximate space <-> tabs were also fixed too. With this, I was able to build and run Cinder apps on a Windows 8.1 guest running in VMWare Fusion.


andrewfb:
I don't have a setup that I can directly replicate this on, but I forced the code to execute to test it (having defined CINDER_GL_ES_2 as well) and it looks good. Thanks for this.

PR #1026, opened 2015-07-12
Add TinyEXR source files to iOS compile step

notlion:
iOS project builds are broken due to these files not being in the iOS compile targets. There's also an issue with boost::filesystem not linking in the x86_64 simulator. I'm looking into that.


notlion:
The boost::filesystem thing turned out to be a false alarm. It was an error in my project build settings.

andrewfb:
Thanks for this fix...

notlion:
No problem! Also can I just say that deadfab is probably the coolest ref hash that i've gotten so far :sunglasses:

PR #1027, opened 2015-07-13
Shouldn't Fbo::getTextureBase accept depth attachments as well?

simongeilfus:


simongeilfus:
Should we have two warning messages in case the constant is illegal AND unknown? 

```c++
if( ( (attachment < GL_COLOR_ATTACHMENT0) || (attachment > MAX_COLOR_ATTACHMENT) ) && ( (attachment != GL_DEPTH_ATTACHMENT) &&  (attachment != GL_DEPTH_STENCIL_ATTACHMENT) ) ) {
                auto constant = gl::constantToString( attachment );
                if( !constant.empty() )
		     CI_LOG_W( "Illegal constant for texture attachment: " << constant );
                else
		     CI_LOG_W( "Unknown constant for texture attachment: " << attachment );
	}
```

andrewfb:
Good catch - thanks for this.

simongeilfus:
Great, just one little detail. In my last commit [I made the warning message more verbose](https://github.com/simongeilfus/Cinder/commit/a0e390bafdbf16afcd8e7a3234f520b3c8522cb8#diff-2bf74e05126920694e11c8987f5eb641R506) by using gl::constantToString. Which means that currently if an unknown constant is used the warning will be wrong. 

I think those are the options :
* Leave the way it was before. Message would look like "Illegal constant For texture attachment: 30056". A bit annoying...
* Leave it the way it is now and have sometimes warning like "Illegal constant for texture attachment: ". Not really useful.
* Have a second if and a more verbose message.

simongeilfus:
Just sent a PR for the last option.

PR #1029, opened 2015-07-14
Made Fbo::getTextureBase more verbose (addition to previous commit)

simongeilfus:
See https://github.com/cinder/Cinder/pull/1027

### DIFF
--- a/src/cinder/gl/Fbo.cpp
+++ b/src/cinder/gl/Fbo.cpp
@@ -502,7 +502,13 @@ Texture2dRef Fbo::getTexture2d( GLenum attachment )
 
 TextureBaseRef Fbo::getTextureBase( GLenum attachment )
 {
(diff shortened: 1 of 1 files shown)


andrewfb:
Hey - do you have a specific use case where the constant is unknown? I suppose we could modify the constantToString() method to loop over the max color attachments or something technically, but otherwise it seems like we have all the cases covered?

richardeakin:
I really like the extra verbosity in these error messages, was meaning to suggest using `constantToString()` here. Though I agree, we should try to cover all cases we can conceive of to give more clues to the user how to fix their Format.

simongeilfus:
No specific case in mind, except when you would call getTextureBase with something like  GL_COLOR_ATTACHMENT_0 + i that might give something over the max color attachment. That was the idea behind having two messages, "illegal" would be for an existing constant but not properly used, and "unknown" would be for un-existing constant.

andrewfb:
I see the logic but personally I think this may be a bit overengineered given how rare this case will be in practice. Open to dissenting opinions though...

PR #1034, opened 2015-07-18
Add PolyLineT scaled() and offsetted().

drewish:
Following the pattern in RectT adding some methods to return copies rather than modifying in place.

### DIFF
--- a/include/cinder/PolyLine.h
+++ b/include/cinder/PolyLine.h
@@ -54,10 +54,11 @@ class PolyLineT {
 	T			getPosition( float t ) const;
 	T			getDerivative( float t ) const;
 
-	void		scale( const T &scaleFactor, T scaleCenter = T() );
-	void		offset( const T &offsetBy );
-	void		reverse();
-	
+	void			scale( const T &scaleFactor, T scaleCenter = T() );
+	PolyLineT<T>	scaled( const T &scaleFactor, T scaleCenter = T() ) const;
+	void			offset( const T &offsetBy );
+	PolyLineT<T>	getOffset( const T &offsetBy ) const;
+	void			reverse();
 	PolyLineT<T>	reversed() const;
 
 	//! Returns whether the point \a pt is contained within the boundaries of the PolyLine
--- a/src/cinder/PolyLine.cpp
+++ b/src/cinder/PolyLine.cpp
@@ -64,10 +64,26 @@ void PolyLineT<T>::scale( const T &scaleFactor, T scaleCenter )
 }
 
 template<typename T>
+PolyLineT<T> PolyLineT<T>::scaled( const T &scaleFactor, T scaleCenter ) const
+{
+	PolyLineT<T> result( *this );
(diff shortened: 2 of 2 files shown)


richardeakin:
Seems nice to me to have the congruency. Are we all cool with the name `offsetted()`. I personally can't think of anything better that follows the existing patterns.

gaborpapp:
I think RectT has `getOffset()` instead of `offseted()`. So one of them should change for consistency.

drewish:
I tried to follow the pattern as best I could but I'm not really a fan of that name. I'm open to what ever the consensus is. 

I'm going to rebase it to pick up the reverse addition.

include/cinder/PolyLine.h
```diff
@@ -54,10 +54,11 @@ class PolyLineT {
...
 
-	void		scale( const T &scaleFactor, T scaleCenter = T() );
-	void		offset( const T &offsetBy );
-	void		reverse();
-	
+	void			scale( const T &scaleFactor, T scaleCenter = T() );
+	PolyLineT<T>	scaled( const T &scaleFactor, T scaleCenter = T() ) const;
+	void			offset( const T &offsetBy );
+	PolyLineT<T>	offsetted( const T &offsetBy ) const;
+	void			reverse();
```
drewish: I guess GitHub uses 8 spaces per tab? I'm using 4 per the [Contributing guidelines](https://github.com/cinder/Cinder/blob/master/CONTRIBUTING.md).

drewish:
Oh speaking of the Rect class... I'd love to follow it's pattern and add transform() and transformed() methods.

drewish:
Ah so looks like transform/transformed will both run into the same precision issues were touched on in #1007. [The suggestion there](https://github.com/cinder/Cinder/pull/1007#issuecomment-118723504) was a freestanding function and I think that might work well for this. It would avoid a lot of issues with trying to get the right precision and dimensions (I believe you'd want to turn vec2 into vec3 and vec3 into vec4 so you could use homogeneous coordinates).

drewish:
Any interest in this?

richardeakin:
I think there is, and thanks for contributing this work, however we're sort of in a focus on getting through changes necessary for 0.9.0 release. These are pure additions, so perhaps we should wait to sift through the api choices until 0.9.1-dev begins.

drewish:
@richardeakin seems like stuff has settled down, mind giving it another look now?

richardeakin:
We'll try to get these looked at very soon, along with your other PolyLine improvements. Personally I'm wrapping up a 3 month project that has been consuming my time for the last few weeks. Thanks again for these contributions..

andrewfb:
Just to be frank, the naming between `Rect`, `Area`, `PolyLine` and `Shape2d` / `Path2d` are all in serious need of cleanup and unification, particularly in light of our move to GLM. I'm up for merging this, acknowledging that we're due for a serious, breaking refactor in the near future.

PR #1037, opened 2015-07-20
fix link to Mac OS X project setup in Chapter 1 of Hello, Cinder 

tracyscott:
libcinder.org should also be updated.

http://libcinder.org/docs/v0.8.6/hello_cinder_chapter1.html


andrewfb:
Thanks for this - we're reworking the site and docs but this should get fixed in the nightly docs gen in the current system.

PR #1038, opened 2015-07-21
Added Debug log level

lucasvickers:
I added an additional logging level and macro, ```LEVEL_DEBUG``` and ```CI_LOG_D```.

I also re-generated the DebugTest app so that I could run it on 10.10.4 and eventually use it to test WinRT.  I tested on OS X and Window 8, hopefully it doesn't cause trouble for anyone.

[This commit](https://github.com/cinder/Cinder/commit/11f01050fc77911156f09e67d74d20034b131b69) is the core change.  @richardeakin I need your feedback on what I did for the enum values.  Check the message for the above mentioned commit.  I'm open to solving it via changing the CI_MAX_LOG_LEVEL values instead if we think it's more logical.


lucasvickers:
@richardeakin please let me know your feedback.  

I agree that it makes more sense if LEVEL_ERROR > LEVEL_DEBUG, but I think the old CI_MAX_LOG_LEVEL was confusing.  Before I found it inconsistent that LEVEL_VERBOSE was 0, but you needed to set CI_MAX_LOG_LEVLEL 6 to enable verbose logging.

This change breaks old functionality, but I think it's easier to understand that
```#define CI_MIN_LOG_LEVEL 2```
Sets the minimum logging level to LEVEL_INFO (2)

richardeakin:
I think switching to `CI_MIN_LOG_LEVEL` instead of max makes sense, and now would be the time to do it if we're going to. What about leaving `LEVEL_VERSBOSE` at 0, the rest are increasing in level, and then by default we set the `CI_MIN_LOG_LEVEL` to 2 (`LEVEL_INFO`) for release mode ([this block](https://github.com/cinder/Cinder/blob/master/include/cinder/Log.h#L289-L295))?

lucasvickers:
[This](https://github.com/lucasvickers/Cinder/commit/7a03670605d3163a235eb5f78444bf8fcabbbe51) commit set the ```LEVEL_VERBOSE``` back to 0 and default to ```LEVEL_INFO`` for release.

The ```LEVEL_VERBOSE = 1``` was me fighting the compiler in the wrong way.

Sorry about the messy PR

richardeakin:
Okay all looks great from here. On the contrary, I like that this is a more consistent interpretation of the level values.

PR #1042, opened 2015-07-23
Ading back .icns and .ico files to DebugTest

lucasvickers:
I got too aggressive when removing binaries, saw these are referenced by the xcode/vcc projects.


lucasvickers:
Unneeded PR, will remove the .ico from the VS project instead.

PR #1048, opened 2015-07-30
Support for App Transport Security on OS X and iOS

pizthewiz:
Introduced in OS X 10.11 and iOS 9, [App Transport Security](https://developer.apple.com/library/prerelease/ios/technotes/App-Transport-Security-Technote/) requires a secure connection from app to a web service to prevent accidental disclosure and just generally be secure by default. A few samples load content via web service and exceptions are required in some cases to be runnable on 10.11+ / iOS 9+. In each case, the most narrow exception only to the immediate web service domain.

### Samples
Samples were found via:
```sh
$ grep -r -i --include \*.cpp '"http' samples | less
```

#### FlickrTestMultithreaded / flickrTest
The API request is (now) made over HTTPS, but unfortunately it doesn't support [forward secrecy](https://en.wikipedia.org/wiki/Forward_secrecy) so an exception is still required. Apparently all is well with the server configuration for the image loads on `https://farm*. staticflickr.com`.

#### Earthquake
The API request is made over HTTP.

#### QuickTimeAdvanced
The videos are loaded over HTTP.

### Later Pass
There are a couple samples that aren't working currently that will need similar treatment in a second pass:
* _svg/GoodNightMorning
* Kaleidoscope

### DIFF
--- a/samples/Earthquake/xcode/Info.plist
+++ b/samples/Earthquake/xcode/Info.plist
@@ -20,6 +20,17 @@
 	<string>????</string>
 	<key>CFBundleVersion</key>
 	<string>1.0</string>
+	<key>NSAppTransportSecurity</key>
+	<dict>
+		<key>NSExceptionDomains</key>
+		<dict>
+			<key>earthquake.usgs.gov</key>
+			<dict>
+				<key>NSThirdPartyExceptionAllowsInsecureHTTPLoads</key>
+				<true/>
+			</dict>
+		</dict>
+	</dict>
 	<key>NSMainNibFile</key>
 	<string>MainMenu</string>
 	<key>NSPrincipalClass</key>
--- a/samples/FlickrTestMultithreaded/src/FlickrTestMultithreadedApp.cpp
+++ b/samples/FlickrTestMultithreaded/src/FlickrTestMultithreadedApp.cpp
@@ -58,7 +58,7 @@ void FlickrTestMTApp::loadImagesThreadFn( gl::ContextRef context )
 	vector<Url>	urls;
 
 	// parse the image URLS from the XML feed and push them into 'urls'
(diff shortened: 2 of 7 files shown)


pizthewiz:
The Kaleidoscope sample now includes an exception request as well though it doesn't compile as-is, some matrix stuff needs to be updated.

richardeakin:
FYI the kaleidescope sample is due to be updated in both technique and to use 0.9.0 in the coming days.

About responding to the purpose of this PR, I haven't as I really don't know anything about these new security features and haven't had the time to learn about them.  But to instigate comment, what do we gain from a sample's perspective in adding these extra security settings? Will it be required for the sample to function, come OS X 10.11 and iOS 9? Or is it rather added security and 'the right thing to do'?

pizthewiz:
Quite simply, OS X 10.11 and iOS 9 will not load HTTP/HTTPS requests without the explicit exceptions. From [App Transport Security Technote](https://developer.apple.com/library/prerelease/ios/technotes/App-Transport-Security-Technote/):
> App Transport Security is a feature that improves the security of connections between an app and web services. The feature consists of default connection requirements that conform to best practices for secure connections. Apps can override this default behavior and turn off transport security.
>
> ### Default Behavior
> All connections using the NSURLConnection, CFURL, or NSURLSession APIs use App Transport Security default behavior in apps built for iOS 9.0 or later, and OS X 10.11 or later. Connections that do not follow the requirements will fail.

I have to imagine existing apps are not bound by these rules and it only kicks in if one builds with OS X 10.11 or iOS 9 as a deployment SDK (or some similar mechanism) - something worth a mention in the OS X and iOS guide docs for sure.

(I'm going to delete the `nscurl` comment above, it seems to dominate the space without being super helpful).

pizthewiz:
I just realized the `Info.plist` changes will need to be copied to the iOS projects as well…

andrewfb:
Thanks for tracking this down. I don't have a OS 9 or 10.11 machine to test these myself. And it does appear we'll need to match the work for iOS 9.

PR #1050, opened 2015-07-31
Geom source mods combination

andrewfb:
Adds new geom:: combination syntax, makes SourceMods-user facing, adds geom::Constant.


richardeakin:
Aside from these obviously being quite powerful additions, I wanted to post a couple comments.

Provided the following (modified from your example on the [forum post](https://forum.libcinder.org/topic/rfc-geom-update)):
```cpp
TriMesh meshA( ... );
TriMesh meshB( ... );
return gl::Batch::create( meshA & meshB, ... );
```

if I want to optimize that, would I write the following?:

```cpp
return gl::Batch::create( &meshA & &meshB, ... );
```

Been trying to think through if there could be any ambiguity between & as a reference and the overloaded one.

Other thing is, I don't quite understand the description of the optimized path using pointer to geom::Source ("meaning, don't take the address of a temporary, or return a geom::SourceMods that refers to a local variable by pointe"). For example, what copying is avoided in your sample, as opposed to leaving off the &?:

```cpp
TriMesh mesh( ... );
return gl::Batch::create( mesh >> geom::Translate(), ... );
```

Wondering how the `&mesh` avoids the copy.

Lastly, what is the expected result of the following?

```cpp
auto result = meshA & meshB >> geom::Translate();
```

andrewfb:
That example is correct (for combining two meshes by pointer):
`return gl::Batch::create( &meshA & &meshB, ... );`
There's a similar [one in the test].(https://github.com/cinder/Cinder/blob/geomSourceMods/test/GeomSourceMods/src/GeomSourceModsApp.cpp#L139)

The same rules that distinguish unary& (address-of) and bitwise& apply here, so we should be fine in term of disambiguation. To clarify, the `&` is not creating a reference in the C++ sense of that word - it's taking the address (making a `Source*`).

The copying that is avoided by making a pointer to the `Source*` is the copying of the `TriMesh` and its data. If we construct the `SourceMods` from pointers (`&meshA` rather than simply `meshA`) we're promising that scope of `meshA` will outlast the `SourceMods`. So the rules are identical to when you would pass the address of something back from a function, for example. Value of a local variable is fine, pointer (or reference) is not.

In this new design, just about any expression will lead to something safe, and predictable according to normal C++ operator precedence. However if you have an advanced case (pretty much limited to adding `Modifier`s to a `TriMesh`) then the address-of can be an opt-in optimization when you can ensure your `TriMesh` and your `SourceMods` share a scope.

The phrase
`auto result = meshA & meshB >> geom::Translate();`
abides by the same [operator precedence](http://en.cppreference.com/w/cpp/language/operator_precedence) rules as anything else in C++, which is to say, `meshA` is combined with a translated `meshB`. It's equivalent to:
`auto result = meshA & (meshB >> geom::Translate());`

PR #1056, opened 2015-08-06
Changed parameter for Fbo::blitTo from Fbo to FboRef.

paulhoux:
I chose not to use a reference to a shared_ptr, as we want to make sure to Fbo remains valid while blitting.


richardeakin:
Good catch, I'm surprised that that was allowed to compile, I would expect `Fbo` (and other gl resource types) to be noncopyable.

Other thing, we've been moving towards passing in all `shared_ptr` objects as const& (note: not shared_ptr<const T>() types) to avoid the extra `shared_ptr` overhead, IMO this one should follow that pattern as well.

PR #1058, opened 2015-08-09
Added ci::app::launch as a way to launch a Cinder app without macros

sepehr-laal:
The standard way of running an app in Cinder has been traditionally calling a macro to define the entry point. I specifically want to delay the launch of the application in my `mian()`. With the existence of the `ci::app::launch()` A user will have the freedom to choose between calling a macro or launching manually and making sure that both calls will result in the same behavior.

### DIFF
--- a/include/cinder/app/App.h
+++ b/include/cinder/app/App.h
@@ -25,6 +25,11 @@
 
 #include "cinder/Cinder.h"
 
+namespace cinder { namespace app {
+template< typename AppType, typename RendererType, typename ...Args >
+void launch( const char* title, Args... args );
+} } // namespace cinder::app
+
 #if defined( CINDER_MAC )
 	#include "cinder/app/cocoa/AppMac.h"
 	namespace cinder { namespace app {
--- a/include/cinder/app/cocoa/AppCocoaTouch.h
+++ b/include/cinder/app/cocoa/AppCocoaTouch.h
@@ -264,11 +264,16 @@ void AppCocoaTouch::main( const RendererRef &defaultRenderer, const char *title,
 	AppBase::cleanupLaunch();
 }
 
+template< typename AppType, typename RendererType, typename ...Args >
+void launch( const char* title, Args... args ) {
+	cinder::app::RendererRef renderer( new RendererType );
+	cinder::app::AppCocoaTouch::main<AppType>( renderer, title, args... );
+}
+
 #define CINDER_APP_COCOA_TOUCH( APP, RENDERER, ... )									\
 int main( int argc, char * const argv[] )												\
 {																						\
-	cinder::app::RendererRef renderer( new RENDERER );									\
-	cinder::app::AppCocoaTouch::main<APP>( renderer, #APP, argc, argv, ##__VA_ARGS__ );	\
+	cinder::app::launch< APP, RENDERER >( #APP, argc, argv, ##__VA_ARGS__ );			\
 	return 0;																			\
 }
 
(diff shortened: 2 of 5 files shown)


richardeakin:
Thanks for taking the time to throw this together. Though I wonder, would something like the following get you what you need?

```cpp
int main( int argc, char* argv[] )
{
	app::RendererRef renderer( new app::RendererGl() );
	app::App::main<MyAppSubclass>( renderer, "My App Name", argc, argv, prepareFn );
	return 0;
}
```

The platform specific AppBase subclasses are typedef'ed to `App`, so you can just use `App` and the right one will be chosen based on preprocessor defines.

sepehr-laal:
@richardeakin That is actually how I have it set up. The problem is that I never needed to compile my app on other platforms (rather than Windows) and today I just noticed the signature of `ci::app::App::main` is different on Mac vs. Windows vs WinRT.

I thought if we had something like `ci::app::App::main` which stays with the same signature on all platforms, it'd be great. The main reason I personally need this to be a function and not a macro is that I parse settings entries from an XML file and feed them into the `prepareFn` and also I start my app in a NodeJS module (DLL).

richardeakin:
Ah yea, they are different, forgot about that. As I recall, it was because the command line args are parsed differently on windows so they were wasted arguments. But thinking about it, once you have to go from your own main function then there are bound to be little differences from platform to platform, not unlike how argument variables and the app name is retrieved.

Other thing is, this part of the code is already a tad on the confusing side, as it is using variadic templates, variadic macros, optional arguments, etc. Not saying I'm against adding an additional function call at startup, but that it should be weighed with respect to the added complexity.  What I think would be the best of both worlds would be to improve on the existing `AppType::main<T>` functions so they are easier to use directly for purposes such as what you're doing, since you should be able to write your own main functions when the need arises.

Now, there is also a different sort of problem altogether - we can't pass the Renderer as a template parameter as it is optionally taking a `Renderer::Options` (`RendererGl::Options` more typically), this is why it is passed in as a `RendererRef` argument to the templated main function.

Curious about the XML - are you meaning that you can't regularly parse it from the prepareFn (when using the `CINDER_APP` macro), or that you're doing a custom main<AppT> based on what your settings are?

sepehr-laal:
Regarding improving `ci::app::App::main` and not adding additional API, I am in favor of that as well, I just did not want to break backward compatibility. If you have something specific in mind, I would be more than happy to implement it and submit a separate PR.

I have a little game engine implemented on top of Cinder that uses a bunch of XML files to configure itself on startup. Before when it used to be `prepareSettings` as a member of `ci::app:App`, it was super convenient and reliable since I could just read XMLs in app constructor and have all configs ready before `prepareSettings` is hit. Now I have to access Engine's singleton in `prepareFn` which is a bit ugly! But I am pretty sure there was a good reason `prepareSettings` was removed.

Back to the `App::main` topic, I think if `RendererRef` is passed before `const char*` argument, `ci::app::launch` will turn into a superset for all `ci::app::App::main`s. Like this:

```C++
namespace cinder { namespace app {
template< typename AppType, typename ...Args >
void launch( RendererRef, const char* title, Args... args );
} } // namespace cinder::app
```

paulhoux:
@sepehr-laal : sorry I didn't come across this PR earlier, just wanted to say that ```prepareSettings``` was not removed entirely, it just became a bit different:
```
class MyApp : public App {
public:
    static void prepare( Settings *settings ); 
};

CINDER_APP( MyApp, RendererGl, MyApp::prepare )
```
Or simply define a free function:
```
void prepareSettings( App::Settings* settings ) {}

class MyApp : public App {
public:
};

CINDER_APP( MyApp, RendererGl, prepareSettings )
```

richardeakin:
Going to close this as there is currently a fairly simple way to achieve the same thing. Thanks for the discussion and contribution nonetheless!

PR #1061, opened 2015-08-16
Added Unsigned Int Sampler Buffer

ryanbartley:


ryanbartley:
This solves the problem brought up [here](https://forum.libcinder.org/#Topic/23286000002435011).

PR #1069, opened 2015-08-19
Adding Logging sample app

lucasvickers:
Added a simple logging sample app.  Some colleagues I had spoken with didn't instinctively know where to look for logging examples, and I think it makes sense to have a straight forward example available for users.

### DIFF
--- a/include/cinder/Log.h
+++ b/include/cinder/Log.h
@@ -189,10 +189,10 @@ public:
 	bool isConsoleLoggingEnabled() const				{ return mConsoleLoggingEnabled; }
 
 	void enableFileLogging( const fs::path &filePath = fs::path(), bool appendToExisting = true );
-	void enableRotatingFileLogging( const fs::path &folder, const std::string& formatStr, bool appendToExisting = true);
+	void enableFileLoggingRotating( const fs::path &folder, const std::string& formatStr, bool appendToExisting = true);
 	void disableFileLogging();
 	void setFileLoggingEnabled( bool enable, const fs::path &filePath = fs::path(), bool appendToExisting = true );
-	void setRotatingFileLoggingEnabled( bool enable, const fs::path &folder, const std::string &formatStr, bool appendToExisting = true );
+	void setFileLoggingRotatingEnabled( bool enable, const fs::path &folder, const std::string &formatStr, bool appendToExisting = true );
 	bool isFileLoggingEnabled() const					{ return mFileLoggingEnabled; }
 
 	void enableSystemLogging();
--- a/include/cinder/Log.h
+++ b/include/cinder/Log.h
@@ -189,10 +189,10 @@ public:
 	bool isConsoleLoggingEnabled() const				{ return mConsoleLoggingEnabled; }
 
 	void enableFileLogging( const fs::path &filePath = fs::path(), bool appendToExisting = true );
-	void enableFileLogging( const fs::path &folder, const std::string& formatStr, bool appendToExisting = true);
+	void enableRotatingFileLogging( const fs::path &folder, const std::string& formatStr, bool appendToExisting = true);
 	void disableFileLogging();
 	void setFileLoggingEnabled( bool enable, const fs::path &filePath = fs::path(), bool appendToExisting = true );
-	void setFileLoggingEnabled( bool enable, const fs::path &folder, const std::string &formatStr, bool appendToExisting = true );
+	void setRotatingFileLoggingEnabled( bool enable, const fs::path &folder, const std::string &formatStr, bool appendToExisting = true );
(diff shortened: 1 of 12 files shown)


num3ric:
Yes, very useful/easy to access reference!

See I didn't know that file logging would roll over after midnight. Actually just to confirm, it means the file will be cleared? Obviously the fear is having a log file that ends consuming gigs of disk space...

lucasvickers:
I see how it was a little confusing, thanks for pointing it out!  I updated the files that should help your understanding.  I'm also writing out the details below, so let me know if you think it's all clear in the code/examples now.

If you set up a [standard log file](https://github.com/cinder/Cinder/blob/master/include/cinder/Log.h#L102) then it won't roll over the file.  If you disable appending it will just overwrite it every time you start.  

If you set up a [rotating logger](https://github.com/cinder/Cinder/blob/master/include/cinder/Log.h#L104), then it will roll over at midnight.  If you are rotating logs, Cinder is just going to roll the logs over, not delete the older one.  So yeah, you could in theory use up gigs of disk space.

Specific to lots of logs, on a recent install I wrote a [simple python log trimmer](https://github.com/lucasvickers/python-log-trimmer) that handles the logic of trimming logs for us.  I've just beefed up the README.  I'm not sure it makes sense to put more advanced log management into cinder, but I'd be open to some type of global documentation for install best practices.  @andrewfb @richardeakin maybe there's somewhere in the new documentation that talks about this kinda stuff?

pizthewiz:
This sample is absolutely invaluable!

One area of logging that still trips me up is setting the log level for a file logger at runtime. `CI_MIN_LOG_LEVEL` assigns the logging macros at compile-time, so I'm not sure how it would work with the system logger which has its own log level as well.

lucasvickers:
Added an explanation to the sample.  Please take a look.  Thanks!

pizthewiz:
That definitely explains the interplay between `CI_MIN_LOG_LEVEL` and `setSystemLoggingLevel()`, thanks for the additional clarification.

richardeakin:
Sample looks great, it will definitally be valuable in introducing people to our new logging tools. A couple suggestions:

1) Would someone be interested in writing a guide for the [new docs](https://forum.libcinder.org/#Topic/23286000002431027). Ideally, we would have a welcoming and easy to read doc page that is accompanied with a concise sample to show the logging in practice.

2) We don't yet really have any policy around the 'voice' we should be using in docs, which we're still exploring, but I think it would be clearer if we try to reference the components that are responsible for some functionality instead of accrediting 'cinder' directly (or using third person if appropriate). What I mean is the following line (from [here](https://github.com/cinder/Cinder/pull/1069/files#diff-d45900795d2cdd3fefc55f0c2d98c245R72)):

```cpp
// Cinder will create the folder structure for you, and defaults to appending to
// the logging file.
log::manager()->enableFileLogging( "/tmp/logging/cinder.log" );
```

could be written as:

```cpp
// Enable file logging to the specified file path, the folder structure will be constructed if necessary.
// By default, logs will be appends to the same file between application executions.
log::manager()->enableFileLogging( "/tmp/logging/cinder.log" );
```

Per your question, what does the word 'install' mean in this case - are we talking installing an app, or installation works?  I don't believe thate is a place for documenting either, but we'd be open to adding one. There's lots of inside info for both topics, would be great to get them written down.

pizthewiz:
A guide would be excellent!

lucasvickers:
I'll take a stab at rewriting the comments as well as a logging guide for the new docs system.  You are 100% right my referring to cinder itself is rather confusing.

Regarding the meaning of an installation, I was referring to my installation of an app on a dedicated machine for a permanent installation.  The general topic of installations is pretty broad, and it covers things like best practices for automating the installation of apps on machines, as well as best practices for setting up the application/computer installations for long term reliability.  The long term reliability stuff doesn't pertain directly to cinder, and it may not even make sense to be hosted on a cinder related page, but I'm sure it would be considered valuable information for our professional users.

lucasvickers:
Cleaned up the comments, and also renamed one of the logging methods to improve readability and fix a bug related to overloaded function evaluation.

I'll move the more detailed logging information into a guide for the new docs system.

richardeakin:
Cool, getting ready to merge, although I would propose we use `enableFileLoggingRotated()` instead of `enableRotatingFileLogging()` to take advantage of editor's auto complete and docs with alphabetical ordering. We normally try to append suffixes like this when there are related methods so they are easier to find.

lucasvickers:
That makes sense to me, thanks for the tip.  I'm opting for ```enableFileLoggingRotating``` over ```enableFileLoggingRotated``` because I think the ```ing``` tense makes more sense given it's an ongoing operation.

richardeakin:
Ok yea, that one works too.

PR #1077, opened 2015-08-21
Added sample for camera persp

keithters:
Added a new sample that demonstrates how the CameraPersp works in Cinder.


richardeakin:
Ooh, very nice.  A couple notes on image assets:

* To reduce binary weight and make them easier to update, we're trying to reuse the icon images from [samples/data](https://github.com/cinder/Cinder/tree/master/samples/data), rather than adding a new copy.
* That is a purdy environment map, although it is 8.5mb. Should we either compress it, or find a smaller one?  Also, we talked about placing assets like that in the samples/data folder so multiple samples can use them, although we haven't done a great job of that for existing samples.

Issue #1035, opened 2015-07-19
cocoa::getWindowContext() returns null on OS X 10.11 El Capitan


pizthewiz:
I waited a while to report this thinking it was an issue with the first developer seed, but it has continued into the second and third seeds as well. Strangely, when running the _Renderer2dBasic_, the app doesn't draw anything and after a bit of investigation, the fetched `CGContextRef` is `null`:
```C++
void Renderer2dApp::draw()
{
	// Render using CoreGraphics on the mac; origin is lower left
#if defined( CINDER_COCOA )
	CGContextRef context = cocoa::getWindowContext();
```

The code that fetches the context is in [src/cinder/cocoa/CinderCocoa.mm](https://github.com/cinder/Cinder/blob/master/src/cinder/cocoa/CinderCocoa.mm#L143-150):
```Objective-C
CGContextRef getWindowContext()
{
#if defined( CINDER_MAC )
	return (CGContextRef)[[NSGraphicsContext currentContext] graphicsPort];
#else
	return ::UIGraphicsGetCurrentContext();
#endif
}
```

The [NSGraphicsContext docs](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSGraphicsContext_Class/index.html#//apple_ref/occ/instp/NSGraphicsContext/graphicsPort) suggests that the `graphicsPort` property has been deprecated in OS X 10.10 (though it doesn't generate a warning) and one should instead use the `CGContext` property. Reworking `getWindowContext()` to dymaically use the new property when available is perhaps more correct but still doesn't return a valid `CGContextRef`:
```Objective-C
CGContextRef getWindowContext()
{
#if defined( CINDER_MAC )
	if( [[NSGraphicsContext currentContext] respondsToSelector:@selector(CGContext)] ) {
		return (CGContextRef)[[NSGraphicsContext currentContext] performSelector:@selector(CGContext) withObject:nil];
	} else {
		return (CGContextRef)[[NSGraphicsContext currentContext] graphicsPort];
	}
#else
	return ::UIGraphicsGetCurrentContext();
#endif
}
```

[...]

PR #1078, opened 2015-08-22
Avoid direct/uninitialized display access on OS X / iOS

### LINKED ISSUE #1074: Cinder 0.9: AppCocoaViewTest project crashes
The AppCocoaViewTest project crashes for me on 
```console () << "Display: " << *(win-> getDisplay ()) << std :: endl ; ```
in 
```MyCinderApp ::setup()```.

it looks like the getDisplay() method returns null.

pizthewiz:
Fixes #1074

It seemed as though the display finding functions directly accesses the `mDisplays` member variable which hasn't been lazy loaded when using `CinderViewMac`. Worth noting that the app runs into a context error (#1035) on OS X 10.11 unless the `#define USE_RENDERER2D` is removed from `MyCinderApp.h`


andrewfb:
Nice find & fix - thanks...

PR #1079, opened 2015-08-22
Kaleido sample update

num3ric:
So, Rich, well... I won't be spending the time to reproduce these exact visuals in Glsl I'm afraid. These updates suffice.

If anyone has some free time to lose, here it is: https://www.shadertoy.com/view/ltl3WN


richardeakin:
Nice - thanks for updating - only one more to go (ahem @flight404 :))!

Issue #1071, opened 2015-08-19
Clarifying terminology in gl:: depth (or, 'depth read' versus 'depth test')


richardeakin:
I recently found the following (from [here](https://www.opengl.org/wiki/GLAPI/glDepthFunc#Notes), and in more general [here](https://www.opengl.org/wiki/Depth_Test)) info when trying to do some more complex depth + blending:

> Even if the depth buffer exists and the depth mask is non-zero, the depth buffer is not updated if the depth test is disabled. In order to unconditionally write to the depth buffer, the depth test should be enabled and set to GL_ALWAYS (see glDepthFunc​).

So, I take that to mean is that the word 'test' is actually a better fit rather than 'read', since you can't write unless you `glEnable( GL_DEPTH_TEST )` as well. We made the name changes to use 'read' and 'write' instead of 'test' and 'mask' because they are easier to remember, but this was somewhat of an oversite.  The effected API's that I know of are:

* The arguments to `ScopedDepth` ([here](https://github.com/cinder/Cinder/blob/master/include/cinder/gl/scoped.h#L244-L248)).
* `gl::enableDepthRead()` (as opposed to the prospective `gl::enableDepthTest()`).

I'll also note that `ScopedDepth( false, true )`, meaning disable read but unable write, doesn't currently work - nothing gets written to the z buffer as `GL_DEPTH_TEST` is off.

On a related note, is whether we should rename depth 'write' calls to use the word 'mask' instead, to match GL. I think this one is more questionable, as [opengl docs](https://www.opengl.org/wiki/GLAPI/glDepthMask) themselves almost use the words interchangebly (" enable or disable writing into the depth buffer"), but we're soliting the opinion of others since we have the opportunity to make the name change now if that is decision.

richardeakin:
We talked about it here and came up with the following solution for how `gl::ScopedDepth*` should work:

```cpp
gl::ScopedDepth depthScope( true ); // enable test and write, doesn't touch depth func
gl::ScopedDepth depthScope( false ); // disables test and write, doesn't touch depth func
gl::ScopedDepth depthScope( true, GL_LESS ); // enable test and write with scoped depth function as well

gl::ScopedDepthTest depthTestScope( true ); // enables test, doesn't touch write or depth func
gl::ScopedDepthTest depthTestScope( true, GL_LESS ); // enables test and sets depth func, doesn't touch write

gl::ScopedDepthWrite depthWriteScope( true ); // enables write, doesn't touch test or depth func
```

[...]

PR #1094, opened 2015-08-30
Use 'test' terminology instead of 'read' for gl::enableDepthTest + Scoped (and enableStencilTest)

### LINKED ISSUE #792: enableStencilWrite/disableStencilWrite undefined
What's the plan for filling in these methods?

richardeakin:
Explained in #1071.  Other than making the api modifications in that issue, I also renamed `gl::enableStencilRead()` / `gl::disableStencilRead` in the header to `gl::enableStencilTest()`, etc, which was already going by that name in the .cpp (but inaccesible). We still need to fix #792 by either removing or implementing `enableStencilWrite`, I'll wait until feedback to do something about that.

Updated all samples and tests. We should be aware that existing code of the form:

```cpp
gl::ScopedDepth depth( true, false ); // or:
gl::ScopedDepth depth( true, true );
```

will be broken after this - at least on clang I'm seeing that the compiler accepts the second bool as the GLfunc for depth function. When I tested with the Earthquake sample, this lead to a `GL_INVALID_OPERATION` at least, also ScopedDepth is a new method anyway so I think it is ok for us to make breaking changes now for the better.

### DIFF
--- a/include/cinder/gl/scoped.h
+++ b/include/cinder/gl/scoped.h
@@ -238,20 +238,44 @@ private:
 
 #endif
 
-//! Scopes state of depth testing and writing
+//! Scopes state to control both depth testing and writing together. See information on `GL_DEPTH_TEST` and `glDepthMask()`.
 struct ScopedDepth : private Noncopyable {
-	//! Enables or disables both depth comparisons and writing to the depth buffer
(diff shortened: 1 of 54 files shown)


richardeakin:
Superseded by #1123.

PR #1095, opened 2015-08-31
Plug a couple of OS X / iOS leaks

pizthewiz:
I was trying to check out the clang address sanitizer in Xcode 7 and instead ended up running the static analysis tool which flagged these. It doesn't appear as though these are called through the update cycle or anything so I don't think they'd represent significant leaks or fixes - regardless they'd probably benefit from some quick validation.


andrewfb:
Thanks for tracking these down...

pizthewiz:
No problem, always happy to help out.

PR #1102, opened 2015-09-09
ObjLoader sample has some alpha issues

simongeilfus:
There's a small issue with this sample shaders.
Right before outputing the color, the whole RGBA is multiplied by the variable ```diffuse``` making the mesh transparent in some places. This creates some transparency issues as shown here:

screenshot 2015-09-09 10 10 48
screenshot 2015-09-09 10 10 51

Here's a corrected version. See how the two compares side by side:
screenshot 2015-09-09 10 10 43


simongeilfus:
This is probably related to the recent changes with the default alpha testing. It could be solved with a ```gl::disableAlphaBlending()``` but imo it looks more "correct" to do it in the shader.

andrewfb:
Thanks for this...

Issue #1104, opened 2015-09-09
ES3: Texture1dRef referenced but not defined


BanTheRewind:
Including CINDER_GL_ES_3 in preprocessor macros for iOS and iOS-sim result in Texture1dRef being referenced without being defined in "include/cinder/gl/Texture.h:156", causing build to fail.

andrewfb:
Should be fixed by #1105

BanTheRewind:
👍

PR #1105, opened 2015-09-09
Fixes issue #1104

BanTheRewind:
Fixes issue #1104. It looks like these overloads are really just intended for VS. I updated gl/scoped to match the condition which defines Texture1dRef.


andrewfb:
Thanks for this catch & fix

PR #1112, opened 2015-09-13
Separated LoggerFile and LoggerFileRotating, added .h comments

lucasvickers:
Updates:

- LoggerFile and LoggerFile have been saparated in order to prevent any function overload issues, and because the classes are now cleaner.
- Fairly extensive comments have been added to the .h file, and tested against the doxNext docs generator.
- The ordering of class function definitions in the .cpp has been cleaned up a little.

### DIFF
--- a/include/cinder/Log.h
+++ b/include/cinder/Log.h
@@ -73,13 +73,18 @@ struct Metadata {
 extern std::ostream& operator<<( std::ostream &os, const Location &rhs );
 extern std::ostream& operator<<( std::ostream &lhs, const Level &rhs );
 
+//! Logger is the base class all logging objects are derived from.
+//!
+//! \see LoggerConsole, LoggerFile, LoggerFileRotating
 class Logger {
   public:
 	virtual ~Logger()	{}
 
 	virtual void write( const Metadata &meta, const std::string &text ) = 0;
-
+	
+	//! Setter for enabling or disabling timestamps inside this loggers messages.
 	void setTimestampEnabled( bool enable = true )	{ mTimeStampEnabled = enable; }
+	//! Getter for checking if timestamps are enabled in this loggers messages.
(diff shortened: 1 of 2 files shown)


lucasvickers:
Per comments found in [this issue](https://github.com/gregkepler/Cinder/issues/40), ```\breif``` has been removed from these comments.

include/cinder/Log.h
```diff
@@ -116,36 +125,66 @@ class LoggerFile : public Logger {
...
+};
+
+//! LoggerFileRotating will write to a specified file, either appending to or overwriting that
+//! file.  The filename will be re-evaluated during the first logging event occuring past midnight.
+class LoggerFileRotating : public LoggerFile {
+  public:
+	//! LoggerFileRotating constructor that creates a rotating log file.  The filename will be
+	//! re-evaluated when the first logging event occurs after midnight.
+	//! File appending is configurable via \p appendToExisting.  \p formatStr will be passed to
+	//! [strtime](http://www.cplusplus.com/reference/ctime/strftime/) to determine the file name.
```
richardeakin: About the length function, I think we could refer to the ctime function as just `strftime()` (and make sure to get the spelling right), users can search for docs on it via whatever manner they like. I also agree though, detailed docs on parameters in the header like this is okay.

lucasvickers:
Spelling aside, it's nice to have a link to strftime in the html docs.  It does come at the price of not especially pretty in-code documentation.

lucasvickers:
FYI, I'm going to move a bulk of the comments over to related html source, and streamline this file some more.  That should take the url out of the source code.

lucasvickers:
After gaining a better understanding of the html doc capabilities, I've minimized the in-code comments.

This is ready for review/merge.

lucasvickers:
This is going to be simplified and pulled into a large overhaul of logging.

Issue #1106, opened 2015-09-09
make Arcball take vec2 instead of ivec2


araid:
I often use ``Arcball`` to easily rotate 3D objects from the camera perspective. I call it from a variety of inputs other than the mouse, like touch, Leap motion or just keyboard.

The fact that ``mouseDrag`` takes an ``ivec2`` causes small rotations to be choppy due to integer rounding, as you can see here: https://gist.github.com/araid/274bf3f2bc6b4ece2ea7
That would be fixed by using ``vec2`` instead, and it wouldn't break anything.

In fact I would just rename all of its methods to get rid of the "mouse" part, since the class doesn't seem exclusive to mouse to me, but that's not really important.

notlion:
Also, at least on OS X mouse position is represented as floats and inputs like graphics tablets set the cursor to sub pixel positions.

lithium-snepo:
If this goes through, can it be applied to `CameraUi` as well?

andrewfb:
This makes sense to me, and I think the `CameraUi` change would too. Anybody up for a PR to this effect?

andrewfb:
Should be fixed by @lithium-snepo's commit above.

PR #1115, opened 2015-09-16
changed Arcball and CameraUi to take float vectors

lithium-snepo:
addresses #1106


andrewfb:
Thanks for this and apologies for the delay.

PR #1122, opened 2015-09-24
Handle \ continuations as part of a line.

### LINKED ISSUE #1121: ObjLoader fails to parse OBJ files with line continuations marked by `\`
When loading OBJ files that break vertex or face properties across multiple lines, parsing fails with an exception: `stoi: no conversion`.

Example problematic line (note the trailing `\` to indicate continuation on the following line):
```
f 27050/26938/27050 27041/26929/27041 27044/26932/27044 \
 27047/26935/27047
```

OSX Finder and Unity3D both parse the OBJ successfully, so I would expect the Cinder parser to handle this, too.

sansumbrella:
Add `IStreamCinder::readLineWithContinuation(char)` method to handle Unix-style line continuations in text files.
`ObjLoader` uses `readLinesWithContinuation` to enable parsing of OBJ files saved from Rhino/3DS Max that sometimes split faces/vertices across multiple lines.

Fixes #1121

### DIFF
--- a/include/cinder/Stream.h
+++ b/include/cinder/Stream.h
@@ -120,6 +120,7 @@ class IStreamCinder : public virtual StreamBase {
 	void		readFixedString( char *t, size_t maxSize, bool nullTerminate );
 	void		readFixedString( std::string *t, size_t size );
 	std::string	readLine();
+	std::string readLineWithContinuation( char lineContinuationChar = '\\' );
 	
 	void			readData( void *dest, size_t size );
 	virtual size_t	readDataAvailable( void *dest, size_t maxSize ) = 0;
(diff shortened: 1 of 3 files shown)


sansumbrella:
Performance comparison with previous implementation:
OBJ Loading on a (cleaned) 23MB file.

Previous implementation: `1596.96ms`
Lambda without continuation checking: `1564.28ms`
With continuation checking: `1653.18ms`

So adding continuation checking makes things about 3% slower than before. Unrolling the lambdas doesn't look like it would be any benefit (and perhaps would be counter-productive, given that the newline-only check in a lambda is faster).

The real benefit is that we can now properly load OBJ's that crashed Cinder applications before. Note that these aren't bad OBJ files, as they are successfully parsed by OS X's Finder and Unity3D.

sansumbrella:
On OSX, this is almost certainly the correct behavior for reading a line. I'm unsure how this affects Windows, though if Rhino is outputting files with `\` as a continuation character, it's probably the correct behavior there, too.

sansumbrella:
There are trailing `\` characters throughout the OBJ spec posted [here](http://www.martinreddy.net/gfx/3d/OBJ.spec).

Issue #1107, opened 2015-09-10
LogManager indiscriminately removes loggers by type


lucasvickers:
@richardeakin following our convo last night.

LoggerImplMulti's [removal](https://github.com/cinder/Cinder/blob/master/src/cinder/Log.cpp#L362-L369) routine remove any logger that matches a type.  Therefore it relies on only having one Logger of each type in the stack.

LoggerImplMulti's [add](https://github.com/cinder/Cinder/blob/master/src/cinder/Log.cpp#L53) routine, on the other hand, blindly pushes back Logger pointers.

The end result is that combining the helper functions such as ```enableFileLogging``` and the advanced usage of adding your own loggers via ```addLogger``` can create unexpected results.

Take for example the code below
```
#include "cinder/app/App.h"
#include "cinder/app/RendererGl.h"
#include "cinder/gl/gl.h"
#include "cinder/Log.h"

#include <algorithm>

using namespace ci;
using namespace ci::app;
using namespace std;

class RandomTestApp : public App {
  public:
	void setup() override;
	void mouseDown( MouseEvent event ) override;
	void keyDown( KeyEvent event ) override;
	void update() override;
	void draw() override;
	
protected:
	log::LoggerFile *log;
};

void RandomTestApp::setup()
{
	log = new log::LoggerFile( "/tmp/custom.log", true );
	
	log::manager()->addLogger( log );
	log::manager()->enableFileLogging( "/tmp/managed.log", true );
}

void RandomTestApp::mouseDown( MouseEvent event )
{
	CI_LOG_I( "test log" );
}

void RandomTestApp::keyDown( KeyEvent event )
{
	log::manager()->disableFileLogging();
	CI_LOG_I( "removed the managed log." );
}

void RandomTestApp::update()
{
}

void RandomTestApp::draw()
{
	gl::clear( Color( 0, 0, 0 ) ); 
}

CINDER_APP( RandomTestApp, RendererGl )
```
If you do the following steps.

1.  Run it
1.  Click the mouse
1.  ```tail -f /tmp/manage.log /tmp/custom.log```
1.  Click the mouse

Then the results you get are not as expected:
```
$ tail -f managed.log custom.log 

==> managed.log <==
|info   | 2015-09-09.21:03:28 virtual void RandomTestApp::mouseDown(cinder::app::MouseEvent)[34] test log

==> custom.log <==
[...]

PR #1127, opened 2015-09-27
Logging restructure

lucasvickers:
Related to #1107

#### Tasks completed in this rewrite
* LogManager's enable/disable routines were removed
* LogManager's interface was standardized with addLogger and removeLogger, etc
* LoggerMulti was removed and replaced with an internal vector of Loggers
* DebugTest was rewritten to be an easier to use test case
* Logger example updated
* Restructured locking so that there is no risk of crash, and wrote test case to confirm this

#### Performance
The good news is we now prevent a crash when adding and removing loggers from threads.  I confirmed the crash exists in the main branch, and confirmed the crash is fixed in this PR.

The bad news is that, in order to stop this crash, we had to move the lock mutex to the lowest level write() routine.  This really starts to show when you are logging from multiple threads.  

_For tests below, each thread logged 1 million times to a single log file._

##### Main branch (localized locks, not thread safe)
```
|debug  | void DebugTestApp::testPerformance()[244] Single logging took 0.01393 ms / log.
|debug  | void DebugTestApp::testPerformance()[248] 2 threads logging took 0.02628 ms / log.
|debug  | void DebugTestApp::testPerformance()[252] 3 threads logging took 0.06138 ms / log.
```

##### This branch (lock in write(), thread safe)
```
|debug  | void DebugTestApp::testPerformance()[280] Single logging took 0.01441 ms / log.
|debug  | void DebugTestApp::testPerformance()[284] 2 threads logging took 0.06125 ms / log.
|debug  | void DebugTestApp::testPerformance()[288] 3 threads logging took 0.10274 ms / log.
```

This is, of course, probably worse case since my test case has multiple threads logging from within loops.  I think we'll be able to improve this considerably when we can introduce shared_locks into the code base.

### DIFF
--- a/include/cinder/Log.h
+++ b/include/cinder/Log.h
@@ -81,7 +81,7 @@ class Logger {
 
 	void setTimestampEnabled( bool enable = true )	{ mTimeStampEnabled = enable; }
 	bool isTimestampEnabled() const					{ return mTimeStampEnabled; }
-
+	
   protected:
 	Logger() : mTimeStampEnabled( false ) {}
 
@@ -90,7 +90,9 @@ class Logger {
   private:
 	bool mTimeStampEnabled;
 };
-
+	
+typedef std::shared_ptr<Logger>	LoggerRef;
+	
 class LoggerConsole : public Logger {
   public:
 	void write( const Metadata &meta, const std::string &text ) override;
@@ -171,20 +173,16 @@ protected:
 //! This is primarily used by LogManager as it's base Logger, when multiple log outputs are enabled (ex. console and file)
 class LoggerMulti : public Logger {
   public:
-	void add( Logger *logger )							{ mLoggers.push_back( std::unique_ptr<Logger>( logger ) ); }
-	void add( std::unique_ptr<Logger> &&logger )		{ mLoggers.emplace_back( move( logger ) ); }
+	void add( LoggerRef logger )			{ mLoggers.push_back( logger ); }
 
-	template <typename LoggerT>
-	LoggerT* findType();
+	void remove( LoggerRef logger );
 
-	void remove( Logger *logger );
-
-	const std::vector<std::unique_ptr<Logger> >& getLoggers() const	{ return mLoggers; }
+	const std::vector<LoggerRef>& getLoggers() const	{ return mLoggers; }
 
 	void write( const Metadata &meta, const std::string &text ) override;
 
   private:
-	std::vector<std::unique_ptr<Logger> >	mLoggers;
+	std::vector<LoggerRef>	mLoggers;
 };
 
 class LogManager {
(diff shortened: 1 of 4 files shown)


lucasvickers:
The logic I was using to tally time / log was incorrect.  I was doing:
```
time_per_log = ( run1.duration + run2.duration ) / ( run1.num_logs + run2.num_logs )
```
Which is incorrect because the timers were running in a concurrent manner.  If I switch to using the max time any of the loggers took, things are a bit more reasonable:
```
time_per_log = std::max( run1.duration, run2.duration ) / ( run1.num_logs + run2.num_logs )
```
Produces these results:
##### Main branch (localized locks, not thread safe)
```
Single logging took 0.01446 ms / log.
2 threads logging took 0.01303 ms / log.
3 threads logging took 0.02161 ms / log.
```
##### This branch (lock in write(), thread safe)
```
Single logging took 0.01497 ms / log.
2 threads logging took 0.03369 ms / log.
3 threads logging took 0.03494 ms / log.
```

include/cinder/Log.h
```diff
@@ -159,92 +196,49 @@ class LoggerSystem : public Logger {
...
-	Logger* getLogger()	{ return mLogger.get(); }
-	//! Returns a logger of a specifc type, or nullptr is that type of Logger is currently not in use.
+	void removeLogger( const LoggerRef& logger );
+	//! Returns a vector of raw pointers that contain all active loggers of a specifc type.
 	template<typename LoggerT>
-	LoggerT* getLogger();
-	//! Returns a vector of all current loggers
-	std::vector<Logger *> getAllLoggers();
-	//! Returns the mutex used for thread safe loggers. Also used when adding or resetting new loggers.
+	std::vector<LoggerT*> getLoggers();
```
richardeakin: Since we're switching to LoggerRef's before, might as well return a `std::vector<LoggerRef>` here too (can be done with `std::dynamic_pointer_cast<LoggerT>()` in impl).

richardeakin:
Other than above comment, looks good to me. Thanks for getting through all of this!

include/cinder/Log.h
lucasvickers: I struggled with that a little bit.  I really wanted to return a ```std::vector<shared_ptr<LoggerT>>```, but returning a shared_ptr of a different template is problematic.

Part of me thought it was strange to return ```std::vector<LoggerRef>``` since you requested ```LoggerT```, but I think it's more consistent than the raw pointers, and makes more sense given the rest of the interface.

I'll make that change.

Thanks for the review!

richardeakin: Despite my suggestion above to return LoggerRef (wasn't thinking to deeply), what if we did it [like this](https://github.com/richardeakin/Cinder/commit/3e5d810fa29a5cfa7e4c03473ab7e3b380967d6e)?  That way if the type of logger you're requesting has special parameters you have the right type in the vector and can act on them directly, without needing to type cast again.

If this is agreeable, then I'll go ahead and merge the PR along with that change.

lucasvickers: Got it, I didn't fully understand the workings of ```std::dynamic_pointer_cast<LoggerT>()``` when you first mentioned it.  This is definitely the ideal solution.

PR #1128, opened 2015-09-27
Support App Transport Security on iOS

pizthewiz:
Bring over c56a404 from the OS X Kaleidoscope project for iOS - it appears as though Kaleidoscope is the only iOS sample that requests a URL. Tested with the Xcode 7.1 Beta 2 / iOS 9 simulator.

The Kaleidoscope project also needs to link to the Accelerate framework, but I ran into another issue building with Xcode 7.x and I'm thinking many of the iOS samples need updates.

The `cinder_iphone_sim` scheme of the Cinder project itself seems to have gone sideways and refuses to build with Xcode 7.x. To work around it to test this particular issue, I killed the `cinder_iphone_sim` scheme and simplified to have the `cinder_iphone` scheme support both the device and simulator, `lipo`'ed the `lib/ios/` and `lib/ios-sim` boost libs together, and then simplified the Kaleidoscope project settings to link to:
- `libcinder-iphone_d.a`
- `libcinder-iphone.a`

### DIFF
--- a/samples/Kaleidoscope/xcode_ios/Instascope-Info.plist
+++ b/samples/Kaleidoscope/xcode_ios/Instascope-Info.plist
@@ -24,6 +24,17 @@
 	<string>1.0</string>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
+	<key>NSAppTransportSecurity</key>
+	<dict>
+		<key>NSExceptionDomains</key>
+		<dict>
+			<key>api.instagram.com</key>
+			<dict>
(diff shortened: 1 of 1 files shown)


andrewfb:
If you delete everything out of your `cinder.xcodeproj` besides the `project.pbxproj` does the `cinder_iphone_sim` target work again? This is what I've had to do in the past when Xcode gets bent.

pizthewiz:
Sadly no dice on the `cinder_iphone_sim` target, I spun it out as #1131.

PR #1129, opened 2015-09-28
Perform line continuation logic in ObjLoader.

### LINKED ISSUE #1121: ObjLoader fails to parse OBJ files with line continuations marked by `\`

sansumbrella:
Closes #1121

### DIFF
--- a/src/cinder/ObjLoader.cpp
+++ b/src/cinder/ObjLoader.cpp
@@ -154,6 +154,11 @@ void ObjLoader::parseMaterial( std::shared_ptr<IStreamCinder> material )
         if( line.empty() || line[0] == '#' )
             continue;
 
+		while( line.back() == '\\' && ! material->isEof() ) {
+			auto next = material->readLine();
+			line = line.substr( 0, line.size() - 1 ) + next;
+		}
+
         string tag;
         stringstream ss( line );
         ss >> tag;
@@ -191,7 +196,12 @@ void ObjLoader::parse( bool includeNormals, bool includeTexCoords )
 		string line = mStream->readLine(), tag;
         if( line.empty() || line[0] == '#' )
             continue;
-        
+
+		while( line.back() == '\\' && ! mStream->isEof() ) {
+			auto next = mStream->readLine();
+			line = line.substr( 0, line.size() - 1 ) + next;
+		}
+
 		stringstream ss( line );
 		ss >> tag;
 		if( tag == "v" ) { // vertex
--- a/test/unit/src/ObjLoaderTest.cpp
+++ b/test/unit/src/ObjLoaderTest.cpp
@@ -1,23 +1,72 @@
 
 #include "catch.hpp"
 #include "cinder/ObjLoader.h"
+#include "cinder/TriMesh.h"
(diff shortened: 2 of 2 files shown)


richardeakin:
Will this work if the `next` line also ends with a `\` continuation char? Example from [here](http://www.martinreddy.net/gfx/3d/OBJ.spec):

```
cstype bmatrix
	deg 3 3
	step 3 3
	bmat u  1       -3      3       -1      \
		0       3       -6      3       \
		0       0       3       -3      \
		0       0       0       1
	bmat v  1       -3      3       -1      \
		0       3       -6      3       \
		0       0       3       -3      \
		0       0       0       1
```

sansumbrella:
Good catch. Yes.

Also added a check in case someone has a really poorly formatted file that ends with `\`. Not sure how critical that is.

paulhoux:
Ola, will this also work if there is empty space after the ```\```?

sansumbrella:
@paulhoux no it won't. Gotta love edge cases.

sansumbrella:
There isn't a `strip(const std::string &str)` function in cinder now by any chance?

richardeakin:
Do we need to support trailing whitespace? C preprocessor doesn't.

sansumbrella:
For now, I would opt to not support whitespace after the line continuation.

OSX Finder does manage to parse the OBJ with pathological whitespace. But trimming trailing whitespace is pretty lowhanging fruit for someone in control of an OBJ (where knowing there may be line continuations somewhere in your file is less so; and they may help a lot with human readability).

A bigger question: if we handle trailing whitespace after the line continuation marker, what do we do with it? Keep it as the start of the next line? Trim it and throw it away? What if there we other trailing characters other than whitespace?

paulhoux:
If trailing white space is not common in OBJ files, I have no problem if we don't support it here. There currently is no ```trim()``` method in Cinder, but there is in Boost and you can find a cross-platform solution here: http://stackoverflow.com/questions/216823/whats-the-best-way-to-trim-stdstring

sansumbrella:
Yeah, I didn't want to pull in the boost hammer since my understanding is we're trying to move away from it.

It seems like the line continuations themselves aren't all that common (since no one has reported an OBJ that broke the loader this way before). But some combo of 3DS and Rhino managed to create them.

sansumbrella:
This is ready to merge and solves problems I have encountered with OBJ files exported from Rhino.

Added unit tests for the OBJ parsing which helped narrow down where the crash was happening (and where line continuations resulted in wrong vertex data). Modifications to OBJ loader pass the new tests.

richardeakin:
Sorry for the delay, everything looks good here, thanks for the updates..

PR #1130, opened 2015-09-28
Logger restructure2

richardeakin:
See commit messages.

### DIFF
--- a/include/cinder/Log.h
+++ b/include/cinder/Log.h
@@ -82,9 +82,6 @@ class Logger {
 
 	virtual void write( const Metadata &meta, const std::string &text ) = 0;
 
-	template<typename LoggerT, typename... Args>
-	static std::shared_ptr<LoggerT> makeLogger( Args&&... args );
-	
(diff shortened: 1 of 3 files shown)


include/cinder/Log.h
```diff
@@ -82,9 +82,6 @@ class Logger {
 
 	virtual void write( const Metadata &meta, const std::string &text ) = 0;
 
-	template<typename LoggerT, typename... Args>
-	static std::shared_ptr<LoggerT> makeLogger( Args&&... args );
```
lucasvickers: My thinking behind this was: provide the user with a way to make a logger before inserting it into the LogManager.  Perhaps they wanted to adjust certain things before adding it to the LogManager?

richardeakin: That makes sense to me, but they can still use `std::make_shared<LoggerT>()` directly as the constructor is public. This did get me thinking, we should probably make Logger inherit from `ci::Noncopyable` at this point, to indicate that they should be created on the heap. I'll add that to this PR.

lucasvickers: both make sense to me.

PR #1137, opened 2015-09-30
Double quote HTML attributes in doc sources

pizthewiz:
This change just resolves the issue when editing the HTML source files, it appears as though the generator fixes the issue at build time so it isn't seen by doc consumers.


pizthewiz:
My file editor flags validation issues so this is just some low-hanging fruit.

gregkepler:
hey there @pizthewiz, thanks for chiming in. This was actually something I already made an adjustment to today. It should be fixed in the merge that's coming this evening. If it looks like there's still something wacky there, we'll take a look again.

pizthewiz:
Ahh good deal, thanks!

PR #1140, opened 2015-10-01
Normalize GitHub, CinderBlock, Xcode and OS X naming

pizthewiz:
Use a handful of regular expressions to normalize naming of a couple of terms. Not particularly important but consistency isn't so bad. :wink:


andrewfb:
Good stuff man - regexes are even more anal retentive than I am. Thanks for this.

pizthewiz:
Ahh, I mostly needed them to avoid stomping links and filenames.

PR #1144, opened 2015-10-01
Fix broken CinderBlocks guide render

pizthewiz:
The page wasn't rendering properly due to couple of broken tags and I've also removed the `armv7s` reference and replaced it with `arm64` given [glNext: removing armv7s](https://forum.libcinder.org/topic/glnext-removing-armv7s).


andrewfb:
Thanks again...

PR #1148, opened 2015-10-02
Updated git guide. Fixed outdated/broken links.

gregkepler:


gregkepler:
Not sure if this should be merged into 0.9.0 or master if that ship has sailed.

andrewfb:
Are these holding us back on addressing the broken link on the site? If not I'd like to look at them tonight as there's one or two very minor things I'd tweak. Also I would actually generally write `0.9.1dev` rather than `0.9.1.dev`.

PR #1149, opened 2015-10-04
Added logging guide

lucasvickers:
Finally finished the logging guide.  Let us all bask in its glory.

PR #1150, opened 2015-10-04
Added supplementary html for LoggerFile and LoggerFileRotating

lucasvickers:
Added some supplementary html for the LoggerFile and LoggerFileRotating classes

### DIFF
--- a/docs/htmlsrc/_docs/log/Logger.html
+++ b/docs/htmlsrc/_docs/log/Logger.html
@@ -8,22 +8,30 @@
 	<!-- 
 	cinder::ChannelT prefix 
 	-->
-    
-    <ci prefix dox="cinder::log::LoggerFile">
-	<br>
-    <br>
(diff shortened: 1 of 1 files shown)


lucasvickers:
@gregkepler formatting updated based on your comments to #1149.  Let me know if you have any other comments.

richardeakin:
Is this still needed in addition to #1149?

lucasvickers:
This is supplementary html that gets pulled into the class reference, so it's different then #1149 which was a standalone guide.

richardeakin:
Happy to merge this, although I think something still needs to be added to [this page](https://libcinder.org/docs/branch/master/guides/index.html) for the logging guide to show up - should we try to get both in at that point?

gregkepler:
Yeah, it will need to be linked up from the Guides page. Not sure if it makes sense to be under "Working with Cinder" or "Cinder APIs".

richardeakin:
I think Cinder APIs makes more sense, since this generally how you use different modules within ci::. The 'Working with Cinder' section is more conceptual, from what I see.

PR #1157, opened 2015-10-10
Fix leaked NSScreen instances from app::PlatformCocoa::getDisplays()

pizthewiz:
Unlikely to be a big deal, but the `NSScreen` instances were retained without a paired release - I think it was forgotten from ed74713b. Quickly verified via _test/DisplayTest_.


richardeakin:
Thanks piz.

PR #1168, opened 2015-10-26
Unit testing framework

sansumbrella:
There are currently a handful of ad-hoc schemes for unit testing in Cinder's `test/` directory and many of the projects in that directory have stagnated. To help alleviate confusion, I propose a single project for running all unit tests and the use of the [CATCH](https://github.com/philsquared/Catch) framework to write and run them.

This change brings a handful of benefits:
- simpler to add new tests
- easier to run all tests
- easier to maintain the tests
- promotes more readable tests

I  quickly copied some existing tests into the new project and made them use CATCH macros instead of custom testing harnesses. They should give an idea of what the transformation looks like. These include:
SignalsTest, base64Test, SystemTest

Some of those tests output a lot to console; aside from base64's count to 1000, I left that console output alone.

I haven't made a visual studio project yet since I only have an OSX machine at work.


richardeakin:
I'm all for moving to the CATCH unit testing framework, I like that it is all contained in a single header and it seems to be widely used. Thanks for initiating this.

I'm trying to think whether it will be more maintainable / testable if the test source files are all in one folder, or separated into their own folders like before (SignalsTest in the signals folder, I've wrote some for audio in the test/_audio/unit folder). While I think there's obvious benefit to running all the tests at once (like testing before a big merge or release), I also like being able to quick run just the tests pertaining to one topic, ex. Signals when something changes in cinder/Signal.h. Is there a way we can manage both?

One thing that comes to mind is to use CMake for the unit tests, for the most part they can be run from the command line anyway.

Also I think we should stick to CamelCase file naming where possible for consistency.

So I guess my main point of contention is how we organize the tests, while I'm very much in support of us adding CATCH to facilitate this.

sansumbrella:
Running tests from the command line would be great (the XCode target is a 'command line program', after all). I'm not familiar with setting up CMake for these things, so pointers would be helpful. Or if it is simpler to just use Make by itself (probably not with multiple platforms), that would be cool, too. Catch has an example [CMake external project configuration](https://github.com/philsquared/Catch/blob/master/docs/build-systems.md#cmake), though I'm not yet sure what to make of it.

Underscores in filenames are just there as a result of my test template. I'm not partial one way or the other.

Catch provides a mechanism for running specific tests by passing command-line options. Once the unit tests are built, you can run them from the command line. For example:

```bash
# list the available test cases
./CinderUnitTests -l
  Basic Signal Tests
  Signals
  Priority group disconnection
  Scoped connections 
  Signal result collection
  Connection Toggling
  System_test
  ObjLoader
  Base64_test

# Run a specific test case, in this case, signals
./CinderUnitTests "Signals"
```

See [specifying which tests to run](https://github.com/philsquared/Catch/blob/master/docs/command-line.md#specifying-which-tests-to-run) for more.

sansumbrella:
Playing with that option makes me think I would prefer a single TEST_CASE at each file level, since you probably won't want more granularity for running a specific group of tests than that.

sansumbrella:
Also, I haven't played with the [tags in catch](https://github.com/philsquared/Catch/blob/15317632f3caf59a3ba3da2ec3dc3ee189bbbc9a/docs/test-cases-and-sections.md#tags), but they seem powerful if you want to group tests across files.

richardeakin:
Setting up to build with cmake is pretty straightforward especially if you're just linking to libcinder.a. For command line you don't need all the extra cruft to build a .app or MSW windowed exe. It's certainly easier to maintain accross platforms with CMake rather than using Makefiles. But I can add these later as well, we're planning to add cmake support sometime soonish so unit tests can be covered then too.

Catch's test specifying mechanism sounds good for running the tests, but it still means you'd have to build them all at the same time. I guess my concern is just that, build times (or if one test is failing to build. I also suppose this is offset in repeated builds when you're only modifying one test cpp.

I think one example we might be able to look at is GLM, I know it uses CMake for its unit testing, although not Catch.

I guess I'm not opposed to moving the audio unit tests into a general unit tests folder. In that case I'd like to propose the following layout (keep in mind there hasn't been much discussion about this with the rest of the team):

test/                   <--- our test apps for dev purposes
test/unit/            <--- all unit tests in here
test/unit/audio/  <--- ci::audio specific unit tests
test/unit/gl/        <--- ci::gl specific unit tests

We could also use a scheme like this to name the tests for the test runner (ex. audio/Fft, or gl/ShaderPreprocessor).

Thanks again David, I think this will go a long way in promoting people to add some test code when they submit PRs for new features.

sansumbrella:
Grouping the tests by the namespace they test makes sense to me. I would like to keep all the unit tests grouped together, since it makes it easier to find/manage them.

```
test/unit/
- tests.cpp # main file for tests
- cinder/
  - audio/
  - signals/
  - gl/
  - svg/
```

And as you suggested, naming test cases based on their namespace (ignoring cinder::) would make it easy to navigate what is available and filter tests.

I'll keep massaging the directory structure on this PR. Will just use Xcode until I get some time/help for the CMake stuff.

richardeakin:
That all sounds great from my point of view, briefly spoke with Andrew and he's on board with it too. Let us know when you think its ready to be merged.

richardeakin:
I saw you made a bunch of commits, how is this coming along?  Also, what is the CinderUnitTests.dSYM folder, is that meant to be checked in?

After looking at the folder structure locally, I don't see the added benefit of the extra cinder folder in tests/unit as all these tests are for cinder code and it doesn't map to a namespace or anything. Not that I'm hugely against it if this is what you prefer, just seems like something we could do without.

I ran the UnitTests project here, looks fine for a first pass. We need to add one for VS but I'm not in front of a windows computer, and we can do it later anyway.

richardeakin:
One other thing, seems like a good idea to move the catch.hpp and TestsMain.cpp file into a test/unit/include or test/unit/src folder. To me it makes sense to move all the unit testing headers and cpps into test/unit/src/*, which will make things clearer once we start adding build files for more platforms.

sansumbrella:
Yeah, I think it's in pretty good shape for a first pass. Thanks for giving it a look.

Are you suggesting essentially renaming the `cinder` folder to `src` and moving `catch.hpp` and `TestsMain.cpp` in there? That makes sense to me.

We could remove the existing Signals unit tests (though not the performance tests) before merging, since they are now covered by the catch unit tests. I deleted most of the other old tests that I replicated, especially since their project files wouldn't build against the latest Cinder.

sansumbrella:
I'll take a look at moving the files once more tomorrow and double-check the Signals unit tests before deleting the previous unit test file. Again, leaving the perf tests alone since they aren't in the unit tests.

richardeakin:
Yea that all sounds good, on removing the old signals test too. It'll be nice to have catch.hpp shipping with cinder, we can remove yet one more dependency on boost.

sansumbrella:
I cleaned up those last two bits we talked about. I think this is ready to merge as a preliminary unit test approach.

sansumbrella:
@richardeakin I removed the .dSYM file. It's something XCode creates when you make a command-line app target, but isn't needed.

sansumbrella:
Probably the preferred way to run these right now is

```
$ ./CinderUnitTests ~[noisy]
```

This way you avoid running the random distribution tests that plot to console.

richardeakin:
Thanks again! This is a great addition.

PR #1175, opened 2015-11-02
Added a TorusKnot primitive.

paulhoux:
Also fixed a few minor issues with some of the primitives.

### DIFF
--- a/include/cinder/GeomIo.h
+++ b/include/cinder/GeomIo.h
@@ -559,12 +559,65 @@ class Torus : public Source {
 	int			mNumRings, mNumAxis;
 };
 
+class TorusKnot : public Source {
+public:
+	TorusKnot();
+
+	// Enables colors. Disabled by default.
+	TorusKnot&	colors( bool enable = true ) { mHasColors = enable; return *this; }
+	TorusKnot&	subdivisionsAxis( int subdiv ) { mSubdivisionsAxis = subdiv; return *this; }
+	TorusKnot&	subdivisionsHeight( int subdiv ) { mSubdivisionsHeight = subdiv; return *this; }
+	//! Specifies the \a P and \a Q parameters of the torus knot.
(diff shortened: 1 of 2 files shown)


paulhoux:
Here's a few images to show you what it looks like:

include/cinder/GeomIo.h
```diff
@@ -559,12 +559,65 @@ class Torus : public Source {
...
+	Primitive	getPrimitive() const override { return Primitive::TRIANGLES; }
+	uint8_t		getAttribDims( Attrib attr ) const override;
+	AttribSet	getAvailableAttribs() const override;
+	void		loadInto( Target *target, const AttribSet &requestedAttribs ) const override;
+	TorusKnot*	clone() const override { return new TorusKnot( *this ); }
+
+protected:
+	void		calculate( std::vector<vec3> *positions, std::vector<vec3> *normals, std::vector<vec2> *texCoords, std::vector<vec3> *colors, std::vector<vec3> *tangents, std::vector<uint32_t> *indices ) const;
+
+	inline int	gcd( int a, int b ) const
```
richardeakin: Any reason this function is in the header, or could it be moved to a freestanding method in the cpp's anonymous namespace?

richardeakin:
Looks pretty sweet.. both the screenshots and the impl.

include/cinder/GeomIo.h
paulhoux: I agree it is a nice function to have in Cinder anyway, but since I only needed it here, and I didn't want to clutter ci::math<> (as I don't know how long it will be around), I decided to inline it in the class itself.

richardeakin: I still don't see the benefit of having it in the header, which is usually where you want to look at as an api reference, rather than in GeomIo.cpp, where the implementation details live.

Though a `gcd()` function in CinderMath would be fine to me, I think it could potentially be more involved to work across all desired types... so I think placing it in GeomIo.cpp in an anoynmous namespace would be ideal.

PR #1178, opened 2015-11-04
Additional ray-sphere intersection code, similar to AxisAlignedBox.

paulhoux:
Also added a convenience method to render the sphere.


richardeakin:
Looks good to me, nice addition.

PR #1179, opened 2015-11-04
Added new OSC block, removed existing. Updated TUIO block.

ryanbartley:


pizthewiz:
One other thing that might need some attention is the TUIO block and samples which depended on the old OSC block right?

ryanbartley:
Working on the patch.

pizthewiz:
Strong work, especially when I'd have to imagine the TUIO stuff is both pretty gnarly and probably not heavily used.

ryanbartley:
:) gnarly is a perfect word for it!

richardeakin:
Lotsa samples, nice. However I don't see the src/Osc.h and src/Osc.cpp files.

Also I see there are app icons for the samples, can you remove those and point to the ones in samples/data folder? We're planning to do a purge one day and each icon we can avoid reduces the binary package by about half a meg.

richardeakin:
Ah, I see - Osc.h / cpp are at the very end but their diffs don't show up ('maximum number of lines exceeded.')... not a problem though to view them, though.

richardeakin:
Hey Ryan, this looks great, you've accomplished a huge upgrade. While I still haven't run the code but I assume you and @pizthewiz have so impl is correct, a couple nits I wanted to bring up:

* Should the classes be in the cinder namespace (more specifically, `ci::osc` like they were before)? One benefit I see of this is that it will be distinguished from code that uses raw oscpack, which is / was in the top-level osc namespace.
* I think it is a good style to place the body of `if()` statements on a new line (sorry I can't comment in-line where I saw this because github isn't allowing me to, but for example [here](https://github.com/ryanbartley/Cinder/blob/7f8e307786f96a5eb894975313c3a24e4703d47d/blocks/OSC/src/Osc.cpp#L464-L476)). Other than improving legibility (IMO), the main benefit of this is that when you're stepping through code it is easy to tell when you made it into the if block, or if skipped over it. Given that many debuggers like to stutter on things like shared_ptrs or for other odd reasons, landing directly in the body makes it clear what's going on.
* Can you mark the end of the namespace with a comment (like [here](https://github.com/ryanbartley/Cinder/blob/7f8e307786f96a5eb894975313c3a24e4703d47d/blocks/OSC/src/Osc.h#L918)). Also there's a stray semicolon [here](https://github.com/ryanbartley/Cinder/blob/7f8e307786f96a5eb894975313c3a24e4703d47d/blocks/OSC/src/Osc.h#L916) - ok now I admit I'm really nitpicking. :)

Thanks for this, we've needed to move away from oscpack for a long time!

ryanbartley:
Finally updated the Tuio Block, going to add an RFC for this and then I'll fix those changes you requested, Rich. There are also a few minor changes that I'm going to make to Osc after working on this block.

pizthewiz:
I don't have any history with TUIO but I'd be happy to test if that would help push things forward.

ryanbartley:
This is ready to merge. I also tested on Ubuntu and it works great. I'll do some more testing on a couple of embedded devices as soon as it's merged. Thanks for the help guys! Let me know if there are any other outstanding things.

richardeakin:
Thanks for making those changes to the OSC stuff. On TUIO, is there any reason that the statc [Object createFromSetMessage()](https://github.com/cinder/Cinder/blob/dd16254f0f4ab2276df845f45b604355e64299f2/blocks/TUIO/include/TuioObject.h#L47-L60) isn't an overload constructor? In general static create methods return a shared or unique pointer.

ryanbartley:
I believe that is from the old Tuio block.

richardeakin:
Ah, I see alot of that is - sorry, hard to tell what is new as I'm not able to use github diffs on this PR (github complains it is too long).

ryanbartley:
Sorry about that. This is definitely a big and unfortunately tethered PR that deals with multiple issues.

ryanbartley:
you can see the actual tuio specific code if you look at my [branch](https://github.com/ryanbartley/Cinder/tree/OscBlockUpdate/blocks/TUIO)

paulhoux:
Gratz @ryanbartley  and thanks for your commitment and hard work.

ryanbartley:
:+1:

kitschpatrol:
This is great!

I've been testing it out, and have run into some trouble with failed reconnections after restarting an app with a TCPReceiver. I ended up seeing improvements by setting the reuse_address option to true on the TCPReceiver acceptor like so: https://github.com/kitschpatrol/Cinder/commit/d739faaa8b732b1b5ffbc12f75e3441416ac4f5f

Wondering if anyone else has had similar issues, and if the above fix has any downsides.

ryanbartley:
Yeah, that's a common problem with Tcp Connections, specifically. Basically, your app is finished and closed out and not receiving anything else, but the OS still has the socket open. The socket::close function is more akin to a "wish list" that you'd send to santa. :) If you do a rapid quit and restart, more than likely the OS is still holding on to the socket, waiting for any last second messages because of the OS level timeout for TCP Connections. 

Setting the reuse_address is a good way to address the problem. However, it's OS dependent whether the request is actually taken into account, like the close function, Windows is particularly murky in this scenario. It may not be necessary to add the set_option inside of the bind function, as you can just create your acceptor, set the options and then hand it to us [here](https://github.com/cinder/Cinder/blob/master/blocks/OSC/src/Osc.h#L786).

sansumbrella:
I was chatting with @pizthewiz about this in a different context and I generally think setting reuse_address is a good idea.

I have had times in the past (untested with the new OSC block) where I wanted to have multiple applications using the same TCP socket on my computer. Usually when testing communication between many instances of the same application. Socket creation would fail if I didn’t allow multiple connections on the socket.

I am not as familiar with all the networking stuff as @ryanbartley, so I'm not sure whether that kind of configuration would work on all platforms. What is the downside to (requesting) allowing address reuse that I'm missing? It seems like a sane default to me.

ryanbartley:
Short answer is that there isn't much of a downside to this change, but that it won't work all of the time, which makes me hesitant to add this to the implementation. Sometimes even when using the SO_REUSEADDR command, the socket may be doing something that won't allow it to be reused, such as a buffer filled waiting to be read or sent. This [_article_](http://stackoverflow.com/questions/14388706/socket-options-so-reuseaddr-and-so-reuseport-how-do-they-differ-do-they-mean-t) explains a lot of the reasons that it happens. 

TLDR (which apologies that it may still be TLDR), "for TCP sockets, there can be a relatively long delay between adding data to the send buffer and having the TCP implementation really send that data. As a result, when you close a TCP socket, there may still be pending data in the send buffer, which has not been sent yet but your code considers it as sent, since the send() call succeeded." This causes what is called a TIME_WAIT phase for the socket which is configurable, but most systems have it as long as 2 mins. Down below that in the connect section, he speaks about the EADDRINUSE error received from calling the connect function on a socket. Basically, there are 5 values that differentiate a socket, {<protocol>, <src addr>, <src port>, <dest addr>, <dest port>}. If you were to shut down your application and restart it in such a time that the original socket was in a TIME_WAIT, you'd most likely try to rebind your address to the same 5 numbers. "If data arrived for either one of the two connections [socket in TIME_WAIT, or socket just opened], the system could not tell which connection the data belongs to." Hence, the error. Even if you used SO_REUSEADDR, you'd still get an error because you've chosen the exact set of numbers that you just can't choose. 

There's another [_article_](http://serverfault.com/questions/329845/how-to-forcibly-close-a-socket-in-time-wait) that echoes this but from a different perspective. "[SO_REUSEADDR] option tells the kernel that even if this port is busy (in the TIME_WAIT state), go ahead and reuse it anyway. If it is busy, **but with another state**, you will still get an address already in use error. It is useful if your server has been shut down, and then restarted right away while sockets are still active on its port." So, also if the socket is within another state other than TIME_WAIT, you'll get an address already in use error. 

I thought about this addition a lot while writing this because it seemed to make the problem better, but realized that it just seemed I had more luck with the specific state of the system more than I fixed the problem. Basically, the socket is doing a ton of work to make sure the connection state is reliable when you actually get connected. When you disconnect, it has to continue that reliability after your gone, either resending lost packets or waiting to receive expected packets or just waiting to hear anything at all from the other side.The real problem is that TCP is complicated and properly shutting down a socket [_can_](https://msdn.microsoft.com/en-us/library/windows/desktop/ms738547(v=vs.85).aspx) be complicated. The gist is, call shutdown ( which, now I realize, should probably be a part of the TCP interface ), then read from the socket, to fully empty the buffers or receive any other "errors" that the socket has, and finally close the socket. Even this could create the side effect as the OS won't necessarily tell you when it's TIME_WAIT is up. Sorry about how long this post became :( Hope this all makes sense and I'll look at the addition of a shutdown method soon.

kitschpatrol:
Wow thank you for these very helpful responses and links. Clearly this is a bit of an iceberg. Good to know that REUSEADDR improves the odds but doesn't actually deliver any guarantees. Sounds like shutdown is the way to go. Thanks again.

PR #1181, opened 2015-11-04
Changed hard-coded _WIN32_WINNT values to their readable counterparts.

paulhoux:


richardeakin:
So Andrew has been telling me that there are cases where `_WIN32_XXX` isn't defined yet at the time the preprocessor is running - I was using this for win audio stuff but its been slowly moving to the hex values. Does anyone know for sure if it is safe to use things like `_WIN32_WINNT`? I do like that it is more readable, but then if it's not guaranteed to be defined then its problematic.

andrewfb:
I'd say in general these sorts of constants should be numeric because the "future" will not be defined in the "past." For example, a previous version of the XP SDK would not have `_WIN32_WINNT_VISTA` defined yet. Paul apparently tested this with the toolsets we actually support though (including v120_xp). To some extent I think of this as a question of form, and idiomatically you would use the numeric value and then comment, as we were previously. However it doesn't strike me as too big a deal either way.

paulhoux:
Sorry for just merging and then not weighing in afterwards. 

I compiled Cinder for all targets and architectures using the v120_xp and v140_xp toolsets, to make sure they all knew what "Vista" was. So at least I am pretty sure we didn't break anything, as long as people use VS2013 and up.

Note that "Win8" is not defined in v120_xp, so if we have Win8-specific code, we'd still need to use 0x0602. 

I like the readable version of the define, but if you'd rather use the numeric version, that's fine with me.

Issue #1182, opened 2015-11-04
Can't Download From Website


computersarecool:
Not sure if this is the right place to put this but the link on the websites "Getting started with Windows page"  (https://libcinder.org/docs/guides/windows-setup/index.html) specifies the download link as:

https://libcinder.org/download/

That returns a 404 as the last back slash needs to be removed (i.e. to this: https://libcinder.org/download)

However, once I figured that out and downloaded the zip, and VS (on Windows), I get the following error, 
`cannot open file 'cinder-v140_d.lib'`
(I followed the directions from the website but just installed VS 2015 so it might be a problem with that) 

Any ideas on this?

pizthewiz:
Good catch, it appears as though the [OS X Setup](https://libcinder.org/docs/guides/mac-setup/index.html) guide also points to the wrong, non-HTTPS link. You could create  a PR changing https://github.com/cinder/Cinder/blob/master/docs/htmlsrc/guides/windows-setup/index.html#L28:
```html
<p>This guide will help you get started using Cinder on Windows. Start by downloading Cinder itself from the site's <a href="http://libcinder.org/download/">Download Page</a> if you haven't already. Choose the Visual C++ 2013 version if you don't already have Visual C++ installed.</p>
```

and https://github.com/cinder/Cinder/blob/master/docs/htmlsrc/guides/mac-setup/index.html#L22:
```html
<p>If you are using one of our <a href="http://libcinder.org/download/">packaged releases</a>, you already have everything you need. If you're working from GitHub you'll want to follow the build instructions in our <a href="../git/index.html">Cinder + Git</a> guide.</p>
```

computersarecool:
Thank you, however, as I am just beginning to look at Cinder, I do not think I am quite ready to fork the repo just to send a PR for those two typos.

paulhoux:
Hi @computersarecool ,

the missing ```cinder-v140_d.lib``` file is due to the fact that you're using VS 2015 and there are only precompiled version of the library for VS 2013. No worries, though. Simply open the ```cinder.sln``` file and compile both the ```Debug``` and ```Release``` targets in VS 2015 and you should be good to go.

computersarecool:
I did that and it worked. Thank you kindly @paulhoux

PR #1183, opened 2015-11-07
Fix download links and migrate links to https

pizthewiz:
Fixes part of the issue @computersarecool reported in #1182.

Might be worth merging into release_v0.9.0 if that's what on the site now?


andrewfb:
Thanks for this - we're looking at tweaking the server config for this too.

PR #1185, opened 2015-11-09
Fixed parameter type to glStencilMask.

paulhoux:
```gl::stencilMask``` took a GLbool instead of a GLuint.


richardeakin:
Nice catch.

PR #1187, opened 2015-11-09
Updated Geometry sample.

paulhoux:
* Added support for Torus Knot.
* Each primitive can now have its own tweakable parameters.
* Camera recenter now maintains view direction.
* Camera recenter uses fixed timestep.

### DIFF
--- a/samples/Geometry/assets/phong.frag
+++ b/samples/Geometry/assets/phong.frag
@@ -2,6 +2,7 @@
 
 uniform sampler2D	uTex0;
 uniform int			uTexturingMode;
+uniform ivec2       uFreq;
 
 in VertexData	{
 	vec4 position;
@@ -13,7 +14,7 @@ in VertexData	{
 out vec4 oFragColor;
 
 // Based on OpenGL Programming Guide (8th edition), p 457-459.
-float checkered( in vec2 uv, in int freq )
+float checkered( in vec2 uv, in ivec2 freq )
 {
 	vec2 checker = fract( uv * freq );
 	vec2 edge = fwidth( uv ) * freq;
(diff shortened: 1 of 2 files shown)


paulhoux:
Can anyone try this on the Mac?

pizthewiz:
Looks good to me on a MacBook Pro with an NVIDIA GeForce GT 750M running OS X 10.11.2b2.

screen shot 2015-11-13 at 09 25 59

pizthewiz:
There was a warning logged to the console, but it didn't seem fatal:
```
|warning| void cinder::gl::GlslProg::logMissingUniform(const std::string &) const[938] Unknown uniform: "uFreq"
```

richardeakin:
Paul I know you like the fixed time step, but do you really think it is necessary for an already complicated sample? It doesn't buy much in this case, and is unrelated to the purpose this sample ties to demonstrate.

paulhoux:
Without it, on slow systems (like my 30Hz UHD monitor), re-centering the camera becomes too sluggish. Can't use a Timeline either, because people might want to move the camera while it is re-centering. I agree that it complicates the code a bit, but I don't see this sample as an actual sample anyway. It's a showcase of all geometric primitives, not a tutorial.

richardeakin:
I understand the want to keep the sample animation speedy on low end machines, although at the same time I do see this is code as a sample, actually its the first sample users will look at to try to understand how to use the geoms. If you recall, a while back I spent some time refactoring it to make it a bit more digestible for this reason (added helper routines, moved shaders to external glsl files, things like that).

In any event, lets try not to blow it up any more, unless that means more geoms. :)

richardeakin:
Hey Paul, just tried running your updates here on mac and I'm noticing that the call to `settings->disableFrameRate()` is causing the animation to be much too fast. Not sure why, as the fixed time step is supposed to be a problem to this, but it is clearly moving 5x the speed it is supposed to (getting around 250fps here).

paulhoux:
Hi Rich, if with "animation" you mean the camera re-centering, then I wasn't able to reproduce your issue.  It took the same amount of time regardless of the frame rate. However, I've removed the fixed timestep because you'd never accept the sample otherwise. Also fixed the ```uFreq``` warning that @pizthewiz mentioned.

richardeakin:
Hah, well no that's not true I was planning to merge if it worked fine on Mac, but it looked like the camera was on crack here.  I just did a checkout of your latest changes though and it looks fine to me at both 60 and 30 fps on Mac, so I think this is good to go.

PR #1188, opened 2015-11-12
GLSL preprocessor regex replacement

andrewfb:
Replaces regex's we've been using for preprocessing `#version` and `#include` directives in GLSL with handwritten equivalents. The primary motivation is to remove our dependency on std::regex to simplify Linux + GCC ports. A secondary benefit is improved performance, on MSW in particular. These changes benchmark as 5.5x perf in Debug and 3.7x perf in Release for me.

### DIFF
--- a/src/cinder/gl/ShaderPreprocessor.cpp
+++ b/src/cinder/gl/ShaderPreprocessor.cpp
@@ -27,15 +27,131 @@
 #include "cinder/Utilities.h"
 #include "cinder/Log.h"
 
-#include <regex>
-
 using namespace std;
 
 namespace cinder { namespace gl {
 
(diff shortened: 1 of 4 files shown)


pizthewiz:
While I curse gcc-4.8, the performance difference is interesting, could `std::regex` have really been that inefficient, even in libc++?

andrewfb:
In very broad strokes, http://www.joshbarczak.com/blog/?p=970 sees similar perf deltas. Obviously not 1:1 comparison for a variety of reasons though. I haven't dug too deep, but I'd imagine things like inlining and similar are nearly impossible when your input into std::regex is a string.

PR #1189, opened 2015-11-12
Tweaked the SMAA sample.

paulhoux:
By taking advantage of shader pre-processor defines, we only need a single vertex and fragment shader file.


richardeakin:
Very nice updates, needed to update the xcode project but worked great after I did that here.

One thing though, I'm seeing two `gl::Texture2d` warnings from [this](https://github.com/cinder/Cinder/blob/4637ac041ef647326bcc988778fe73481d196eae/samples/_opengl/PostProcessingAA/src/smaa/SMAA.cpp#L59) and [this](https://github.com/cinder/Cinder/blob/4637ac041ef647326bcc988778fe73481d196eae/samples/_opengl/PostProcessingAA/src/smaa/SMAA.cpp#L65) texture create, saying `loadTopDown( true )` was ignored. @paulhoux or @andrewfb do you know why this is and should we remove those calls from the sample?  Maybe it is because the texture is created with raw bytes?

paulhoux:
Yeah, I think that's the reason. I will get rid of the warnings tonight. Thanks for letting me know!

paulhoux:
See #1204 .

PR #1192, opened 2015-11-16
Added the missing platonic solid and their wireframe version.

CCortex:
Hi,

I added some geometry :
Tetrahedron
Octahedron
Dodecahedron

The only missing thing is the "right" uv text.
Tell me guys how you want it implemented.

### DIFF
--- a/src/cinder/GeomIo.cpp
+++ b/src/cinder/GeomIo.cpp
@@ -1350,6 +1350,368 @@ void Icosphere::loadInto( Target *target, const AttribSet &requestedAttribs ) co
 }
 
 ///////////////////////////////////////////////////////////////////////////////////////
+// Tetrahedron
+float Tetrahedron::sPositions[4*3] = { 
+    1.0f, 1.0f, 1.0f,    -1.0f, 1.0f, -1.0f,   1.0f, -1.0f, -1.0f,    -1.0f, -1.0f, 1.0f };
+
+float Tetrahedron::sTexCoords[12 * 2] = {
+    0.0f, 0.0f, 0.0f, 1.0f, 1.0f, 0.0f,
+    0.0f, 0.0f, 0.0f, 1.0f, 1.0f, 0.0f,
+    0.0f, 0.0f, 0.0f, 1.0f, 1.0f, 0.0f,
+    0.0f, 0.0f, 0.0f, 1.0f, 1.0f, 0.0f};
+
+uint32_t Tetrahedron::sIndices[12] = {
+    1, 2, 3,
+    0, 3, 2,
+    0, 1, 3,
+    0, 2, 1};
+
+Tetrahedron::Tetrahedron()
+    : mHasColors( false )
+{
+}
+
+size_t Tetrahedron::getNumVertices() const
+{
+    return 12;
+}
+
+size_t Tetrahedron::getNumIndices() const
+{
+    return 12;
+}
+
+void Tetrahedron::calculate( vector<vec3> *positions, vector<vec3> *normals, vector<vec3> *colors, vector<vec2> *texcoords, vector<uint32_t> *indices ) const
+{
+    positions->reserve( 12 );
+    normals->resize( 12 ); // needs to be resize rather than reserve
+    colors->reserve( 12 );
+    texcoords->reserve( 12 );
+    indices->reserve( 12 );
+
+    for (size_t i = 0; i < 12; ++i) {
+        positions->emplace_back( *reinterpret_cast<const vec3*>(&sPositions[sIndices[i]*3]) );
+        texcoords->emplace_back( *reinterpret_cast<const vec2*>(&sTexCoords[i*2]) );
+        indices->push_back( (uint32_t)i );
+    }
+
+    // calculate the face normal for each triangle
+    size_t numTriangles = indices->size() / 3;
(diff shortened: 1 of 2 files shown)


paulhoux:
Oh hey, thanks for the pull request. I guess it's kind of hard to come up with a good mapping algorithm for these solids, again fueling the a discussion we had earlier about separating the UV-mapping from the primitives completely.

For now, we could simply map the solids in a similar way as the icosahedron, where each triangle is mapped independently.

src/cinder/GeomIo.cpp
```diff
@@ -1350,6 +1350,368 @@ void Icosphere::loadInto( Target *target, const AttribSet &requestedAttribs ) co
...
+    target->copyAttrib(Attrib::COLOR, 3, 0, value_ptr(*colors.data()), colors.size());
+    target->copyAttrib(Attrib::TEX_COORD_0, 2, 0, value_ptr(*texcoords.data()), texcoords.size());
+
+    target->copyIndices(Primitive::TRIANGLES, indices.data(), indices.size(), 1);
+}
+
+
+///////////////////////////////////////////////////////////////////////////////////////
+// Dodecahedron
+#undef PHI	// take the reciprocal of phi, to obtain an icosahedron that fits a unit cube
```
paulhoux: Change icosahedron into dodecahedron :)

CCortex:
Correction done.
For the uv mapping, we'll just leave it like this for now until the separation is made then.

richardeakin:
Separating UV-mapping from `geom::Source`s?  One of their main utilities is providing texcoords, so I don't see why we'd ever want to remove that functionality. We could possibly think of a way to offer more flexible mappings, actually I think this is already possible with a `geom::Modifier` but perhaps we could add something built in for primitives that could use more complex mappings.

richardeakin:
The implementations look good to me, although I'm seeing many white space irregularities. Please use spaces within brackets (like [here](https://github.com/cinder/Cinder/pull/1192/files#diff-de816c4ab58b2e2285213136e2967fbdR1498)) and around math operations (like [here](https://github.com/cinder/Cinder/pull/1192/files#diff-de816c4ab58b2e2285213136e2967fbdR1401) and most calls within that function). Of course there is the more complete style guide: [CONTRIBUTING.md](https://github.com/cinder/Cinder/blob/master/CONTRIBUTING.md).

paulhoux:
To explain a bit more what I mean with separating UV-mapping: we've discussed having a ```geom::Mapping``` class of sorts, providing mappers for spherical, cylindrical, cube and planar mapping, to name just a few. The primitives themselves would not have code in them to provide texcoords, but by applying one of the mappers you could easily add them. As you mentioned, ```geom::Mapping``` would probably extend ```geom::Modifier```.

paulhoux:
I think the code inside the ```geom``` namespace will likely change in the future, as I can see a lot of duplicated functionality in there. For instance, a lot of primitives can be constructed by extruding a 2D shape (e.g. a circle) along a line (to create a cylinder) or a curve (to create a torus). Having a few basic functions and using them from with the respective classes would reduce the code tremendously while making it easier to add primitives later. I will not tackle this in the near future, though, too much stuff on my plate already.

paulhoux:
> For the uv mapping, we'll just leave it like this for now until the separation is made then.

I'd much prefer to have the new primitives support some form of mapping, otherwise they stand out as the only ones that don't. It would feel incomplete. @CCortex  Do you think you can add texcoords?

richardeakin:
I agree that any new primitives should support at least some sort of uv mapping, or else when you try to apply a texture to them (or just expect texcoords in the shader) it will appear broken. I should also say thanks for the contribution, having more geoms to play with is always nice.

(sorry, derailing on geom refactor below)

Re `geom::Mapper`: if you're talking using this as an implementation detail and a way to improve code organization, I can see how that can be an improvement. However if it means that from then on out, you need to remember an extra call in order to get basic tex coords (like say, on a Cube) that seems not as nice to me as having the texcoords built into the geom. To me, a geom::Source should always be able to provide some sort of tex coords as is, but maybe that could mean using a `Mapper` internally.

So I guess my concerns are about API - for example I would much rather have a Cylinder class than having to construct a cylinder by extruding a circle, even though you can do that of course if you wanted to. Likewise, I'd like that Cylinder to have a default texcoord mapping. Though I can see the gain by implementing these higher level geom classes using reusable components.

paulhoux:
Oh yes, I totally mean to keep a Cylinder class, but its internals could be rewritten to make use of more generic algorithms, instead of having very similar code for each of the primitives (the way it is now).

I agree that the geom::Mapper should be optional, with uv-coordinates being available by default. Not entirely sure how to code this, but we'll figure it out.

paulhoux:
Hi @CCortex ,

I just tried the new primitives in the Geometry sample, but there are a few issues:
 * The Octahedron and Dodecahedron have missing texture coordinates.
 * The WireDodecahedron should be clean, without triangulation.
 * The Dodecahedron does not fit inside a 1x1x1 cube, which is what we chose to be the default.
 * The Tetrahedron does not align well with the coordinate frame. This makes it hard to use the primitive in applications. Preferably, it should be aligned with the y-axis, where the y-axis goes through one vertex (the 'top') and through the middle of the opposite polygon.

These issues need to be fixed before we can accept the pull request.

Issue #1191, opened 2015-11-14
gl::Fbo and layered rendering (or 3d attachments)


simongeilfus:
Hey all,
There's a few small changes we should do to properly support that type of attachments in our ```Fbo``` class. At the moment trying to attach a ```GL_TEXTURE_3D``` will result in a ```GL_INVALID_ENUM``` error (and from the code, I would say the same will happen when trying to attach a ```GL_TEXTURE_1D```). This will unfortunately not be caught as a ```FboException``` and will fail silently if not checking for gl errors. Which is definitely bad!

Snippet that trigger the error:
```c++
{
	auto tex3d		= gl::Texture3d::create( 128, 128, 128, texFormat );
	auto fboFormat	= gl::Fbo::Format().attachment( GL_COLOR_ATTACHMENT0, tex3d );
	try {
		mFbo		= gl::Fbo::create( 128, 128, fboFormat );
	}
	catch( const ci::Exception &exc ) { CI_LOG_EXCEPTION( "Fbo init", exc ); }
	
	CI_CHECK_GL(); // GL_INVALID_ENUM
}
```

There's two approach for a fix. The first one is to check for the texture target when attaching textures:

```c++
void Fbo::attachAttachments()
{
	// attach Renderbuffers
	for( auto &bufferAttachment : mAttachmentsBuffer )
		glFramebufferRenderbuffer( GL_FRAMEBUFFER, bufferAttachment.first, GL_RENDERBUFFER, bufferAttachment.second->getId() );
	
	// attach Textures
	for( auto &textureAttachment : mAttachmentsTexture ) {
		auto textureTarget = textureAttachment.second->getTarget();
		if( textureTarget == GL_TEXTURE_1D ) {
			glFramebufferTexture1D( GL_FRAMEBUFFER, textureAttachment.first, textureTarget, textureAttachment.second->getId(), 0 );
		}
		else if( textureTarget == GL_TEXTURE_2D ) {
			glFramebufferTexture2D( GL_FRAMEBUFFER, textureAttachment.first, textureTarget, textureAttachment.second->getId(), 0 );
		}
		else if( textureTarget == GL_TEXTURE_CUBE_MAP ) {
			textureTarget = GL_TEXTURE_CUBE_MAP_POSITIVE_X;
			glFramebufferTexture2D( GL_FRAMEBUFFER, textureAttachment.first, textureTarget, textureAttachment.second->getId(), 0 );
		}
[...]

PR #1193, opened 2015-11-16
Updated Fbo::attachAttachments to modern opengl on desktop

simongeilfus:
And also added an extra case in the checkStatus method to deal with incomplete layered fbo.

See discussion https://github.com/cinder/Cinder/issues/1191

### DIFF
--- a/src/cinder/gl/Fbo.cpp
+++ b/src/cinder/gl/Fbo.cpp
@@ -358,10 +358,14 @@ void Fbo::attachAttachments()
 	
 	// attach Textures
 	for( auto &textureAttachment : mAttachmentsTexture ) {
+#if ! defined( CINDER_GL_ES )
+		glFramebufferTexture( GL_FRAMEBUFFER, textureAttachment.first, textureAttachment.second->getId(), 0 );
+#else
 		auto textureTarget = textureAttachment.second->getTarget();
 		if( textureTarget == GL_TEXTURE_CUBE_MAP )
 			textureTarget = GL_TEXTURE_CUBE_MAP_POSITIVE_X;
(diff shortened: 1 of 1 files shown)


simongeilfus:
Hey all,
Any opinion or strong feelings against this PR? :)

simongeilfus:
I might add a bit of info as the discussion I linked to is a really specific case, and is clearly confusing about the idea of this PR.

This PR is basically about the fact that our ```gl::Fbo``` class can be used for any type of texture; ```1D, 2D, 3D, Cubemaps, Arrays```, etc... The issue we have though is that it doesn't make any sense to call ```glFramebufferTexture2D``` when attaching anything other than a ```2D``` texture. So when it's available (GL 3.2), why not use the generic version of this function ```glFramebufferTexture```?

I believe that this will also correct a few issues that people have been reporting with ```gl::FboCubeMap```.

So in a few words this PR is "Let's not try to make everything 2D, where we can have *D"; especially when it's a one-liner.

I tested most samples that use a ```gl::Fbo``` or a ```gl::FboCubeMap``` on OS X and everything works fine here!

andrewfb:
OK - yes, this looks good to me. Merged. I'd still love for someone to tackle layered rendering but this is at least part of the picture.

simongeilfus:
Thanks Andrew,
Do you mean having a sample showing layered rendering? I'm working on a cascaded shadow map sample at the moment, this might be a good candidate as it's a broadly used technique. Or maybe simply another shadow mapping sample with multiple shadows, but this might be a bit more redundant. Both are a good way to show how to use a ```gl::Texture3d/GL_TEXTURE_2D_ARRAY``` with a ```gl::Fbo```, and imo would make more sense than with a ```gl::TextureCubeMap```.

Issue #827, opened 2015-04-17
Fbo / Texture should provide a getFormat() method


richardeakin:
As discussed in #813, we don't want to store the Format object on these classes because this will be diffuclt to keep in sync with after-construction changes to an Fbo or Texture class. So instead, these classes should recreate a compatible Format and return that.

PR #1217, opened 2015-12-06
Texture*::getFormat() const

paulhoux:
Constructs a Texture*::Format based on the actual OpenGL state. In response to issue #827. The `asInt` and `asFloat` utility functions were needed to convert the border color values, but I have decided against auto-conversion. The functions are nice additions by themselves, though. See also: http://stackoverflow.com/questions/20762952/most-efficient-standard-compliant-way-of-reinterpreting-int-as-float

### DIFF
--- a/src/cinder/gl/Texture.cpp
+++ b/src/cinder/gl/Texture.cpp
@@ -31,7 +31,6 @@
 #include "cinder/gl/scoped.h"
 #include "cinder/ip/Flip.h"
 #include "cinder/Log.h"
-#include "cinder/Utilities.h"
 #include <stdio.h>
 #include <algorithm>
 #include <memory>
--- a/src/cinder/gl/Texture.cpp
+++ b/src/cinder/gl/Texture.cpp
@@ -697,7 +697,8 @@ TextureBase::Format::Format( GLenum target, GLuint textureId )
 	mMipmappingSpecified = true;
 	glGetTexParameterIuiv( mTarget, GL_TEXTURE_BASE_LEVEL, &mBaseMipmapLevel );
 	glGetTexParameterIiv( mTarget, GL_TEXTURE_MAX_LEVEL, (GLint*)&mMaxMipmapLevel );
-	if( mMaxMipmapLevel == requiredMipLevels( w, h, d ) - 1 ) mMaxMipmapLevel = -1;
+	if( mMaxMipmapLevel == requiredMipLevels( w, h, d ) - 1 )
+		mMaxMipmapLevel = -1;
 	glGetTexParameterfv( mTarget, GL_TEXTURE_BORDER_COLOR, (GLfloat*)mBorderColor.data() );
 	if( mBorderColor[0] < 0.0f || mBorderColor[0] > 1.0f || ! isnormal( mBorderColor[0] ) ||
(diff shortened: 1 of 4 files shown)


src/cinder/gl/Texture.cpp
```diff
@@ -666,7 +667,55 @@ TextureBase::Format::Format()
...
+	glGetTexParameterIiv( mTarget, GL_TEXTURE_MAG_FILTER, (GLint*)&mMagFilter );
+	mMinFilterSpecified = ( mMinFilter != GL_LINEAR || mMagFilter != GL_LINEAR );
+	mMipmapping = ( mMinFilter == GL_LINEAR_MIPMAP_LINEAR || 
+					mMinFilter == GL_LINEAR_MIPMAP_NEAREST || 
+					mMinFilter == GL_NEAREST_MIPMAP_LINEAR || 
+					mMinFilter == GL_NEAREST_MIPMAP_NEAREST );
+	mMipmappingSpecified = true;
+	glGetTexParameterIuiv( mTarget, GL_TEXTURE_BASE_LEVEL, &mBaseMipmapLevel );
+	glGetTexParameterIiv( mTarget, GL_TEXTURE_MAX_LEVEL, (GLint*)&mMaxMipmapLevel );
+	if( mMaxMipmapLevel == requiredMipLevels( w, h, d ) - 1 ) mMaxMipmapLevel = -1;
```
paulhoux: According to the style guide, this should be on a separate line.

richardeakin:
Interesting approach. I'm wondering though, as I think you want this so that `gl::Texture::Format` can be used as a key into a texture cache, would doing all these gl property queries be expensive? Seems like if that were the case, we could copy the variables over to read-only, private ivars on gl::Texture, which you could then use to reconstruct a new gl::Texture::Format. I know this seems sort of pointless rather than the initial version, which just stored an mFormat object on the gl::Texture, but it still seems like possibly better than issuing a bunch of `glGetTexParameterIui*()` calls at what might possibly be draw time.

FWIW this was we resolved to do during the app refactor, when we wanted to avoid storing the App::Settings in the impls as it became confusing. You can see it in the mac impl [here](https://github.com/cinder/Cinder/blob/767fefe0ae1102323185865a951e5ffb17759588/src/cinder/app/cocoa/AppImplMac.mm#L85-L101).

paulhoux:
Well, we used to have the format stored as a member variable, but it was decided against that to reduce the memory footprint of the texture. Querying the format from the current state is not *very* expensive, but of course much more expensive than keeping the format around in one way or the other. I personally would not choose to reconstruct the format of a texture more than once. But it does come in handy if you want to duplicate a texture, even a texture that you did not create yourself. With the code in this PR, all you need is a texture name (id) to reconstruct the format. I wrote this code as per Andrew's instructions, although he hasn't had a chance to review it yet.

richardeakin:
Okay, makes sense, especially if he thinks performance isn't a problem in this case.

The rationale (for me anyway) behind not storing mFormat ivars is that we were ending up with the same variables in two places of the class, and this became confusing over time, creating bugs.

paulhoux:
Yeah, now that I think of it, it wasn't so much about memory footprint as it was about keeping the stored format up-to-date.

PR #1224, opened 2015-12-11
Fixed issue with FboCubeMap from previous PR

simongeilfus:
***Really sorry about this!***

My last changes have broken our dear FboCubeMap class! I'm not sure how, but it seems like I missed this in the tests I've done for my last PR.

I'll start working on a better solution, but for now this fix the issue.

What was happening is that using ```glFrameBufferTexture``` with a cubemap texture (instead of ```glFrameBufferTexture2D```) makes the Framebuffer expect a depth attachement of the same target (and therefore a cubemap depth attachement / 6 faces depth attachement). Ideally this should be dealt with when creating the depth buffer, checking for the format there, but this involves other changes and it's probably better to make that happen with a few other changes that we recently discussed.

Sorry again!

### DIFF
--- a/src/cinder/gl/Fbo.cpp
+++ b/src/cinder/gl/Fbo.cpp
@@ -358,10 +358,16 @@ void Fbo::attachAttachments()
 	
 	// attach Textures
 	for( auto &textureAttachment : mAttachmentsTexture ) {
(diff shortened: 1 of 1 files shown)


andrewfb:
Thanks for this fix

PR #1234, opened 2015-12-13
RectT::moveULTo(), getMoveULTo()

richardeakin:
For parity with Area, which has the same methods.

### DIFF
--- a/include/cinder/Rect.h
+++ b/include/cinder/Rect.h
@@ -56,12 +56,14 @@ class RectT {
 	void		clipBy( const RectT &clip );
 	RectT		getClipBy( const RectT &clip ) const;
 	Area		getInteriorArea() const;
+	//! Translates the RectT by \a off
 	void		offset( const Vec2T &offset );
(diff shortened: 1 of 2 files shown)


paulhoux:
These look good to me. Is there any reason why you moved the implementation to the cpp file? This prevents inlining.

richardeakin:
Trying to clean up the header so it is more readable. The compiler still optimizes non-virtual methods even when they are in a translation unit, the big performance impact is virtual methods with more than one overload. Also I haven't looked at the assembly but it probably isn't inlining anyway since RectT::set() is implemented in the cpp (as it should be IMO).

Honestly I think we are inlining way to much in these classes, which causes code bloat since the assembly gets moved to each file that is including the file with inlined methods. Explained [here](https://isocpp.org/wiki/faq/inline-functions#inline-and-perf).

Aside from that, I don't think it would be possible to create a case where RectT was ever a performance bottleneck and something like inlined methods saved the day, even if you tried.

paulhoux:
Fair enough, good read.

PR #1237, opened 2015-12-15
vc2013 build files for unit tests

SethGibson:
Adding Visual Studio 2013 build files for unittests per Rich's request and making (hopefully) minor changes to existing tests for cross-platform and vs friendliness

### DIFF
--- a/test/unit/src/signals/SignalsTest.cpp
+++ b/test/unit/src/signals/SignalsTest.cpp
@@ -102,8 +102,9 @@ TEST_CASE( "signals/Signals" )
 	{
 		class MemberClass {
 		public:
+			MemberClass() { mAccum = 0; }
 			void handler()	{ mAccum++;	}
-			int mAccum = 0;
+			int mAccum;
 		};
 
 		MemberClass instance;
--- a/test/unit/src/signals/SignalsTest.cpp
+++ b/test/unit/src/signals/SignalsTest.cpp
@@ -102,7 +102,7 @@ TEST_CASE( "signals/Signals" )
 	{
 		class MemberClass {
 		public:
-			MemberClass() { mAccum = 0; }
+			MemberClass(){ mAccum = 0; }
 			void handler()	{ mAccum++;	}
 			int mAccum;
 		};
--- a/test/unit/.gitignore
+++ b/test/unit/.gitignore
@@ -2,6 +2,7 @@ CinderUnitTests
 vc2013/[Dd]ebug*/
 vc2013/[Rr]elease*/
 vc2013/[Xx]64/
+vc2013/[Dd]ata/
 *.sdf
 *.opensdf
 *.suo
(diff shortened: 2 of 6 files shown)


richardeakin:
Thanks! A quick note that we discussed depositing the build products into a folder at test/unit/vc2013/build or something, as this will automatically get ignored by the root .gitignore file. Also will allow you to build multiple configurations / targets that can live side by side.

test/unit/src/signals/SignalsTest.cpp
```diff
@@ -102,8 +102,9 @@ TEST_CASE( "signals/Signals" )
 	{
 		class MemberClass {
 		public:
+			MemberClass() { mAccum = 0; }
 			void handler()	{ mAccum++;	}
-			int mAccum = 0;
```
sansumbrella: Ah, bummer there aren't default initializers in VS yet.

SethGibson: Sadly, it's not even a VS issue, it's specific to VS 2013. The original definition with the default initializer builds just fine in 2015 :(

richardeakin: I believe it's called ['Non-static data member initializers'](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2756.htm) and according to [this](https://msdn.microsoft.com/en-us/library/hh567368.aspx), they are supported in VS2013. But maybe it is because `MemberClass` is declared locally and VS2013 just has a bug related to that?

Anywho, change doesn't effect the test much.

SethGibson: My assumption was that it had something to do with [this](http://connect.microsoft.com/VisualStudio/feedbackdetail/view/896703/a-fatal-error-c1506-unrecoverable-block-scoping-error-in-c-compiler-is-reported-for-class-with-default-member-values-definitions-inside-a-nested-block), as I this is the error I was getting at compile time. if i moved the class def outside of the SECTION, it works just fine also.

SethGibson:
Alright, small RFC:unit test paths for VS, mainly aimed at @richardeakin  and @sansumbrella  , but would love to hear from anyone else. Here's what I'm proposing after yesterday's discussion: $(ProjectDir)build\$(PlatformToolset)\$(Configuration)\$(PlatformTarget) and $(ProjectDir)obj\$(PlatformToolset)\$(Configuration)\$(PlatformTarget) for intermediates, so for example:

C:\lib\Cinder\tests\unit\vc2013\build\v120\Debug\x86

Thoughts, Comments, Diatribes?

richardeakin:
That's about standard, only thing I'd propose differently is that the obj folder live at `$(ProjectDir)\build\obj\`. [We ignore everything inside a build folder](https://github.com/cinder/Cinder/blob/767fefe0ae1102323185865a951e5ffb17759588/.gitignore#L11), so we wouldn't need any more .gitignore entries that way.

SethGibson:
Once more, this time with updated output and intermediate directories.

PR #1240, opened 2015-12-17
TUIO samples: namespace update for 0.9.0

musiko:


andrewfb:
Thanks for submitting a PR, however we're *just* about to replace the TUIO implementation with this nearly complete PR, #1179. Also as a quick note, gl::color() is probably the correct thing here; it's a function that sets the current color. ci::Color is the class that represents colors. So the code as changed just creates a temporary variable that then evaporates. In general Cinder's functions are lowercase, and classes are uppercase, to distinguish in cases like this.

richardeakin:
Closing this as #1179 was merged.

PR #1241, opened 2015-12-17
Added a GlslProg::create() method for transform feedback shaders.

paulhoux:
This allows you to quickly create the transform feedback shader by just supplying a file, a list of varyings and optionally a transform feedback format. Attributes don't need to be specified, because GlslProg already is able to query them from the source files. Thanks to this addition, I was able to add support for transform feedback shaders to a caching system, which would otherwise have been harder.

### DIFF
--- a/include/cinder/gl/GlslProg.h
+++ b/include/cinder/gl/GlslProg.h
@@ -351,6 +351,9 @@ class GlslProg {
 	static GlslProgRef create( DataSourceRef vertexShader, DataSourceRef fragmentShader = DataSourceRef() );
 	static GlslProgRef create( const std::string &vertexShader, const std::string &fragmentShader = std::string() );
 #endif
+#if defined( CINDER_GL_HAS_TRANSFORM_FEEDBACK )
+	static GlslProgRef create( DataSourceRef vertexShader, const std::vector<std::string>& varyings, GLenum format = GL_INTERLEAVED_ATTRIBS );
+#endif
 	~GlslProg();
 	
 	void			bind() const;
--- a/src/cinder/gl/GlslProg.cpp
+++ b/src/cinder/gl/GlslProg.cpp
@@ -433,6 +433,12 @@ GlslProgRef GlslProg::create( const string &vertexShader, const string &fragment
 }
 	
 #endif
+#if defined( CINDER_GL_HAS_TRANSFORM_FEEDBACK )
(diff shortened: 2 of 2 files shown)


richardeakin:
Sounds reasonable to add TF-oriented constructor to me, if nothing else than to make it clear to first time users of the technique what is needed. Maybe at least one of the samples should make use of this new create() method?

I wish we'd document the constructors, especially when they have  arguments with defaults because some lesser IDEs (cough, cough, Xcode) tend to not even show those variables during auto-completion, but at the same time I realize the other ones aren't documented, either.

paulhoux:
Yeah, it's a good idea to use it in a sample. Will add it to this PR tomorrow.

ryanbartley:
I agree. I also think there should be a constructor that takes a geometry shader as well.

richardeakin:
@ryanbartley well that one becomes a slippery slope - then do we need a constructor for tesselation shaders too?  compute? TF is sort of unique because it is commonly its own standalone vert shader.

richardeakin:
So, there actually are `GlslProg::create()` methods that can take geometry shaders, I hadn't realized it because they default to empty.

It also got me thinking, what with the changes @paulhoux is proposing here, we could remove the default-empty fragment shader in [these constructors](https://github.com/cinder/Cinder/blob/dd16254f0f4ab2276df845f45b604355e64299f2/include/cinder/gl/GlslProg.h#L340-L352), since AFAIK the only time you would use a vert-shader-only GlslProg would be in the case of transform feedback.

cinder:
My own opinion is that for just about every case that's not just vert+frag, `GlslProg::Format` should be used. In my view, tessellation shaders aren't such a common use case that the ratio of saved typing to complication of the class makes sense here, though it seems others feel differently?

paulhoux:
@andrewfb : there is a slight issue with ```gl::GlslProg::Format``` that is important in this discussion, and that is that all parsing is done during construction of the format. This of course includes file access. In those cases where you'd like to delay this, for example if you want to do this asynchronously, you need to have a way to specify the source files, as well as all other parameters (e.g. attributes), without using the format. In my opinion, this slightly defeats the purpose of the format as a descriptor of your shader.

richardeakin:
@paulhoux you can load the glsl into std::string's async and then construct the entire format at once. But yea, I don't see why you couldn't create the entire Format on a background thread and then copy (move if you prefer) to the main thread to load. It's actually nice that the file i/o is happening at the Format for this reason come to think of it, because it allows you to at least handle that part async if you wish. Tempting to move the shader preprocessing there as well..

PR #1242, opened 2015-12-20
Moving audio unit tests to test/unit/src/audio /w CATCH

richardeakin:
I'm not too familiar yet with CATCH so please correct if I switched to using inappropriate macros (really just using `REQUIRE()` for everything).

Still need to update the vc2013 project file, will do that next time I'm in front of a windows machine.


richardeakin:
So I've added the audio unit tests to the VS2013 project and noticed a couple worrying things.

For one, the Fft unit test is failing for FFT sizes = 4 (not very useful except when testing). I created a separate issue for that (#1263), will investigate when I have time but for now just disabled it on windows.

Other thing is, I noticed that the files in data folder are copied to the .exe's folder so they can be found, which causes them to show up in VCS as modified. I propose we just rename data folder to assets and use Platform::loadAsset() as that works in headless situations like this.

Lastly, @sansumbrella I have [this test](https://github.com/richardeakin/Cinder/blob/unit_test_audio/test/unit/src/audio/FftUnit.cpp) that I'm running in a for loop, is there a more informative way to do it with CATCH? Ideally, I'd like to know what parameters were passed into computeRoundTrip() when something fails, whether that means making it its own section or not, I don't know.

sansumbrella:
I'm down to use `Platform::loadAsset()` in the other tests. Didn't realize that existed.

For adding more information on failure, I'll poke around at the CATCH docs. In the meantime, you can have it break on failure by passing a flag to the application (`-b` or `--break`).

sansumbrella:
For adding contextual information, you can use Catch's [logging macros](https://github.com/philsquared/Catch/blob/master/docs/logging.md).

For example, the INFO log below will report when the second check fails (check is similar to require, but allows tests to continue on failure). INFO is block scoped, so it's good to use in the `computeRoundTrip` function.
```c++
  SECTION("Check logging functionality")
  {
    INFO("Something, something… " << 5 << ", " << 10);
    CHECK(true);
    CHECK(false);
  }
```

test/unit/src/audio/FftUnit.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include <iostream>
+
+using namespace ci::audio;
+
+namespace {
+
+void computeRoundTrip( size_t sizeFft )
+{
+	dsp::Fft fft( sizeFft );
```
sansumbrella: Adding `INFO( "FFT size: " << sizeFft );` at the top will print the requested info when any of the assertions in the function block fail.

sansumbrella: Or just `CAPTURE( sizeFft );` and it will print `sizeFft := 4` if an assertion fails.

sansumbrella:
I made a [PR](https://github.com/cinder/Cinder/pull/1277) that changes the data loading to asset loading in the unit tests. No windows machine at work, so I haven't touched the visual studio side.

richardeakin:
Thanks - updated the windows side and merged. I'll look at improving the FFT test when I have some time to figure out why its failing on windows when size = 4, I appreciate the tip..

PR #1247, opened 2015-12-22
Disable search box's autocomplete

### LINKED ISSUE #1246: Disable search box's autocomplete
Currently the browser's autocomplete and the search autocomplete both fire and it gets confusing:
screen shot 2015-12-21 at 9 11 09 pm

I think we should disable the browser autocomplete so you can focus on the search results.

Pulled this out of #1243 since it's the remaining actionable item.

drewish:
Fixes #1246.

http://www.w3.org/wiki/HTML/Elements/input/text has some documentation for the attribute.


richardeakin:
Thanks!

Issue #1265, opened 2016-01-07
Add log::makeOrGet<LoggerT>()


richardeakin:
To grab a LoggerT if it exists, or if not make a new one and return that. Cuts down on boilerplate logic in some cases.

lucasvickers:
Will implement shortly.

richardeakin:
Implemented in #1266.

PR #1266, opened 2016-01-08
added makeOrGet helper function

lucasvickers:
Part of me wants to call it ```makeOrGetLogger``` but I acknowledge that it's a little long.

This is in reference to #1265

### DIFF
--- a/include/cinder/Log.h
+++ b/include/cinder/Log.h
@@ -218,6 +218,9 @@ public:
 	template<typename LoggerT, typename... Args>
 	std::shared_ptr<LoggerT> makeLogger( Args&&... args );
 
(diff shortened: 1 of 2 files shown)


richardeakin:
`makeOrGetLogger()` totally makes sense as it matches `makeLogger()`, my bad.  I don't think it is too long, much shorter than what you have to write without it.

lucasvickers:
ready to merge

richardeakin:
Tested and works great, the following is pretty nice:

```cpp
log::makeOrGetLogger<log::LoggerBreakpoint>()->setTriggerLevel( (log::Level)breakOnLogLevel );
```

Thanks!

PR #1289, opened 2016-02-03
Added fast aabb-sphere intersection test.

paulhoux:
I needed this for an upcoming project and noticed it wasn't in Cinder yet. Based in part on: 
http://www.realtimerendering.com/resources/GraphicsGems/gems/BoxSphere.c


richardeakin:
Really nice addition! 

Not saying we need to now, but I think functionality like this would benefit from unit testing, esp. since we recently added the catch framework and a formalized way to add tests. Maybe something for the future.

PR #1309, opened 2016-02-07
Set mIsCached to false when the cache is empty

### LINKED ISSUE #1308: OSC: nullptr exception when using the copy constructor for received messages
In the new OSC block the copy constructor gives a nullptr exception. It seems to be because `message.mCache` is undefined and `ByteBuffer` calls `size` which requires a reference to `this` (which it doesn't have). 

By changing [this line](https://github.com/cinder/Cinder/blob/master/blocks/OSC/src/Osc.cpp#L113) to `mCache( nullptr )` the problem is fixed, but `mCache` should probably be initialized if `mIsCached` is true.

This happens on Windows, I have not tested on other platforms.

ulyssesp:
Fixes #1308 

Tested on Windows.

### DIFF
--- a/blocks/OSC/src/Osc.cpp
+++ b/blocks/OSC/src/Osc.cpp
@@ -76,6 +76,11 @@ inline double ntohd( int64_t x ) { return (double) ntohll( x ); }
 	
 ////////////////////////////////////////////////////////////////////////////////////////
 //// MESSAGE
+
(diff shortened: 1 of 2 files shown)


ryanbartley:
This solves the problem. Tested.

andrewfb:
Merged - thanks for the find and the fix.

PR #1311, opened 2016-02-12
ignore the .VC.opendb files created by VS 2015

richardeakin:
They were showing up as untracked content whenever I ran a sample.


ryanbartley:
+1

PR #1318, opened 2016-02-17
Implement MSAA hint and setTitle for glfw window.

PetrosKataras:


richardeakin:
Thanks, sorry for the delay.

PR #1321, opened 2016-02-20
Fix Linux version of getDefaultExecutablePath to return parent dir path of the exe.

PetrosKataras:
As is [currently](https://github.com/cinder/Cinder/blob/android_linux/src/cinder/app/linux/PlatformLinux.cpp#L409) `getDefaultExecutablePath` will return the full symlink path to the exe including the exe itself. This fixes it by returning the parent dir of the exe as it happens on the other platforms too.

### DIFF
--- a/src/cinder/app/linux/PlatformLinux.cpp
+++ b/src/cinder/app/linux/PlatformLinux.cpp
@@ -406,7 +406,7 @@ fs::path PlatformLinux::getDefaultExecutablePath() const
     if( ( -1 != len ) && ( len < buf.size() ) ) {
       buf[len] = '\0';
     }
- 	return fs::path( std::string( &(buf[0]), len ) );
+ 	return fs::path( std::string( &(buf[0]), len ) ).parent_path();
 }
 
 void PlatformLinux::sleep( float milliseconds ) 


richardeakin:
I think this means that currently on linux `getDefaultExecutablePath()` returns a path including the actual executable, instead of the parent folder? It is true that on other platforms, the parent folder is returned. For example calling `app::getAppPath()` (which ends up using `getDefaultExecutablePath()` unless you've already changed it with `Platform::setExecutablePath()`) returns the following on:

windows debug mode: "D:\code\cinder\cinder-master\samples\BasicApp\vc2013\Debug\"
OS X debug mode: "/Users/r/code/cinder/cinder-master/samples/BasicApp/xcode/build/Debug"

It does bring up the debate on whether this function could use a clearer name, such as `app::getAppDirectory()` / `Platform::getDefaultExecutableDirectory()`. However if we were to then implement `getAppPath()` / `Platform::getDefaultExecutable()` to return the path including actual file, it would break some external code. Alternatively we can improve the documentation to clarify that the function returns a path to the folder in which the executable lives.

PetrosKataras:
Yes, exactly. I m pretty surprised actually that it didn't come up until now since any code that tries to use `getAppPath()` on linux right now as a directory path will fail. I agree that the naming is misleading in this case and I find `getAppDirectory` / `getDefaultExecutableDirectory` much better suited for the usage intention but as you mention this would potentially break external code.

chaoticbob:
Finally getting some time to follow up on some of these...did we ever get a consensus on this? Would be great if we could remove the ambiguity.

richardeakin:
I think the decision is to accept this change, and look to do a breaking rename for clarification in a later release. We can open a new issue for that.

Issue #1323, opened 2016-02-22
App on OS X 10.11 will slow update loop when window is obscured from view


lucasvickers:
I am curious if anyone knows the cause, which is presumably an OS X feature.  Would also be useful to see if this can be turned off via code.

I am OS X 10.11.1
Cinder master branch

```
void testApp::update()
{
	static std::chrono::time_point<std::chrono::system_clock> last;
	
	time_point<system_clock> now = system_clock::now();
	system_clock::duration delta = duration_cast<milliseconds>( now - last );

	if( duration_cast<milliseconds>(delta).count() > 50 ) {
		CI_LOG_D( "Update took: " << duration_cast<milliseconds>(delta).count() << " milliseconds" );
	}
	
	last = chrono::system_clock::now();
}
```

When the window containing the above code is obscured, I notice unusually large intervals such as:
```
|debug  | virtual void testApp::update()[38] Update took: 10019 milliseconds
|debug  | virtual void testApp::update()[38] Update took: 6896 milliseconds
|debug  | virtual void testApp::update()[38] Update took: 5103 milliseconds
|debug  | virtual void testApp::update()[38] Update took: 10031 milliseconds
|debug  | virtual void testApp::update()[38] Update took: 430 milliseconds
|debug  | virtual void testApp::update()[38] Update took: 1671 milliseconds
```

andrewfb:
A first thing to try might be to disable App Nap:
http://www.cultofmac.com/253064/how-to-turn-app-nap-off-for-specific-apps-on-your-mac-os-x-tips/

kitschpatrol:
Hi Lucas, this is almost certainly App Nap. Cinder has built-in functionality to turn disable it on a per-app basis. Dropping this in the `prepareSettings` function was a fix for me:

`	settings->setPowerManagementEnabled(true); // Prevent App Nap`

lucasvickers:
@andrewfb unfortunately OS X seems to have removed that feature, or at least further obscured it (see image).
@kitschpatrol that seemed work, thank you!

![screen shot 2016-02-21 at 9 36 22 pm](https://cloud.githubusercontent.com/assets/1099950/13208134/eeb47a7a-d8e3-11e5-8760-0675d55dc96d.png)

sansumbrella:
Just to clarify, the power management flag should be set to `false` to disable app nap, correct? If its the opposite, then we need to update the Cinder docs (and probably change the function name, or reverse the behavior).

[...]

PR #1339, opened 2016-02-25
Fixing OS X sleep-prevention code to address #1323

andrewfb:


sansumbrella:
Tested as working on OSX 10.11.3.

`settings->setPowerManagementEnabled(false);` => no screensaver
`settings->setPowerManagementEnabled(true);` => screensaver kicks in

Issue #1343, opened 2016-02-27
Fbo issues on the RPi2


PetrosKataras:
I had a look on the Fbo implementation on the RPi2 based on issues described on [this](https://forum.libcinder.org/#Topic/23286000002626005) forum post.

It seems there are a couple of things that need to be changed in order to make Fbo's work properly on the RPi2.

First thing, is that on the RPi2 there is no support for depth textures ( actually there is a `OES_depth_texture` extension that is available on some devices but this doesn't seem to be the case on the RPi2  )  and this is currently not handled which means that codes that tries to create a depth texture on the RPi2 will just fail. 

Secondly there is no `GL_DRAW_FRAMEBUFFER` and `GL_READ_FRAMEBUFFER` on the RPi2 . The only valid target is `GL_FRAMEBUFFER` which means that `pushFramebuffer` / `popFramebuffer` need to be updated in order to reflect this.

By taking into consideration these things I am able to properly run the FboBasic sample.

chaoticbob:
Got a PR? ;)

PetrosKataras:
@chaoticbob Sure, just not entirely sure what would be the best way to handle the missing depth texture. Should it throw a warning and not create a depth attachment ?? This would allow the app to run but you won't get any depth functionality or should we let it fail and throw a warning to use the default RenderBuffers instead on this platform?

For the `GL_FRAMEBUFFER` target issue I think its enough if we update [this block here](https://github.com/cinder/Cinder/blob/android_linux/include/cinder/gl/platform.h#L134-L179) and disable `CINDER_GL_HAS_FBO_MULTISAMPLING`. Looking at the specific block I think also `CINDER_GL_HAS_DRAW_INSTANCED` should be removed at least for the RPi2 (  I *think* for gles 2.0 at least these two are only available through extensions on *some* devices ).

chaoticbob:
I think the approach should be (for now) to allow the creation of the FBO w/o the depth texture. Looks like it's a known issue that you just can't have a depth texture on the RPI2 - based on a searching and reading. It's not the optimal case and we do need to document it. But there are also use cases where having a color buffer by itself is enough. Tough call.

The RPI2 is the only target we have that has a strict preprocessor macro more or less locked to. So I think branching on the #ifdef a RPI2 block and then an everything else block is fine. What I wouldn't do is try to work it into the rest of the OpenGL ES 2.0 route. That would be a really good way to create a bunch of really hard to track down bugs.

[...]

PR #1354, opened 2016-03-01
Linux RPi2 GL fixes.

PetrosKataras:
This fixes the issue reported on [this](https://forum.libcinder.org/#Topic/23286000002634083) forum thread and includes a fix for the Fbo issue reported on [this](https://forum.libcinder.org/#Topic/23286000002626005) forum thread and discussed here #1343.


andrewfb:
Thanks for this fix

PR #1355, opened 2016-03-01
Dynamic lib target for Cinder on OSX

sansumbrella:
Adds a .dylib target to the Xcode project, following instructions from @simongeilfus.
Useful for those rare circumstances when you _need_ a .dylib.

### DIFF
--- a/xcode/cinder.xcodeproj/project.pbxproj
+++ b/xcode/cinder.xcodeproj/project.pbxproj
@@ -5064,7 +5064,7 @@
 					../lib/macosx/libboost_system.a,
 				);
 				PRELINK_LIBS = "../lib/macosx/libboost_filesystem.a ../lib/macosx/libz.a ../lib/macosx/libboost_system.a";
-				PRODUCT_NAME = "$(TARGET_NAME)_d";
+				PRODUCT_NAME = cinder_d;
 			};
 			name = Debug;
 		};
@@ -5097,7 +5097,7 @@
 					../lib/macosx/libboost_system.a,
 				);
 				PRELINK_LIBS = "../lib/macosx/libboost_filesystem.a ../lib/macosx/libz.a ../lib/macosx/libboost_system.a";
-				PRODUCT_NAME = "$(TARGET_NAME)";
+				PRODUCT_NAME = cinder;
 			};
 			name = Release;
 		};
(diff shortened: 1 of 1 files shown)


richardeakin:
Nice. Does this solve the problem where the full path is hardcoded into the dylib file? For example when I tried to use Simon's libcinder.dylib file on my macbook it was trying to find it at some hardcoded path on his machine.

richardeakin:
Any reason why we need the extra _dynamic suffix?  I understand the target name needs to be unique in the xcode project but it seems that `libcinder.dylib` and `libcinder_d.dylib` would work fine in distinguishing this from the static `.a` versions.

simongeilfus:
Hey, nice! Thanks for starting that up @sansumbrella! 
We can definitely get rid of the _dynamic suffix, it's unnecessary long!
About the issue your mentioning @richardeakin, I believe this is something you would address in the app settings and not in cinder. I think the solution is to add a copy script that copy the .dlyb to the app and then run "install_name_tool -change" on the moved dlyb. If I'm correct, this would only be needed when shipping an app with libcinder.dylib or when using a libcinder.dylib that was not build on the same machine.

richardeakin:
I think you can make a relative install name though, such as [described here](https://developer.apple.com/library/mac/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/RunpathDependentLibraries.html) although that seems a bit outdated as it refers to gcc.  I'm wondering if these are the applicable settings in xcode:

screenshot 2016-03-01 14 12 40

Some sort of of post install for the app does seem necessary as well (whether through bundle copy, or something more involved like building a Framework) for this to be usable in a manner where the app will run on a machine other than that on which it was built. In any event I don't think we want to be hardcoding full paths into the dylib.

vade:
Typically with dylibs you have to run an 'install_name_tool' post process shell script to get the relative install path for your application bundle to properly link the dylib. Ive had to do this on many an occasion, with a shell script something like: 

install_name_tool -change @rpath/libusb-1.0.0.dylib @loader_path/../Frameworks/libusb-1.0.0.dylib "$TARGET_BUILD_DIR/$PRODUCT_NAME.plugin/Contents/MacOS/$PRODUCT_NAME"

For example, which changes the libusb for a plugin's contents binary. I presume something like this would be necessary if I am following the discussion correctly.

richardeakin:
Hm, I was hoping some of those obscure xcode project settings under Linking would help here, but if not then I don't see why we couldn't add a post-build script (under Build Phases -> New Run Script Phase) that did something like what @vade suggests.

I think the best way to test this is to build your app, then copy the bundle to a different machine and try to run it. If the packaged libcinder.dylib can be found then that'll do it.

PR #1360, opened 2016-03-03
Osc and Tuio updates

ryanbartley:
- Fixed edge case in Tuio. In the spec, the source message is optional and in production this was breaking Tuio. Added an address to Osc Messages and use that address as a "source" in Tuio.
- Also implemented socket shutdown for Osc, which fixes a problem people were having with opening a socket soon after closing it. However, there are still edge cases, which can be seen [here](http://stackoverflow.com/questions/35365826/asio-address-already-in-use-on-bind-yet-i-can-still-connect/35394556#35394556).


ryanbartley:
3rd times the charm :)

richardeakin:
Looks very good to me. Haven't tested though, will merge unless there is any other feedback.

ryanbartley:
Thanks! Still very exciting to see a merge! :)

PR #1372, opened 2016-03-08
Linux fix DialogHelper(Zenity) newline ending.

PetrosKataras:
It seems that paths coming from Zenity contain a newline character at the end. This messes up filenames that then appear with this character ending. Use the ObjLoader sample as a test case to reproduce. The fix should not affect other DialogHelpers.


richardeakin:
Thanks, verified that this fixes a seg fault in ObjLoader sample.

PR #1381, opened 2016-03-12
Linux video player fix / change.

PetrosKataras:
Resetting `mIsDone` when manually restarting videos ( through `seekToStart` for example ) which was not handled and also disabling async loading of videos for now until we figure out what is the best way to handle loading so that we are in line with other platforms.

Note that loading will block now until the pipeline pre-rolls ( i.e until the video is fully loaded and all video info are available. ) which of course also means that its gonna be slower than the async version but since some blocks seem to rely on this behavior I think turning it off for now and thinking about it would be the path of least surprise for potential new users.

I am planning to come back to the GstPlayer for some cleaning up and for implementing missing functionality ( like stepping ) once we are ready to talk about the video players on the other platforms too.

### DIFF
--- a/src/cinder/linux/GstPlayer.cpp
+++ b/src/cinder/linux/GstPlayer.cpp
@@ -177,8 +177,10 @@ gboolean checkBusMessages( GstBus* bus, GstMessage* message, gpointer userPlayer
 				gst_message_parse_state_changed( message, &old, &current, &pending );
 				
 				data.mCurrentState = current;
-				
 				g_print( "Pipeline state changed from : %s to %s\n", gst_element_state_get_name( old ), gst_element_state_get_name ( current ) );
+
+				if( current > GST_STATE_NULL && current <= GST_STATE_PLAYING ) data.mIsDone = false;
+
 				switch ( current ) {
 					case GST_STATE_PAUSED: {
 						data.mIsPrerolled = true;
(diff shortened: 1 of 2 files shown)


richardeakin:
Interesting, async video loading?  This does indeed seem useful although a sync route seems good to have as well. I'd imagine a callback once the video does load would be appropriate (or possibly one for failure too). This will indeed take some thinking about how it compares to other platforms, although I don't think we're going to try to begin the cross-platform video api until after 0.9.1 release, which will be when Paul's new MF / MSW implementation lands. I think in the meantime it is good to keep the api's similar, but in the end there is no virtual interface yet, so they don't need to match yet. I think for now the most important thing is nailing the basic functionality on each platform.

src/cinder/linux/GstPlayer.cpp
```diff
@@ -177,8 +177,10 @@ gboolean checkBusMessages( GstBus* bus, GstMessage* message, gpointer userPlayer
 				gst_message_parse_state_changed( message, &old, &current, &pending );
 				
 				data.mCurrentState = current;
-				
 				g_print( "Pipeline state changed from : %s to %s\n", gst_element_state_get_name( old ), gst_element_state_get_name ( current ) );
+
+				if( current > GST_STATE_NULL && current <= GST_STATE_PLAYING ) data.mIsDone = false;
```
richardeakin: Code style nit: can you place the contents of the conditional on a new line? Details [here](https://github.com/cinder/Cinder/blob/master/CONTRIBUTING.md#spacing-and-line-breaks).

PetrosKataras:
Hey, I m gonna close this one since its pretty obsolete after the refactor work that has been going on here https://github.com/PetrosKataras/Cinder/tree/gst-player-refactor-and-osx-support. There should be a new PR coming in the following days with the updates.

PR #1390, opened 2016-03-17
CMake: cinderMakeApp now accepts include directories also.

PetrosKataras:


richardeakin:
Thanks and sorry for the delay on this..

PR #1399, opened 2016-03-25
Add vec2 overload of RectT::scaledCentered()

notlion:
Rectifies (har har) one case where the copying (past tense) methods don't match the in-place.


richardeakin:
Nice one, yep that seemed to have been left out... thanks.

PR #1410, opened 2016-04-02
Allow Surfaces larger than 2GiB

notlion:
Allow Surfaces larger than 2GiB

Previously creating surfaces larger than 2^31 bytes would cause an int32_t overflow on data size calculation, throwing a bad_alloc exception.

This commit converts mRowBytes, mInc and mRowInc to size_t, which causes size calculations to be performed in size_t before the result is used for array allocation and pointer arithmetic. Also, sites where these variables are used have been converted to avoid loss of precision.

### DIFF
--- a/blocks/Cairo/include/cinder/cairo/Cairo.h
+++ b/blocks/Cairo/include/cinder/cairo/Cairo.h
@@ -788,7 +788,7 @@ namespace cinder {
 class SurfaceConstraintsCairo : public cinder::SurfaceConstraints {
  public:
 	virtual SurfaceChannelOrder getChannelOrder( bool alpha ) const;
-	virtual int32_t				getRowBytes( int requestedWidth, const SurfaceChannelOrder &sco, int elementSize ) const;
+	virtual size_t				getRowBytes( int requestedWidth, const SurfaceChannelOrder &sco, int elementSize ) const;
 };
 
 } // namespace cinder
--- a/blocks/Cairo/include/cinder/cairo/Cairo.h
+++ b/blocks/Cairo/include/cinder/cairo/Cairo.h
@@ -788,7 +788,7 @@ namespace cinder {
 class SurfaceConstraintsCairo : public cinder::SurfaceConstraints {
  public:
 	virtual SurfaceChannelOrder getChannelOrder( bool alpha ) const;
-	virtual size_t				getRowBytes( int requestedWidth, const SurfaceChannelOrder &sco, int elementSize ) const;
+	virtual ptrdiff_t			getRowBytes( int requestedWidth, const SurfaceChannelOrder &sco, int elementSize ) const;
 };
 
 } // namespace cinder
--- a/blocks/Cairo/src/Cairo.cpp
+++ b/blocks/Cairo/src/Cairo.cpp
@@ -53,7 +53,7 @@ SurfaceChannelOrder SurfaceConstraintsCairo::getChannelOrder( bool alpha ) const
 	return ( alpha ) ? SurfaceChannelOrder::BGRA : SurfaceChannelOrder::BGRX;
 }
 
-size_t	SurfaceConstraintsCairo::getRowBytes( int requestedWidth, const SurfaceChannelOrder &sco, int elementSize ) const {
+ptrdiff_t SurfaceConstraintsCairo::getRowBytes( int requestedWidth, const SurfaceChannelOrder &sco, int elementSize ) const {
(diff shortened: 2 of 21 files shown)


notlion:
There's also a case to be made here for using `auto` at the usage sites. I would have done that in my own code, but here I opted to stick with explicit types.

richardeakin:
Thanks for this Ryan! Concerning `auto`, IMO its a judgement call on whether it makes the implementation code more readable and understandable. In the case of size_t vs int / int32_t, I think it is actually more readable to use size_t explicitly as it states 'we don't care about the memory layout of this number, but we do care that this number can handle the largest number we'd expect on whatever  architecture this code is running on'.

notlion:
I agree. Someone reading [this](https://github.com/cinder/Cinder/blob/8ff810b1f6f7a94112cfe493d1120ebdccbe2a01/src/cinder/Surface.cpp#L421-L429) code would probably need to know the types in order to understand the arithmetic. And you'd want the conversions anyway, even if the member types were changed back to something less precise.

andrewfb:
Thanks for taking a look at this - definitely interested to improve Cinder's large image support.

This change brings up a few subtle issues I think. One you mentioned Ryan is that the reason this change works is that it promotes rowbytes-dependent math to 64-bit; the value itself does not need to be 64-bit however, unless your image is wider than 2^29. `mPixelInc` should remain 8-bit as it basically correlates to the number of channels in the image. Also in some platforms, rowbytes can be negative (MSW images from GDI, for example), so a wholesale change to `size_t` is probably not ideal.

Another option might be to switch to `int64_t`. However I'm not confident that doesn't have performance implications on ARM in particular, and this is a case where that perf difference might be significant.

My first thought would be that we would prefer large image support on 64-bit machines, and to keep things essentially as they are on 32-bit machines. To that end, I'm curious if simply changing `mRowBytes` (and of course `getRowByes()`) to `ptrdiff_t` solves the use case you're seeing?

richardeakin:
I had a feeling this was going to be slightly more involved than a quick one pass.

> mPixelInc should remain 8-bit as it basically correlates to the number of channels in the image.

Hm, I'm confused - I see the documentation for `Surface::getPixelInc()` (here) says; "amount to increment a T* to increment by a pixel". This indicates that the value indicates by how many bytes you should be incrementing memory, but its not in itself the size of bytes you'd want to increment.

> Also in some platforms, rowbytes can be negative (MSW images from GDI, for example), so a wholesale change to size_t is probably not ideal.

Do you mean that you are decrementing a pointer while traversing the memory?  I don't see how a value that says 'this is how large I am' can be negative. Sorry, maybe we just don't understand the place where this comes in to play (actually I know that's it. :)

Concerning switching to int64_t and the performance problems on some hardware, this is the reason why there is a size_t so it seems to me the solution would be to use it, but make sure this 'negative bytes increment' case is handled correctly.

andrewfb:
`mPixelInc` measures channels because pointer arithmetic is performed in units of T, so to speak. It is 3 or 4 in the normal cases. So in a Surface32f, to advance one pixel, you `myFloatPtr += mPixelInc`. If `mPixelInc` were say 4, adding that to your float* results in an advance of 16 bytes, and one pixel.

`mRowBytes` is the number of bytes to add to advance one row in the image (traversing downward). In some scenarios particularly in MSW, the data pointer for the image points to the first pixel in the last row of a bottom-up image, so adding the negative `mRowBytes` value moves visually downward. Strange but true.

I think we're in agreement regarding the `size_t` tradeoffs. `ptrdiff_t` should solve both concerns (large images on 64-bit, and no perf implications on 32-bit).

notlion:
Ah, so `rowBytes` is more like `rowInc`, since it represents an offset? Since this can be negative I agree that `ptrdiff_t` is the correct type to use. My reasoning for changing `inc` variables to `size_t` was that as far as I can tell they are only used in pointer arithmetic, so a conversion will always be required. They are part of the public interface though, so I can understand the decision to use a smaller type.

I'll make this change later today.

notlion:
I've changed variables containing row increments to `ptrdiff_t` and pixel increment vars are back to `uint8_t` as per @andrewfb's input. This still has the same effect.

It's probably not in the scope of this PR, but maybe `getRowInc()` and `mRowInc` would be more accurate names? getRowBytes comes across to me as a positive size.

andrewfb:
Thanks for your work on this.

notlion:
NP. Thanks for the merge!

PR #1412, opened 2016-04-02
PolyLine should accept closed at initialization

drewish:
This let's me simplify some code like:
```c++
ci::PolyLine2f BuildingPlan::triangle()
{
    ci::PolyLine2f result( { ci::vec2(10, -10), ci::vec2(10, 10), ci::vec2(-10, 10), ci::vec2(10, -10)  } );
    result.setClosed();
    return result;
}
```

Down to a one liner.
```c++
ci::PolyLine2f BuildingPlan::triangle()
{
    return ci::PolyLine2f( { ci::vec2(10, -10), ci::vec2(10, 10), ci::vec2(-10, 10), ci::vec2(10, -10)  }, true );
}
```

### DIFF
--- a/include/cinder/PolyLine.h
+++ b/include/cinder/PolyLine.h
@@ -31,7 +31,7 @@ template<typename T>
 class PolyLineT {
   public:
 	PolyLineT() : mClosed( false ) {}
-	PolyLineT( const std::vector<T> &aPoints ) : mPoints( aPoints ), mClosed( false ) {}
+	PolyLineT( const std::vector<T> &aPoints, bool closed = false ) : mPoints( aPoints ), mClosed( closed ) {}
 	
 	const std::vector<T>&	getPoints() const { return mPoints; }
 	std::vector<T>&			getPoints() { return mPoints; }


mottosso:
I personally really like the named parameter idiom spread across Cinder. Could that work here?

```cpp
ci::PolyLine2f BuildingPlan::triangle()
{
    return ci::PolyLine2f( { ci::vec2(10, -10), ci::vec2(10, 10), ci::vec2(-10, 10), ci::vec2(10, -10)  } ).closed(true);
}
```

Though it should probably apply to the other arguments too.

drewish:
@mottosso yeah that's actually a nice pattern to follow, it'd be backwards compatible if we left the other methods in place.

richardeakin:
I also prefer a named method rather than additional bool as argument, because it is clearer at the call site. That said, I don't believe the classes like Polyline, Shape2d, are currently using the named parameter idiom, not that they shouldn't but need to weigh whether we want to start introducing methods in that fashion there.

andrewfb:
I'm inclined to accept this as written. I'm not opposed to NPI in instances like this, but it does feel like a policy that needs to revisited on a larger scale at some future date.

PR #1417, opened 2016-04-06
Variation of InterfaceGl::addParm that permits using only a "getter" function without a "setter" function.

2666hz:
Added a variation of InterfaceGl::addParm that permits using only a "getter" function without a "setter" function. The resulting param is read-only.
e.g.
	mParams->addParam<float>( "Speed", bind( &TweakBarApp::getSpeed, this ) );
or
	mParams->addParam<float>( "Speed km/h", [&]() -> float { return getSpeed() * (3600 / 1000); } );

### DIFF
--- a/include/cinder/params/Params.h
+++ b/include/cinder/params/Params.h
@@ -125,6 +125,8 @@ class InterfaceGl {
 
 		//!! Sets \a setterFn and \a getterFn as callbacks for this param. the target is ignored in this case.
 		Options&	accessors( const SetterFn &setterFn, const GetterFn &getterFn );
+		//!! Sets \a getterFn as callbacks for this param. the target is ignored in this case.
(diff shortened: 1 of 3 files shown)


richardeakin:
Sorry for the delay on this and thanks for the contribution.. I think this could be implemented by padding in a nullptr to the setter function to the already existing templated `addParamCallbackImpl( setter, getter )`, which would cut down on all the extra required methods other than a the Options wrapper:

```cpp
Options&	 accessors( const GetterFn &getterFn )
{
    return accessors( nullptr, getterFn );
}
```

The `addParamCallbackImpl()` implementation could then mark it as read-only if the setterFn is null.

Also if we're going to merge this, it should follow the [styleguide](https://github.com/cinder/Cinder/blob/master/CONTRIBUTING.md#style-guide) as much as possible.

PR #1424, opened 2016-04-11
Handle no available audio device in a recoverable manner.

richardeakin:
Previously the MSW Media Foundation implementation would blow up on an assertion failure if you tried to make a default audio input or output device and there wasn't connected hardware for that. Now the solution is as follows:

* `Device::getDefault[Input|Output]()` will return nullptr (empty `DeviceRef`) if there is no input or output device.
* Calling `Context::create[Input|Output]DeviceNode()` will cause an exception to be thrown if there is no default input or output device. This can be avoided by first querying for the default input or output device node before doing any other audio graph configuration.

### DIFF
--- a/src/cinder/audio/InputNode.cpp
+++ b/src/cinder/audio/InputNode.cpp
@@ -58,6 +58,14 @@ void InputNode::connectInput( const NodeRef &input )
 InputDeviceNode::InputDeviceNode( const DeviceRef &device, const Format &format )
 	: InputNode( format ), mDevice( device ), mLastOverrun( 0 ), mLastUnderrun( 0 )
 {
+	if( ! mDevice ) {
+		string errorMsg = "Empty DeviceRef.";
+		if( ! audio::Device::getDefaultInput() )
+			errorMsg += " Also, no default input Device so perhaps there is no available hardware input.";
+
+		throw AudioDeviceExc( errorMsg );
+	}
+
 	size_t deviceNumChannels = mDevice->getNumInputChannels();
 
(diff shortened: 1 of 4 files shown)


brucelane:
that is going to be useful for me, thank you!

brucelane:
hi,
would it be possible to remove the exception when the audio in is not present : https://github.com/cinder/Cinder/blob/master/src/cinder/audio/InputNode.cpp#L66
it causes my app to quit
thank you

richardeakin:
@brucelane apologies, I only now saw this. Can you explain a little what you mean by 'audio is not present'?

I would think that if the platform doesn't have a microphone and you try to create an `audio:: InputDeviceNode`, then an exception should be thrown to indicate that it isn't possible. To me this is better than returning a null object as it forces you to deal with the situation and provides better diagnostics.

brucelane:
when there no mic in (or it is disabled)...
it does reach the catch exception handled within a try/catch block, so there is always a crash...

richardeakin:
Do you mean that you are loading the `InputDeviceNode` from within a try / catch block, or you aren't and so you crash because of an uncaught exception?

If you don't want to use a try / catch, you can also check if there is a input device available first:
```cpp
if( audio::Device::getDefaultInput() ) {
     ...
}
```

brucelane:
yes, from within a try / catch block, but I removed it as it was useless
My code is here : https://github.com/videodromm/Cinder-Videodromm/blob/master/src/VDTexture.cpp#L919
getDefaultInput() will definitely be helpful, thank you

PR #1426, opened 2016-04-12
Build fix: Fixed file names in include

liliomk:
This caused build errors on case-sensitive filesystems.


richardeakin:
Thanks for this.

PR #1427, opened 2016-04-13
Fixed write position bug in BufferRecorderNode::process()

martinfm:
Now mWritePos gets updated in process() through std::atomic::compare_exchange_strong().

The update only happens if mWritePos didn't change since the beginningof process().

If it did change, then it means mWritePos has been reset to 0 by anotherthread calling start() and
therefore it must not be updated ( by adding numWriteFrames to it ) or the new recording will start with an offset of numWriteFrames from the beginning of the buffer.


richardeakin:
Thanks again..

PR #1432, opened 2016-04-14
Added mipmap support for 2d texture arrays

num3ric:
I added but then removed the mipmap init call from the void * data construction path, let me know what you think makes sense there.

Updated sample should make it easy to test.


src/cinder/gl/Texture.cpp
```diff
@@ -1673,6 +1672,20 @@ void Texture3d::update( const void *data, GLenum dataFormat, GLenum dataType, in
...
+	if( mMipmapping && mipLevel == 0 ) {
+		glGenerateMipmap( mTarget );
+	}
+}
+
+void Texture3d::initMaxMipmapLevel()
+{
+	if( mMaxMipmapLevel == -1 ) {
+		mMaxMipmapLevel = requiredMipLevels( mWidth, mHeight, mDepth ) - 1;
+#if ! defined( CINDER_GL_ES_2 )
```
sansumbrella: Probably want to cover any ES implementations here. I ran into trouble (all-black textures) with ES3 on tvos.

PR #1444, opened 2016-04-23
DDS default as 24 bit RGB for DDS files with no FOURCC

vade:
Experimentation with the DDS texture format parser and various exporters like Image Magic, GIMP and Graphics Converter lead me to notice that some DDS exporters do not write a valid fourcc pixel format.

This is an attempt to be right half the time. Admittedly its not the best, but maybe this PR could be used to start some discussion for a more robust fix?

Thanks.


vade:
This is properly fixed by #1476 - confirmed with my uncompressed DDS files.

PR #1445, opened 2016-04-23
Migrate some Buffer ivars to Protected

vade:
It might be nice to have BufferRef' subclasses be able to swizzle values for the mData, mAllocatedSize and mDataSize, particularly in the situation where page aligned memory is required for high performance disk reads or writes (unbuffered for example), as well as some types or DMA to PCI transfers.

### DIFF
--- a/include/cinder/Buffer.h
+++ b/include/cinder/Buffer.h
@@ -72,7 +72,7 @@ class Buffer {
 	//! Writes a Buffer to a DataTarget
 	void	write( const DataTargetRef &dataTarget );
 	
-  private:
+  protected:
 	void*	mData;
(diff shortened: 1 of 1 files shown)


richardeakin:
I did actually decide on something similar for the `audio::Buffer` classes, where an `audio::BufferDynamic` can be resized but other classes can just see it as the non-resizable `audio::Buffer`. So I can say, I think it could be a useful change, although it would be nice if there was code that made use of the protected variables rather than making them less restrictive for future purposes.

vade:
Hi RIchard,

I can see about sharing some page aligned buffer code from a client code base. Totally understood about this being a very minimal PR. 

://:; mbl.dev.

Issue #1073, opened 2015-08-21
Cinder 0.9: Tinderbox includes incorrect header file in CocoaView app


MattijsKneppers:
When using Tinderbox to create a CocoaView OpenGL app, it looks like it includes the cinder/app/CinderView.h dependency, which appears to no longer exist in the current master.

It looks like the cocoa approach has been reworked. The AppCocoaViewTest projects suggests that the h file needed is now cinder/app/cocoa/CinderViewMac.h

pizthewiz:
It appears as though the whole [CocoaView/OpenGL](blocks/__AppTemplates/CocoaView/OpenGL) template needs a few updates for 0.9.0.

kjhollen:
Just ran into this today and trying to work around it—having a little trouble figuring out the last parts. I got the app to compile, but it crashes on launch.

I've made a basic app called CinderCocoaViewTestApp that's just the result from running TinderBox. Here's what I've changed so far (happy to repost in another format that's not a git diff—wasn't sure what would be helpful):

```
diff --git a/include/CinderCocoaViewTestApp.h b/include/CinderCocoaViewTestApp.h
index 80b045c..f74210a 100644
--- a/include/CinderCocoaViewTestApp.h
+++ b/include/CinderCocoaViewTestApp.h
@@ -1,6 +1,6 @@
 #pragma once
 
-#include "cinder/app/AppCocoaView.h"
+#include "cinder/app/cocoa/AppCocoaView.h"
 
 class CinderCocoaViewTestApp : public cinder::app::AppCocoaView {
   public:
diff --git a/include/CinderCocoaViewTestDelegate.h b/include/CinderCocoaViewTestDelegate.h
index 0a82fc4..06dc2b3 100644
--- a/include/CinderCocoaViewTestDelegate.h
+++ b/include/CinderCocoaViewTestDelegate.h
@@ -1,11 +1,11 @@
 
-#include "cinder/app/CinderView.h"
+#include "cinder/app/cocoa/CinderViewMac.h"
 #include "CinderCocoaViewTestApp.h"
 #import <Cocoa/Cocoa.h>
 
 @interface CinderCocoaViewTestDelegate : NSObject <NSApplicationDelegate>
 {
-       IBOutlet CinderView             *cinderView;
+       IBOutlet CinderViewMac          *cinderView;
        IBOutlet NSWindow               *window;
        IBOutlet NSColorWell    *colorWell;
        
[...]

PR #1460, opened 2016-05-09
updated CocoaView OpenGL app template so it builds and works with v0.9.0

sroske:
Integrated changes from @kjhollen into the CocoaView OpenGL app template. Additionally changed the template to initialize the user application with a default renderer before it is instantiated.

Should fix issue #1073


andrewfb:
Apologies for the delay and thanks for this fix.

PR #1463, opened 2016-05-15
Added convenience gl::getVendorString().

isegal:
Adds cinder::gl::getVendorString() function which works similarly to cinder::gl::getVersionString() but for obtaining GL_VENDOR information.

### DIFF
--- a/include/cinder/gl/wrapper.h
+++ b/include/cinder/gl/wrapper.h
@@ -77,6 +77,8 @@ bool isExtensionAvailable( const std::string &extName );
 //! Returns the OpenGL version number as a pair<major,minor>
 std::pair<GLint,GLint>	getVersion();
 std::string getVersionString();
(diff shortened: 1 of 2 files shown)


richardeakin:
Sorry for the delay on this... Just a suggestion, but what if we just add a gl::getString() wrapper, so you can get the other ones too? Like:

```cpp
std::string vendorName = gl::getString( GL_VENDOR  );
std::string rendererName = gl::getString( GL_RENDERER );
```

paulhoux:
I like both suggestions, so maybe implement as follows:
```
std::string getString( GLenum name ) { return std::string( glGetString( name ) ); }
std::string getVendorString() { return getString( GL_VENDOR ); }
```

andrewfb:
Merged with the addition Paul proposed. Thanks for this.

PR #1464, opened 2016-05-15
CMake: Fix TK1 and RPi compilation process.

PetrosKataras:
The include path for mpg123 ( on the TK1 has to be manually built and installed under `/opt/local/` ) was not used. This fixes it and also separates the export directories of the config file so that you can build and maintain different GL targets that Cinder supports in parallel. It also fixes RPi compilation which was broken due to some forgotten / misplaced vars.


richardeakin:
This looks good except I think that using `CINDER_TARGET_GL` directly from within `cinderConfig.buildtree.cmake.in` will be problematic moving forward as build variants become more complex, for example with vulkan. Can we use `CINDER_LIB_DIRECTORY` instead? It is [defined in confiigure.cmake](https://github.com/cinder/Cinder/blob/3edb937b9123a77049faa46776e34c3510cdb324/proj/cmake/configure.cmake#L113) and already has `CINDER_TARGET_GL` as part of its path for platforms that need that.

PetrosKataras:
Sure, we can go with `CINDER_LIB_DIRECTORY` -- The export path is a bit more verbose right now but as you mention, moving forward with different build variants this should be something that gives us more flexibility.

richardeakin:
Thanks for the change. We can perhaps simplify the export paths going forward, but I like that the config variables are kept renderer-agnostic this way.

PR #1467, opened 2016-05-18
CMake: Implement option for building Cinder as a shared library.

PetrosKataras:
Set [BUILD_SHARED_LIBS](https://github.com/cinder/Cinder/compare/cmake...PetrosKataras:cmake-shared-lib-build-option?expand=1#diff-02073d153aae95890ea625f9a8b2c7b4R6) to `ON` for building Cinder as a shared library ( Currently tested only under Linux. )

### DIFF
--- a/include/cinder/audio/linux/DeviceManagerJack.h
+++ b/include/cinder/audio/linux/DeviceManagerJack.h
@@ -32,24 +32,24 @@ class DeviceManagerJack : public DeviceManager {
 	DeviceManagerJack();
 	virtual ~DeviceManagerJack();
 
-	const std::vector<DeviceRef>& getDevices()									override;
-	DeviceRef getDefaultOutput()												override;
-	DeviceRef getDefaultInput()													override;
+	const std::vector<DeviceRef>& getDevices()									override {};
+	DeviceRef getDefaultOutput()												override {};
+	DeviceRef getDefaultInput()													override {};
 
-	std::string getName( const DeviceRef &device )								override;
-	size_t getNumInputChannels( const DeviceRef &device )						override;
-	size_t getNumOutputChannels( const DeviceRef &device )						override;
-	size_t getSampleRate( const DeviceRef &device )								override;
-	size_t getFramesPerBlock( const DeviceRef &device )							override;
+	std::string getName( const DeviceRef &device )								override {};
+	size_t getNumInputChannels( const DeviceRef &device )						override {};
+	size_t getNumOutputChannels( const DeviceRef &device )						override {};
+	size_t getSampleRate( const DeviceRef &device )								override {};
+	size_t getFramesPerBlock( const DeviceRef &device )							override {};
 
-	void setSampleRate( const DeviceRef &device, size_t sampleRate )			override;
-	void setFramesPerBlock( const DeviceRef &device, size_t framesPerBlock )	override;
+	void setSampleRate( const DeviceRef &device, size_t sampleRate )			override {};
+	void setFramesPerBlock( const DeviceRef &device, size_t framesPerBlock )	override {};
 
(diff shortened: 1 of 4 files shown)


chaoticbob:
This is really cool! Out of curiosity, which category of samples have you tested this against?

PetrosKataras:
Tested initially with BasicApp and then with [this](https://github.com/PetrosKataras/Cinder-CMakeBlockApp) one for a more complex configuration ( requires assimp, Cinder-Skinning and the OSC block ) . I can push a cmake branch on that one if you would like to test. 

I want to try at some point the runtime stuff so this is a step towards that direction.

richardeakin:
Very nice, I like that it seems to be the de facto way to do this with cmake.

Not sure about the mods to DeviceManagerJack.h, are those necessary? Seems to me that if it isn't implemented yet it is better left out of the build, so that you get compiler warnings until it is finished. Either that, or implement with assertion failures or something so it doesn't silently fail (personally I prefer compile errors and leaving out of build system for now of the two options).

PetrosKataras:
They are just necessary for not having undefined references when building Cinder as a shared lib since now there are just function signatures in this file but I suppose a more appropriate way to handle it would be one of the two options you mention. My vote is also for leaving it out of the build system for now until its properly implemented.

richardeakin:
@PetrosKataras I tried building this branch as shared lib and ran into the following linker errors:

```bash
mkdir build
cd build
cmake -DBUILD_SHARED_LIBS=1 -DCINDER_BUILD_SAMPLE="_opengl/Cube" ..
...
make -j4
...
/usr/bin/ld: ../lib/linux/x86_64//libboost_system.a(error_code.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
../lib/linux/x86_64//libboost_system.a: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
CMakeFiles/cinder.dir/build.make:6281: recipe for target '../lib/linux/x86_64/ogl/Debug/libcinder.so' failed
make[2]: *** [../lib/linux/x86_64/ogl/Debug/libcinder.so] Error 1
make[2]: Leaving directory '/media/r/data/code/cinder/cinder-linux/build'
CMakeFiles/Makefile2:63: recipe for target 'CMakeFiles/cinder.dir/all' failed
make[1]: *** [CMakeFiles/cinder.dir/all] Error 2
```

Looks like something in the boost libs aren't compatible?

Also I made [this branch](https://github.com/richardeakin/Cinder/tree/cmake_petros_shared) that cherry picks the shared_lib changes and then comments out the jack cpps as we discussed (both that branch and yours produces this linker error). I'm happy to merge that once this linker error is resolved, or if yours if you take the same approach. Any ideas?

richardeakin:
Edit: just committed that change to cmake branch that omits jack audio impl cpps for now..

PetrosKataras:
It seems that the x86_64 Boost libs would need to be recompiled with the `-fPIC` flag in order to play nicely with a shared version of Cinder. I didn't face this issue on the TK1 where I actually tried this so maybe there it has been included already? Maybe @chaoticbob has more info on that -- Also @andrewfb upgraded recently the Boost libs to 1.60 on the android_linux branch so it would be interesting to know the flags used there also so that it is future proof.

richardeakin:
Okay, well I don't see any point in holding up this - I'll just cherry-pick your commit and open a new issue around investigating whether rebuilding boost solves that linker error.

PetrosKataras:
Cool, I m closing this one then.

andrewfb:
I believe this should be up-to-date as far as Boost build flags:
https://github.com/cinder/Cinder/blame/android_linux/docs/htmlsrc/guides/boost/index.html#L117

Issue #1454, opened 2016-05-01
Cairo: Wisteria sample fails on call to copyWindowSurface()


richardeakin:
Tested by running the sample and hitting 'p'.  Originally reported here:

https://forum.libcinder.org/topic/invalid-context-in-wisteria-sample-cinder-0-9-0-mac#23286000002673088

andrewfb:
This should be fixed by the above.

PR #1471, opened 2016-05-24
Make sure CGContext exists before attempting to get window contents.

kitschpatrol:
Possible fix for issue #1454. 

In the Wisteria example, copyWindowSurface() is called from a key event, which means the CGContextRef currentRef has been released already by the finishDraw method in RendererImpl2dMacQuartz.mm.

This PR adds a check for the currentRef, and will create a temporary reference if needed.

I think the NSBitmapImageRep was also being released twice, since it's called by convertNsBitmapDataRep with "assumeOwnership" set to true, which should take care of clean-up.

Finally, I noticed that the channel order coming from initWithFocusedViewRect was ARGB instead of RGBA. Updated convertNsBitmapDataRep accordingly. (But not sure if this will break something elsewhere?)

Open to any feedback / revisions / insights. Thank you.

### DIFF
--- a/src/cinder/app/cocoa/RendererImpl2dMacQuartz.mm
+++ b/src/cinder/app/cocoa/RendererImpl2dMacQuartz.mm
@@ -71,11 +71,20 @@
 - (NSBitmapImageRep*)getContents:(cinder::Area)area
 {
 	[view lockFocus];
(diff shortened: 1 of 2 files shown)


andrewfb:
Thanks for this fix and (triple) fix.

PR #1476, opened 2016-06-10
Fixed an issue with uncompressed dds texture

philcn:
it should respect the DDPF_FOURCC bit in the pixel format flags (often
absent in uncompressed dds files) and handle the case when fourcc is not
used.


richardeakin:
Head's up that this broke UWP and _ANGLE builds, both give the following compiler errors:

```
1>  TextureFormatParsers.cpp
1>d:\code\cinder\cinder-linux\src\cinder\gl\textureformatparsers.cpp(283): error C2065: 'GL_BGRA': undeclared identifier
1>d:\code\cinder\cinder-linux\src\cinder\gl\textureformatparsers.cpp(289): error C2065: 'GL_BGR': undeclared identifier
```

@andrewfb Can you see a fix for this?

PR #1489, opened 2016-06-24
fix (app impl mac): windowMovedNotification: use main screen as origin for coordinate space transformation

MattijsKneppers:
In the current master, when the output window is not on the main screen, getWindowPosY() reports the incorrect Y position after moving the window once. This fixes that issue.


richardeakin:
Thanks for this and sorry for the delay..

I can verify this is a bug in master branch and your fix is good. I also found that the position was still off during the app's initial resize, so I'm creating a PR that addresses that as well (including your commit). Going to link to this momentarily..

PR #1492, opened 2016-07-06
Add TARGET_LIBRARIES param to ci_make_app

izumin5210:
e.g. PCL (PointCloudLibrary) needs to call `target_link_libraries`

http://www.pointclouds.org/documentation/tutorials/using_pcl_pcl_config.php

```cmake
ci_make_app(
  SOURCES           ${PROJECT_SOURCES}
  INCLUDES          ${PCL_INCLUDE_DIRS}
  CINDER_PATH       ${CINDER_PATH}
  TARGET_LIBRARIES  ${Boost_LIBRARIES} ${PCL_LIBRARIES}
)
```

### DIFF
--- a/proj/cmake/modules/cinderMakeApp.cmake
+++ b/proj/cmake/modules/cinderMakeApp.cmake
@@ -58,7 +58,7 @@ function( ci_make_app )
 
 	add_executable( ${ARG_APP_NAME} MACOSX_BUNDLE WIN32 ${ARG_SOURCES} ${ICON_PATH} ${ARG_RESOURCES} )
 	target_include_directories( ${ARG_APP_NAME} PUBLIC ${ARG_INCLUDES} )
-  target_link_libraries( ${ARG_APP_NAME} cinder ${ARG_LIBRARIES} )
+	target_link_libraries( ${ARG_APP_NAME} cinder ${ARG_LIBRARIES} )
 
 	if( CINDER_MAC )
 		# set bundle info.plist properties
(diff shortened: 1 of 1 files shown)


richardeakin:
I'm good with this, I think @PetrosKataras also wanted something like this. Right now we're working through how we'll support cinderblocks through cmake, they also require `add_dependency()` so considering what if anything we could do around that..

PetrosKataras:
Looks good to me. Since you can still use `target_link_libraries` directly and it doesn't restrict you, I think its nice to provide an argument on the macro also for less familiar users.

richardeakin:
@izumin5210 I think we're going to move forward with this, afterwards we've also discussed adding a `BLOCKS` argument that does a little more based on cinder's blocks structures (like `add_dependencies()` and knows about cinder's folder structure). A couple requests however:

1) For consistency, what about simply using the name `LIBRARIES`, which follows the current pattern we have with `SOURCES` and `INCLUDES`.

2) As of now we're using tabs in the cmake files instead of spaces, let's keep that consistent so that future additions don't create unnecessary whitespace changes.

If you're short on time and agree, I'm able to do these things after the merge, or I'll wait for feedback from you.

Cheers and thanks again for the contribution!

izumin5210:
@richardeakin Thank you for your feedback 😄 
I fixed the name and replaced spaces with tabs !

izumin5210:
Thank you for your reviews and merge!!! 😂

PR #1494, opened 2016-07-08
add missing files and libs to cmake for gstreamer

patrickFuerst:
_same as #1493 but moved to an extra branch so the PR is not created from the android_linux branch._ 

This solves issues of linking the QuickTimeBasic example. 
On Linux the source files for the Movieplayer were never added to the cinder lib. 
As well as the glib library flag for the targets was missing.


PetrosKataras:
+1 for this one

richardeakin:
Thanks for the fix.

PR #1496, opened 2016-07-12
Fix an invalid path in gensamplebuilds.

tanitta:
Change "cbuilder" to "cibuilder". 
This makes it possible to generate valid build files to build samples on linux.

### DIFF
--- a/tools/linux/scripts/gensamplebuilds
+++ b/tools/linux/scripts/gensamplebuilds
@@ -8,7 +8,7 @@ import stat
 print( "Samples Linux Build File Generate v0.001\n" )
 
 Build = """#!/bin/sh
-$<CINDER_REL_PATH>/tools/linux/cbuilder -app "$@"
+$<CINDER_REL_PATH>/tools/linux/cibuilder -app "$@"
 """
 
 CMakeListsTxt= """# $<APPNAME>


PetrosKataras:
Hey,
thanks for this. This brings up a good question - What are we doing with the linux directory that exists currently and the old build scripts now that we have merged all the CMake work ? My suggestion would be to remove them since there is less room for confusion and errors in this case to my opinion ( I also think that was the original idea no..? maybe I m mistaken ) . If we need something similar we should probably check what is there already and how it could be used with our new CMake configuration or re-write some helper modules.

richardeakin:
Yes all `linux` folders should disappear. I removed the ones from the samples folder but forgot a few, like the one in root and those in test.. its on my todo list unless someone else would like to PR it.

PetrosKataras:
The `test/Linux` could stay no? I think it could come handy in some scenarios and I am also using it currently for the video player.

richardeakin:
Ah yes, Linux tests are fine and welcome of course. Just any `linux` folders that were for project configs should go in favor of the new cmake setup.

richardeakin:
Closing this as we've removed the old linux folders. Thanks again, we'll happily welcome more contributions towards the new cmake system for whatever platform you're working on.

PR #1497, opened 2016-07-12
Added samples demonstrating GPU based level-of-detail techniques and supporting methods in ci::gl

philcn:
Two samples _opengl/LevelOfDetailBasic and _opengl/LevelOfDetailIndirect are created. LevelOfDetailBasic shows basic workflow of calculating LOD on the GPU in an instanced render, using transform feedback and geometry shader. LevelOfDetailIndirect improves the previous sample by utilizing `gl::draw*Indirect` calls, eliminating CPU-GPU synchronization overhead. The sample demonstrates two ways of accessing transform feedback result - atomic counter and query buffer object.

Additions to `ci::gl::BufferObj`, `ci::gl::Query` are made to facilitate those techniques. `gl::Batch`, `gl::VboMesh` and `gl::Context` are extended to support a series of new OpenGL multi/indirect draw calls. 

Linux support for GL debug output is implemented.


richardeakin:
Love the docs. :)  Very nice work.

ryanbartley:
Very well done. Exciting to see this in Cinder. Two things: 

1) Is it necessary to keep the const GLvoid* for the *indirect calls. I know that's the interface for gl but the use case would be offset more often than not. 

2) possible upgrade of the sample and of cinder, while you're at it :), would be to add the functions for glDrawTransformFeedback*, as the query for how many transform feedback primitives written is brutal on the pipeline. That's the reason behind the transform feedback object in the first place, so you wouldn't have to query it. [Here's](https://www.opengl.org/wiki/GLAPI/glDrawTransformFeedback) a link for more. 

Otherwise, it looks great. I've wanted these functions whitelisted for a long time but never got around to makings samples. Great job!

PR #1500, opened 2016-07-14
Run OSC on separate thread in multithreaded samples.

sansumbrella:
Previously, a separate `io_service` was running, but the OSC objects were still using the application's primary `io_service`, which is processed on the main thread.


sansumbrella:
Also, it's worth considering moving the `PacketFraming` parameter to the end of the TCP constructor so the parameters to UDP/TCP are otherwise identical in order.

That way, we could only have a single location for the compiler switch by using type aliases.
```
#if UDP
using Sender = SenderUdp;
#else
using Sender = SenderTcp;
#endif
…
mSender( 10000, destinationHost, destinationPort, Sender::protocol::v4(), *mIoService )
```

PR #1506, opened 2016-07-17
Syncing master and android_linux branches

richardeakin:
There were a few conflicts in MSW files so that platform still needs to be tested, especially UWP.


richardeakin:
There seems to be a minor issue that is breaking the UWP / Angle build due to [this PR](https://github.com/cinder/Cinder/pull/1476), however the conflict resolution looks correct in this PR so I'm going to merge it.

PR #1513, opened 2016-07-24
Fix oversizing of geom::Sphere index buffer

### LINKED ISSUE #1498: geom::Sphere seems to have too many indices?
I came across this while reading through the sphere generation code in an effort to ~~steal~~ learn from it.

In `geom::Sphere::loadInto`, the `indices` array is `resize`d to `numSegments * numRings * 6` on line `1894`, when in fact it seems like only `(numSegments - 1) * (numRings - 1) * 6` indices are ever added to it (see the nested loop on line `1921`). Or maybe there's something going on there which I'm not noticing?

I guess this isn't a significant issue, but if I'm not mistaken, it would imply that some extra useless zeroes are being loaded into index buffers?

Cheers! M

mhintz:
Fixes #1498


paulhoux:
All good, thanks for fixing this.

PR #1514, opened 2016-07-24
Adds scoped texture binding to gl::drawEquirectangular

### LINKED ISSUE #1279: gl::drawEquirectangular does not bind texture...
...and therefor nothing shows up if the function is called.

mhintz:
Fixes #1279 

Thought I'd grab an easy one to cross off the list. I'm not sure if having only scoped texture bind is appropriate here, or if the function needs to save the existing context state first, and then restore it after drawing.


richardeakin:
>  I'm not sure if having only scoped texture bind is appropriate here, or if the function needs to save the existing context state first, and then restore it after drawing.

Actually that's what the `Scoped*` guys do - the previous state will be restored once that object goes out of scope.

Thanks for this!

PR #1515, opened 2016-07-24
Add geom::WireRect convenience class

### LINKED ISSUE #1235: geom::WireRect missing
Just a quick reminder that this is still missing.

mhintz:
Closes #1235

Includes “test”, which uses this instead of geom::WirePlane in the Geometry sample. Looks good to me!


paulhoux:
Thanks for this pull request. What about texture coordinates, normals and colors?

mhintz:
Oh, sure! I based this on the WireRoundedRect class, which doesn't have them, but I'll add them

paulhoux:
Hmm, if that's the case we should add those attributes to `WireRoundedRect` as well. :)

mhintz:
I just had time to take another look at this, and it looks like all of the Wire* classes only include position data, regardless of which attributes are requested in the call to loadInto. So this implementation of WireRect is in keeping with that convention.

That being said, I could see the case for adding other attributes to the Wire* classes. Seems like it should be a separate PR though

richardeakin:
@mhintz that makes sense to me (covering all Wire* attribs separately). Partly because, while colors are straightfoward (and usually chosen to be arbitrary which isn't useful except for debugging), what should the normals or texcoords be for something analogous to GL_LINES? They usually aren't rendered with a lighting model and you can't map a texture to them.

mhintz:
@richardeakin yeah, that makes sense. I can't think of many use cases for texture coordinates where you wouldn't just use the regular non-Wire geom classes. I'm happy to add arbitrary colors to the wireframes though.

This PR stems from a moment a few weeks ago going through the Cinder issue backlog and looking for stuff that I felt was within my knowledge to check off. Hope it helps!

richardeakin:
> This PR stems from a moment a few weeks ago going through the Cinder issue backlog and looking for stuff that I felt was within my knowledge to check off. Hope it helps!

Thanks for contributing! We really do appreciate folks going through open issues and helping where they can. I moved the colors feature to #1547

I'm testing here on OS X and seeing a visual problem with the new `WireRect()`:

screenshot 2016-08-10 11 32 27

Not sure but I think of the positions off, should be in a counter-clockwise circle no? Also not sure how much it matters when using `LINES`. Debug per-vertex colors would help here. :)

mhintz:
Uh oh, that's not what I remember seeing! I'll take a look at it

paulhoux:
Sorry about the confusion regarding normals and so on. Those aren't required for the ```Wire*``` versions of the primitives. I must have been sleeping when I wrote that comment.

paulhoux:
Hi Mark,

I just created a new PR based on this one: #1617   It contains a fix for the incorrect primitive type and the order of the vertices.

Issue #393, opened 2013-12-11
TGA support broken on Windows


hyperknot:
Loading TGA image files seems to be broken on Windows.

    First-chance exception at 0x751CC41F in ImageFileBasic.exe: Microsoft C++ exception: cinder::ImageIoExceptionFailedLoad at memory location 0x0018E7C0.
    First-chance exception at 0x751CC41F in ImageFileBasic.exe: Microsoft C++ exception: cinder::ImageIoExceptionFailedLoad at memory location 0x0018E7C0.
    First-chance exception at 0x751CC41F in ImageFileBasic.exe: Microsoft C++ exception: [rethrow] at memory location 0x00000000.
    First-chance exception at 0x751CC41F in ImageFileBasic.exe: Microsoft C++ exception: [rethrow] at memory location 0x00000000.
    unable to load the texture file!

I tested it with the official ImageFileBasic sample and a Photoshop made TGA image in 16/24/32 bits.

Sample files:
[16-bit](http://storage.z-e-r-o.in/zsolt/2013/12/tga16.tga)
[24-bit](http://storage.z-e-r-o.in/zsolt/2013/12/tga24.tga)
[32-bit](http://storage.z-e-r-o.in/zsolt/2013/12/tga32.tga)

andrewfb:
Hi - we only support the image file formats that WIC does on Windows, which unfortunately doesn't include TGA.

richardeakin:
TGA is now supported on windows thanks to #1517 and stb image.

PR #1517, opened 2016-07-24
Add StbImage read and write to Windows platform to support more image formats.

vinjn:


vinjn:
This change will also fix #393

richardeakin:
This seems good to me - also sometimes for whatever reason people don't seem to have WIC decoders found on their system, at least this would make cinder apps on their computer slightly less broken.

richardeakin:
@BanTheRewind I was able to switch to v140, rebuild cinder and then build BasicApp. Can you try that on your end?

It also realized, we should add those files to `PlatformWinRt` / vc2015_winrt project too..

PR #1522, opened 2016-07-26
Implements vsync for glfw-based linux apps.

PetrosKataras:


andrewfb:
Thanks for this - tested on Ubuntu 15.10

PR #1523, opened 2016-07-27
Added CinderUserDefines.h

richardeakin:
Allows users to create their own file with the same name and if they
build cinder from their own .sln file, theirs will be used instead,
allowing them to customize preprocessor defines within cinder codebase.

I’m open to suggestions as to where to put CinderUserDefines.h, there has to be a dummy one somewhere in cinder codebase that gets used if the user doesn’t provide one in their soln directory. Here, I've added it to the base of the include folder.

### DIFF
--- a/include/CinderUserDefines.h
+++ b/include/CinderUserDefines.h
@@ -1,0 +1,2 @@
+// This is a dummy header that will be #included if the user doesn't provide one in their solution directory
+// If you add a file with the same name as this one to your solution directory, it will be used instead when building libcinder
--- a/vc2013/cinder.vcxproj
+++ b/vc2013/cinder.vcxproj
@@ -159,7 +159,7 @@
     <ClCompile>
       <AdditionalOptions>/MP %(AdditionalOptions)</AdditionalOptions>
       <Optimization>Disabled</Optimization>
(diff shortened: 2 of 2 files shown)


paulhoux:
That's a neat trick, Rich. Let me try that.

paulhoux:
Hi Rich, I checked out your branch and tried it by adding a `CinderUserDefines.h` file to my project. Cinder is part of the solution, so gets recompiled as well. In my custom header I added a `#define CI_MIN_LOG_LEVEL 0` and then called `CI_LOG_V("Test");` from my `setup()` method. The log message should now also be generated for the Release target.

Sadly, this did not work. It did work, however, when I added an `#include "CinderUserDefines.h"` at the top of `cinder/Log.h`, but that solution would require us to add it to (nearly) all Cinder header files and that would be impractical. I am afraid the `<ForcedIncludeFiles>CinderUserDefines.h</ForcedIncludeFiles>` does not work in this case.

Were you able to make it work successfully?

paulhoux:
Wait... if I also add `<ForcedIncludeFiles>CinderUserDefines.h</ForcedIncludeFiles>` to my custom project, it works!

richardeakin:
Yes I also had to add the Force Include Files line to my main app's project for it to work within app code. The coolest thing is that this also affects cinder core code's usage of logs, asserts, whatever preprocessor things you want to adjust on a per-project basis.

For background, this is where I found the trick: http://stackoverflow.com/questions/5272556/can-you-make-a-vc-solution-set-preprocessor-defines-on-loaded-projects/5273036#5273036

richardeakin:
Moved to #1526.

PR #1525, opened 2016-07-28
Added qtime files for CMake in OSX

gaborpapp:
The qtime files are missing from the CMake build. Without this the QuickTimeBasic sample fails to link on OSX with CMake:

```
[100%] Linking CXX executable Debug/QuickTimeBasic.app/Contents/MacOS/QuickTimeBasic
Undefined symbols for architecture x86_64:
  "cinder::qtime::MovieGl::getTexture()", referenced from:
      QuickTimeSampleApp::update() in QuickTimeBasicApp.cpp.o
  "cinder::qtime::MovieGl::MovieGl(boost::filesystem::path const&)", referenced from:
      cinder::qtime::MovieGl::create(boost::filesystem::path const&) in QuickTimeBasicApp.cpp.o
  "cinder::qtime::MovieBase::play(bool)", referenced from:
      QuickTimeSampleApp::loadMovieFile(boost::filesystem::path const&) in QuickTimeBasicApp.cpp.o
  "cinder::qtime::MovieBase::isDone() const", referenced from:
      QuickTimeSampleApp::update() in QuickTimeBasicApp.cpp.o
  "cinder::qtime::MovieBase::isPlaying() const", referenced from:
      QuickTimeSampleApp::loadMovieFile(boost::filesystem::path const&) in QuickTimeBasicApp.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [Debug/QuickTimeBasic.app/Contents/MacOS/QuickTimeBasic] Error 1
make[1]: *** [CMakeFiles/QuickTimeBasic.dir/all] Error 2
make: *** [all] Error 2
```

### DIFF
--- a/proj/cmake/platform_macosx.cmake
+++ b/proj/cmake/platform_macosx.cmake
@@ -28,6 +28,15 @@ list( APPEND SRC_SET_AUDIO_COCOA
 	${CINDER_SRC_DIR}/cinder/audio/cocoa/FileCoreAudio.cpp
 )
 
+list( APPEND SRC_SET_QTIME
+	${CINDER_SRC_DIR}/cinder/qtime/AvfUtils.mm
+	${CINDER_SRC_DIR}/cinder/qtime/AvfWriter.mm
+	${CINDER_SRC_DIR}/cinder/qtime/MovieWriter.cpp
+	${CINDER_SRC_DIR}/cinder/qtime/QuickTimeGlImplAvf.cpp
+	${CINDER_SRC_DIR}/cinder/qtime/QuickTimeImplAvf.mm
+	${CINDER_SRC_DIR}/cinder/qtime/QuickTimeUtils.cpp
+)
+
 # specify what files need to be compiled as Objective-C++
 list( APPEND CINDER_SOURCES_OBJCPP
 	${CINDER_SRC_DIR}/cinder/Capture.cpp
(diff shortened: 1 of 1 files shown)


proj/cmake/platform_macosx.cmake
```diff
@@ -28,6 +28,15 @@ list( APPEND SRC_SET_AUDIO_COCOA
 	${CINDER_SRC_DIR}/cinder/audio/cocoa/FileCoreAudio.cpp
 )
 
+list( APPEND SRC_SET_QTIME
+	${CINDER_SRC_DIR}/cinder/qtime/AvfUtils.mm
+	${CINDER_SRC_DIR}/cinder/qtime/AvfWriter.mm
+	${CINDER_SRC_DIR}/cinder/qtime/MovieWriter.cpp
+	${CINDER_SRC_DIR}/cinder/qtime/QuickTimeGlImplAvf.cpp
```
PetrosKataras: Is there a reason for including `QuickTimeGlmplAvf.cpp` twice?

gaborpapp: The second one ensures that the file is built as objective-c++. I think it does not compile in c++ mode.

PetrosKataras: I see.. Maybe the one at `SRC_SET_QTIME` could be removed then ? It just feels a bit weird to have it as a duplicate especially if its not getting compiling in c++ mode.

gaborpapp: I can remove it, but I was not sure how it is preferred, since most of these files under `CINDER_SOURCES_OBJCPP` are also listed in `libcinder_source_files.cmake`.

PetrosKataras: ping @richardeakin

gaborpapp: I think the OBJCPP files are not built in themselves. This list is used to change their build flags only here https://github.com/cinder/Cinder/blob/android_linux/proj/cmake/platform_macosx.cmake#L59
So I believe adding them twice is necessary at the moment.

PetrosKataras: True, this seems the case actually looking at it closer. Sorry for the noise!

PR #1526, opened 2016-07-29
Add ability to define preprocessor flags that affect cinder from outside of the cinder repo

richardeakin:
This is carried over from #1523, with the source branch moved to the main repo.

Allows users to create their own file with the same name and if they
build cinder from their own .sln file, theirs will be used instead,
allowing them to customize preprocessor defines within cinder codebase.

I’m open to suggestions as to where to put CinderUserDefines.h, there has to be a dummy one somewhere in cinder codebase that gets used if the user doesn’t provide one in their soln directory. Here, I've added it to the base of the include folder.


richardeakin:
RFC has been posted on forum [here](http://discourse.libcinder.org/t/rfc-ability-to-define-preprocessor-flags-on-msw-without-modifying-cinder-repo/227).

Issue #1448, opened 2016-04-25
OSC: pass the undelivered message as parameter in SocketTransportErrorFn


araid:
Having the entire message available when there's a socket error would give the user more options to handle disconnections and allow him to re-send the message if he wanted to.

Read the forum discussion for more info:
https://forum.libcinder.org/topic/rfc-new-osc-block#23286000002674037

ryanbartley:
So, I think this should maybe be a larger discussion on the pertinent arguments for error handler interfaces in general in the block. Sometimes, errors aren't because of a send or receive. Sometimes the sockets explode and the error has, for instance, an endpoint of 0.0.0.0, which isn't necessary and probably confusing but I didn't want to make error handling costly and obtuse. If anyone else would like to chime in that would be great.

ryanbartley:
As far as the problem of lost data with SocketTransportErrorFn, I think it would be easy to allow the passage of osc::Message "data" as vector<uint8_t> to the handler and one could either take hold of the passed vector, or create a new osc::Message with that vector, and resubmit the data when the problem with the socket has been figured out.  This would be a minor change, yet breaking to those who already use the error system inside of the OSC block. 

I believe the above question is still relevant as the error handling right now is very bare bones.

richardeakin:
I'm still new to Ryan's design, but my understanding is that you're using a global error callback (`SenderBase::setSocketTransportErrorFn`) to handle all errors which might occur async, and no longer have information about what Osc::Message caused the error.  I could see it being useful, if nothing else than just to monitor and debug, to know exactly what message caused the error (without having to break with a debugger), and it does seem one solution would be to pass the Message via the error callback.

A different approach that would be closer to the way modern async APIs work would be to add optional parameters to the `send()` methods to handle errors that way. Might even be nice to have an completeFn() as well, so the api could work like:

```cpp
mSender.send( msg, osc::SenderOptions()
    .completeFn( [] { CI_LOG_I( "yay, completed" ); }
    .errorFn( [this, msg] { /* .... report that msg failed or retry */ } )
);
```

If those std::function's are null (unset), then the callbacks wouldn't fire and there's no overhead except a little extra memory usage per message (not that OSC is meant to be a very memory conservative protocol). Just one possibility that came to mind..

ryanbartley:
[...]

PR #1529, opened 2016-08-01
Osc error fix

ryanbartley:
This fix is addressing #1448. Not completed because I want to talk about receiver side.


ryanbartley:
Updated TUIO to reflect the changes in this last commit. Take a look.

PR #1530, opened 2016-08-02
GStreamer video player refactor and macosx support.

PetrosKataras:
The Movie interface should be almost complete now. What remains yet to be implemented is :
- `setActiveSegment`
- `resetActiveSegment`
- `stepBackward`
- `seekToEnd`

These are a bit tricky and they will need some more testing to see what works best.  

This has been tested on Xubuntu 16.04, on the TK1 and on the RPi2 with the latest Raspbian image. Note that the GStreamer libraries that ship with latest on the RPi are super old and one should definitely consider an uninstalled setup for getting proper hardware acceleration. The [create_gst_uninstalled_env.sh](https://github.com/PetrosKataras/GStreamerScripts) script does exactly this for you. In any case for GStreamer versions < 1.6 the player will work with the fallback CPU path.  

This PR also add support for the `GstPlayer` on `macosx`.  The following steps are needed for using the GstPlayer on OS X :  
- Install GStreamer. The latest version is 1.9.1 and can be found [here](https://gstreamer.freedesktop.org/data/pkg/osx/1.9.1/). Start first by installing `gstreamer-1.0-1.9.1-x86_64.pkg` and continue then with `gstreamer-1.0-devel-1.9.1-x86_64.pkg` ( order matters ).
- Define `CINDER_MAC_USE_GSTREAMER` when building Cinder.  
  i.e `cmake .. -DCINDER_MAC_USE_GSTREAMER=1`

Also included under `/test/Linux/GstPlayerRefactorTest/` is an app for testing the video player in various cases like stepping forward, play/pause, seeking. More cases will probably be added in the future.  

Although not 100% complete I would personally consider this work much finer compared to my very initial GstPlayer implementation that currently lives in the `android_linux` branch so I m putting it here in hope of more testing and discussion from interested parties.


richardeakin:
Looks like excellent improvements and I think its fine to get this in before being at 100% feature parity with other video players we've already got.

The video in the test folder - do we need this? It is sort of large (5mb) and also was wondering what the rights are on it.  If we do ship a video, I think it would be better for it to live in [samples/data/video/](https://github.com/cinder/Cinder/tree/master/samples/data) so that any other test / sample can share it.

PetrosKataras:
I could swear that I found this test video in one of the existing Quicktime samples but looking at it now that's not the case :) 

Its a small part of the [Big Buck Bunny](http://bbb3d.renderfarming.net/download.html) from Blender which is licensed under CC-BY so I think it should be ok to include but if we are concerned with the file size I could also replace it with an http stream for now or just require the user to load a local file on its own.

src/cinder/linux/GstPlayer.cpp
```diff
@@ -20,1298 +17,1147 @@
...
 float GstPlayer::getFramerate() const
 {
-    return mGstData.mFrameRate;
+    return mGstData.frameRate;
 }
 
 bool GstPlayer::hasAudio() const
 {
-    return mGstData.mHasAudio;
+    if( ! mGstData.pipeline ) return false;
```
richardeakin: Can you place the return statements on a new line (there are a few below as well)? It makes stepping through the code easier.

src/cinder/linux/GstPlayer.cpp
```diff
@@ -20,1298 +17,1147 @@
...
+
+    gint64 duration;
+    if( isPrerolled() ) {
+        if( ! gst_element_query_duration( mGstData.pipeline, GST_FORMAT_TIME, &duration ) ) {
+            g_warning( "Cannot query duration." );
+            return -1.0;
+        }
+        mGstData.duration = duration;
+    }
+    else{
```
richardeakin: format nit - space between `else` and bracket.

src/cinder/linux/GstPlayer.cpp
```diff
@@ -20,1298 +17,1147 @@
...
-void GstPlayer::createTextureFromMemory()
+void GstPlayer::createTextureFromID()
 {
-	mMutex.lock();
-	std::swap( mFrontVBuffer, mBackVBuffer );
-	mMutex.unlock();
+#if defined( CINDER_GST_HAS_GL )
+    mMutex.lock();
+    std::swap( mCurrentBuffer, mNewBuffer );
+    mMutex.unlock();
```
richardeakin: I don't fully understand the thread syncing logic necessary here, but wouldn't the unlock need to be around the `getTextureID()` as well? Not really sure how this could be called from more than one thread, though.

src/cinder/linux/GstPlayer.cpp
```diff
@@ -20,1298 +17,1147 @@
...
+    }
 }
 
-GstData& GstPlayer::getGstData()
+void GstPlayer::createTextureFromMemory()
 {
-	return mGstData;
+    mMutex.lock();
+    std::swap( mFrontVBuffer, mBackVBuffer );
+    mMutex.unlock();
```
richardeakin: Same threading question here, if the unlock needs to happen after `mFrontBuffer` is used below.

include/cinder/linux/GstPlayer.h
```diff
@@ -16,212 +15,226 @@
...
+	std::thread	 	mGMainLoopThread; // Seperate thread for GMainLoop.
+	
+	std::mutex 		mMutex; // Protect  since the appsink callbacks are executed from the streaming thread internally from GStreamer.
+
+	bool 			mUsingCustomPipeline;
+	GstData 		mGstData; // Data that describe the current state of the pipeline.
+	
+	ci::gl::Texture2dRef	mVideoTexture;
+
+	unsigned char* 	    	mFrontVBuffer = nullptr;
```
richardeakin: Suggestion: instead of using raw pointers here, I like using vector<char> and then passing off its .data(). It avoids having to clean up and then it is very easy and legible to resize as necessary.

test/Linux/GstPlayerRefactorTest/src/GstPlayerTestApp.cpp
```diff
@@ -0,0 +1,316 @@
...
+{
+    try {
+        // load up the movie, set it to loop, and begin playing
+        mMovie = qtime::MovieGl::create( moviePath );
+        mMovie->setLoop();
+        mMovie->play();
+	console() << "Playing: " << mMovie->isPlaying() << std::endl;
+    }
+    catch( ci::Exception &exc ) {
+        console() << "Exception caught trying to load the movie from path: " << moviePath << ", what: " << exc.what() << std::endl;
```
richardeakin: Check out [`CI_LOG_EXCEPTION()`](https://github.com/cinder/Cinder/blob/master/include/cinder/Log.h#L375-L379), really handy for things like this.

MikeGitb:
Is there a special reason, why you are not using `std::lock_guard` or `std::unique_lock` instead of manually locking and unlocking the mutex?

PetrosKataras:
@MikeGitb Hi, I didn't see any immediate benefits in the specific use case but for the sake of modern syntax I went ahead and replaced with `std::lock_guard` and `std::unique_lock` . I still manually unlock the `std::unique_lock` since I would prefer not to introduce a new scope in the places it is used.

patrickFuerst:
I'm currently working on the remaining implementations of 

* setActiveSegment
* resetActiveSegment

In gstreamer it's fairly easy by defining a segment region by doing a proper seek event with the right start and end time set. 

To do proper seamless looping though *SEGMENT_SEEK* should be used and handling looping as soon as the message *GST_MESSAGE_SEGMENT_DONE* gets posted on the bus. 

@PetrosKataras, is there a reason not to do it this way ? 

My suggestion would be to use _SEGMENT_SEEK_ whenever the media should loop, and a normal seek when it shouldn't loop. 

Probably on a desktop computer there will always be a seamless loop. But on a Raspberry Pi a stutter is noticeable.

patrickFuerst:
First test to use seamless loop and segment selection can be found in this commit. a511bf7fa45dbec11eaf7a3dedacee707f08e828

ryanbartley:
@PetrosKataras, what exactly are you worried about with introducing a new scope for the unique_lock?

src/cinder/linux/GstPlayer.cpp
```diff
@@ -20,1298 +17,1183 @@
...
+    }
+    else {
+        resetSystemMemoryBuffers();
+    }
 }
 
 void GstPlayer::resetGLBuffers()
 {
-	if( mInputVideoBuffers ) {
-		while( g_async_queue_length( mInputVideoBuffers ) > 0 ) {
```
paulhoux: This is not exception safe. I'd advise to use ```std::lock_guard``` here.

PetrosKataras: Hi Paul, I think you are commenting on an old diff. This uses `std::lock_guard` [currently](https://github.com/cinder/Cinder/pull/1530/files#diff-2fa55ca57e9d9db9bc25a9b271a53352R1070).

PetrosKataras:
@ryanbartley I am not worried about introducing a new scope in terms of functionality but more in terms of readability in these cases. With the support for both cpu and gpu path the [processNewSample](https://github.com/PetrosKataras/Cinder/blob/4278270c45f2ab6902ecea4a5312b261a93c7e05/src/cinder/linux/GstPlayer.cpp#L1122) function has quite some scoping happening so the addition of a new one was something that I thought we can skip. That said, if the general preference is this then I m happy to add a new scope.

Regarding the use of the g* variants for printing again it has been just a matter of visual preference for distinguishing better messages that are directly related to GStreamer from a Cinder app logging. Also happy to switch if people feel that this would be a better way to go.

Considering all these I have to say that this PR has grown quite a bit and maybe it would be a good idea to get some testing for the actual functionality and if we are happy with that merge it and do micro-optimizations in separate PR's ?

PetrosKataras:
@patrickFuerst Hi Patrick, I haven't had the time to test your seamless loop tests yet but I don't see any reason not to try the SEGMENT_SEEK approach for seamless looping. As you mentioned on Desktop the current approach works pretty well but if we can get better looping results on other devices like the RPi through the seek approach I am down for trying it out.

ryanbartley:
@PetrosKataras you're right. I've checked everything on mac and linux and it works. I'd say personally we should switch the g* log stuff over to cinder as it would be easier to pinpoint problems, but I'll defer.

src/cinder/linux/GstPlayer.cpp
paulhoux: Hmm, strange, I meant to comment on what's currently line 1131. That still has the unique_lock.

patrickFuerst:
As soon as I find some time to test it more I'm gonna make a PR regarding SEGMENT_SEEK.

richardeakin:
We've been discussing about what's left here in order to get this merged into android_linux within the week (which will soon be merged into master). Here is what is left:

- [ ] instead of `.unlock()`ing the `unique_lock`s, use scope the scope to dictate when they unlock, which will be consistent when both exceptions happen and more legible.
- [ ] remove the test bbb.mp4, instead use a FileDropEvent to allow users to load a video from their hard drive.

There are a couple other minor things we've discussed, and more features on the way, but this should be good enough for us to get this massive improvement merged and we'll continue on afterwards with a new PR. Thanks to everyone for the productive conversations and contributions here!

PR #1531, opened 2016-08-02
Added "label" to Params...

paulhoux:
...allowing you to specify an alternative name for your parameter. Normally, each parameter should have a unique name and this shows up in the Params window as well. Using a label, you can override this default name with a more appropriate name.


richardeakin:
Looks good to me, but maybe the doc line should indicate the utility in this? Also using it in the ParamsBasic could be nice, if nothing else than for testing the functionality.

paulhoux:
Thanks, Rich!

PR #1533, opened 2016-08-03
Added GlslProg support for Ssbo

m1keall1son:
Completes the integration of Shader Storage Blocks into GlslProg.

Includes two new structs for cacheing API Introspection data about the active shader storage blocks within the glsl program.

adds a new define to gl/platform.h to compile out the code for unavailable platforms.

Currently only will work on MSW OpenGL version 4.3 and above
Tested on Windows 10

test implementation can be found in:
tests/SSboIntegration 

Let me know how its working!


notlion:
Nice. Just tried this out. Looks like some good functionality to have in GlslProg, along the lines of what we do with Uniform Blocks. Perhaps the test should be fleshed out a bit more, and utilize more of the functionality in this PR. As is, I was able to comment out the call to `shaderStorageBlock()`, and the test still runs correctly. Personally, I'd like to see some kind of use case for the introspection functionality.

PR #1534, opened 2016-08-03
Throw SerialExcWriteFailure on failure to write.

sansumbrella:
Previously, Cinder erroneously threw a Read error on failure to write.


cinder:
Thanks for this

PR #1536, opened 2016-08-06
Cmake linux updates: removing old files

richardeakin:
* Removing the old linux-specific cmake files
* Removing old cibuild scripts
* Updated ENV variable to `CINDER_PATH` for consistency.

Please let me know if I removed something I shouldn't have.

There are still some `linux` folders in [test/Linux](https://github.com/cinder/Cinder/tree/android_linux/test/Linux) - I don't know if I should blow these away before they are updated to work with the new build system, but they won't work until then anyhow. Would appreciate help here as I'm not in front of a linux box much these days.


PetrosKataras:
#1539 should take care of the test Linux folder.

PR #1540, opened 2016-08-09
Add option for manual include/link of a block.

PetrosKataras:
If there is no need for a block to define a target ( e.g header-only blocks ) with this PR you have the option to define `BLOCKNAME_INCLUDES` and `BLOCKNAME_LIBRARIES` in order to specify block dependencies.

First we check the presence of a `BLOCKNAME` target and if not present we check for the existence of the above mentioned variables. If present they are manually added through `target_include_directories` / `target_link_libraries`.


richardeakin:
Superceded by #1560.

PR #1545, opened 2016-08-10
Fix default RPi build.

PetrosKataras:
Relevant forum thread [here](http://discourse.libcinder.org/t/errors-building-cinder-on-raspberry-pi-3/193/4).

GL sources should not be included by default when building on the RPi.


richardeakin:
I think that we shouldn't build in GLLoad sources whenever `CINDER_GL_ES` is set. Other platforms like windows would still want this and it would be nice to keep it unified if at all possible, where android (still on the todo) won't use GLLoad, either.

richardeakin:
Superseded by #1582

PR #1549, opened 2016-08-10
Enable building and running tests with cmake.  Fixes issue 1548.

dankegel:
Seems to work, modulo a boost problem specific to my system...?


dankegel:
dammit, wrong branch.

PR #1550, opened 2016-08-10
cmake: enable building and running catch-style unit test suite. 

dankegel:
After doing 'make', to run the tests, do 'make test' (shows little output) or 'ctest -v' (shows more output) or
Debug/TestMain (bare metal, lets you use catch.hpp options).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,7 @@ set( CMAKE_VERBOSE_MAKEFILE ON )
 project( cinder )
 
 option( CINDER_BUILD_SAMPLES "Build all samples." OFF )
+option( CINDER_BUILD_TESTS "Build all unit tests." OFF )
 set( CINDER_BUILD_SAMPLE "" CACHE STRING "Build a specific sample by specifying its path relative to the samples directory (ex. '_opengl/Cube')." )
 
 set( CINDER_PATH      "${CMAKE_CURRENT_SOURCE_DIR}" )
@@ -46,4 +47,10 @@ elseif( CINDER_BUILD_SAMPLE )
 		message( STATUS "adding sample: ${CINDER_BUILD_SAMPLE}" )
 	endif()
 	add_subdirectory( ${CINDER_PATH}/samples/${CINDER_BUILD_SAMPLE}/proj/cmake )
+endif()
+
+if( CINDER_BUILD_TESTS )
+	enable_testing()
+	add_subdirectory(${CINDER_PATH}/test/unit/proj/cmake )
+	add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} --verbose)
 endif()
--- a/test/unit/proj/cmake/CMakeLists.txt
+++ b/test/unit/proj/cmake/CMakeLists.txt
@@ -1,0 +1,44 @@
+cmake_minimum_required( VERSION 3.0 FATAL_ERROR )
+set( CMAKE_VERBOSE_MAKEFILE ON )
+
+project( TestMain )
+
+get_filename_component( CINDER_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../../../.." ABSOLUTE )
+get_filename_component( UNIT_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../" ABSOLUTE )
+
+include( "${CINDER_PATH}/proj/cmake/modules/cinderMakeApp.cmake" )
+
+SET(SOURCES
+	${UNIT_DIR}/src/Base64Test.cpp
+	${UNIT_DIR}/src/JsonTest.cpp
+	${UNIT_DIR}/src/ObjLoaderTest.cpp
+	${UNIT_DIR}/src/RandTest.cpp
+	${UNIT_DIR}/src/SystemTest.cpp
+	${UNIT_DIR}/src/TestMain.cpp
+	${UNIT_DIR}/src/UnicodeTest.cpp
+	${UNIT_DIR}/src/audio/BufferUnit.cpp
+	${UNIT_DIR}/src/audio/FftUnit.cpp
+	${UNIT_DIR}/src/audio/RingBufferUnit.cpp
+	${UNIT_DIR}/src/signals/SignalsTest.cpp
+)
+
+ci_make_app(
+	SOURCES     ${SOURCES}
+	CINDER_PATH ${CINDER_PATH}
+	INCLUDES "${UNIT_DIR}/src"    # for catch.hpp
+)
+TARGET_COMPILE_DEFINITIONS(
+	TestMain
+	PRIVATE -DUNIT_DIR_ASSETS="${UNIT_DIR}/assets"
+)
+
+IF (APPLE)
+	SET(TESTBINDIR ${CMAKE_BINARY_DIR}/${CMAKE_BUILD_TYPE}/TestMain.app/Contents/MacOS)
+ELSE()
+	SET(TESTBINDIR ${CMAKE_BINARY_DIR}/${CMAKE_BUILD_TYPE})
+ENDIF()
+
+ADD_TEST(
+	NAME TestMain
+	COMMAND ${TESTBINDIR}/TestMain
+)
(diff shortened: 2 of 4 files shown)


lucasvickers:
- On OS X, the binary get's embedded inside the .app folder `./Debug/TestMain.app/Contents/MacOS/TestMain`, any easy way to add that to the search folders for `TestMain`?

- Works on the linux side except ctest seems to suppress the output when run via `make test`.  This is pretty easily resolved with running via `ctest --verbose`.  Not sure if we want to show output by default?

- Any reason you didn't include the unit tests in the subfolders [audio](https://github.com/cinder/Cinder/tree/master/test/unit/src/audio) and [signals](https://github.com/cinder/Cinder/tree/master/test/unit/src/signals)?


Thanks!  This is going to be really useful for our CI.  Removes the need for an additional testing script.

richardeakin:
Yea, this brings to light a little problem with the assets folder, in that the app is expected to live relative to the assets folder. If you're building out-of-source then the app gets built in `${CINDER_REPO}/build/Debug/...`, so it can't find the `${CINDER_REPO}/test/unit/assets` folder. Need to think on this..

lucasvickers:
Just to clarify, I meant for `ctest` to find the binary, otherwise you see:
```
$ make test
Running tests...
Test project /code/lib/ctest/build
    Start 1: TestMain
Could not find executable /code/lib/ctest/build/TestMain
Looked in the following places:
/code/lib/ctest/build/TestMain
/code/lib/ctest/build/TestMain
/code/lib/ctest/build/Release/TestMain
/code/lib/ctest/build/Release/TestMain
/code/lib/ctest/build/Debug/TestMain
/code/lib/ctest/build/Debug/TestMain
/code/lib/ctest/build/MinSizeRel/TestMain
/code/lib/ctest/build/MinSizeRel/TestMain
/code/lib/ctest/build/RelWithDebInfo/TestMain
/code/lib/ctest/build/RelWithDebInfo/TestMain
/code/lib/ctest/build/Deployment/TestMain
/code/lib/ctest/build/Deployment/TestMain
/code/lib/ctest/build/Development/TestMain
/code/lib/ctest/build/Development/TestMain
code/lib/ctest/build/TestMain
code/lib/ctest/build/TestMain
code/lib/ctest/build/Release/TestMain
code/lib/ctest/build/Release/TestMain
code/lib/ctest/build/Debug/TestMain
code/lib/ctest/build/Debug/TestMain
code/lib/ctest/build/MinSizeRel/TestMain
code/lib/ctest/build/MinSizeRel/TestMain
code/lib/ctest/build/RelWithDebInfo/TestMain
code/lib/ctest/build/RelWithDebInfo/TestMain
code/lib/ctest/build/Deployment/TestMain
code/lib/ctest/build/Deployment/TestMain
code/lib/ctest/build/Development/TestMain
code/lib/ctest/build/Development/TestMain
Unable to find executable: /code/lib/ctest/build/TestMain
1/1 Test #1: TestMain .........................***Not Run   0.00 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.00 sec

The following tests FAILED:
	  1 - TestMain (Not Run)
Errors while running CTest
make: *** [test] Error 8
```

dankegel:
Try it now.  "make check" should give you the verbose output we both crave :-),
it should find the tests on MacOSX, and it adds the tests I missed.

test/unit/src/audio/FftUnit.cpp
```diff
@@ -3,7 +3,7 @@
 // FIXME: OOURA roundtrip FFT seems to be broken on windows for sizeFft = 4 (https://github.com/cinder/Cinder/issues/1263)
 #if ! defined( CINDER_MSW )
 
-#include "catch.hpp"
+#include "../catch.hpp"
```
sansumbrella: Shouldn't the header search path for the tests be consistent so we don't need to describe relative paths in the test files themselves?

sansumbrella:
Yay unit tests!

test/unit/src/audio/FftUnit.cpp
dankegel: If you want to move catch.hpp somewhere more sensible, I'm fine with that.  Where do you want it?

Or are suggesting adding a -I .. when building the subdirs?  That'd be kinda ugly.

lucasvickers:
The cmake functionality works really well!  Note: there are some unix unit test issues, opening #1551 to track those.  Pending the ongoing comments about header location and asset directories this looks good to me.

dankegel:
Not sure I get the asset directory one.  Are you objecting to sprinkling
app::Platform::get()->addAssetDirectory(UNIT_DIR_ASSETS);
in each file that uses assets?  I'd love a cleaner solution.

Note that the tests pass for me on MacOS X now, so the above is probably sufficient.

test/unit/src/audio/FftUnit.cpp
richardeakin: I'd imagine adding `INCLUDES "${UNIT_DIR}/src"` when calling `ci_make_app()` would do it. I can't really think of a more reasonable place for catch.hpp to live other than `cinder/test/unit/src/catch.hpp` but of course open to other places.

lucasvickers:
@richardeakin are you okay with `app::Platform::get()->addAssetDirectory(UNIT_DIR_ASSETS);` for now?  I know long term we're talking about increasing the number of parent directories an app will search, but for now it's just isolated to the unit tests.

@dankegel basically it just brought back up a convo that with the cmake folder structures we need to support searching further up the folder index.

test/unit/src/audio/FftUnit.cpp
sansumbrella: I'm not familiar enough with CMAKE to say how to set the search paths or for it to be clear what you mean by build the subdirs.

The catch unit tests can all be (and should be, I believe) compiled into a single binary that can select different test cases to run. So I visualize the file hierarchy as a bunch of things relative to the root of the test project. Effectively, you're building one thing from the root of the `src/` directory and linking in things from the subdirectories. So in XCode, there's just one project at the root and all the include directories are handled therer.

dankegel:
fun fact: increasing the number of parent directories wouldn't help here.

lucasvickers:
Whoops, you are correct.

test/unit/src/audio/FftUnit.cpp
dankegel: This patch does indeed build a single catch unit test executable, just as you expected.

richardeakin: Actually, I can think of a better place to put the catch header - at `cinder/test/unit/include/catch.hpp`. The benefit here that I see is that if other projects that already depend on cinder want to use catch.hpp, they can add that folder to their include path without picking up any of cinder's unit tests.  But the move can be done outside of this PR, and probably should as the xcode and VS project files will need to be updated.

dankegel:
Right then, have a look now.  No more nasty ../check.h's.

richardeakin:
This is looking great. Open to merging now, but wanted to bring up a couple points first:

1)  Should the `enable_testing()` and `add_custom_targets()` in the main CMakeLists.txt file only be added if a certain cache variable is defined, say `CINDER_TEST_ENABLE` or something like that? Seems that it would speed up builds a bit for those just wanting to rebuild libcinder.

2) Fine for now but it seems like making the .app bundle on OS X isn't needed here, since it is meant to be run from the command line. Perhaps in the future we need some sort of `ci_make_app()` variant for command line apps, or give it an extra BOOL argument so it doesn't do all the `MACOSX_BUNDLE WIN32` like things.

dankegel:
how about CINDER_DISABLE_TESTS, on the theory that running tests should be on the easy path?

richardeakin:
Personally, I think the majority of people don't want to build the tests - both people like me who are frequently rebuilding cinder while prototyping various things, or common users who just want to build master branch, shouldn't have to worry about tests. We're planning to have CI do the worrying for us. :)

I was also basing the name and default on what the author of [GLM chose](https://github.com/g-truc/glm/blob/c947b3338bf4442cdc41d70785c7f0d645e750f5/CMakeLists.txt#L29), as I think it works well there.

dankegel:
I added CINDER_BUILD_TESTS next to the existing CINDER_BUILD_SAMPLES, how's that look?

richardeakin:
Yea that makes sense to me, grouping them that way. Looks ready to merge to me!

Thanks again for working on this, really nice to see us moving towards more automation and less manual labor.

dankegel:
I can't wait for it to land, it'll make working on issue 1544 easier.

PR #1556, opened 2016-08-13
Implement linux system info

dankegel:
Fixes bug 1551.

Some of the values are a little approximate, and some are left unimplemented (what does OS version mean on Linux, anyway).

"make check" rather closer to passing on linux with this.


andrewfb:
Thanks for taking a look at this. Anyone have thoughts on using [uname()](https://linux.die.net/man/2/uname) to approximate OS version?

Issue #1429, opened 2016-04-13
Remove _d suffix from libcinder binaries, using folder structure instead


richardeakin:
If we move away from using the `_d` suffix for debug builds, like in `cinder_d.lib` / `libcinder_d.a`, and instead using paths like `Debug/cinder.lib` and `Release/cinder.lib`,  build configurations (Visual Studio projects, xcode, and cmake) will become a little more maintainable, for example you'll be able to set the linked libraries to the same value for all configurations using a `CONFIGURATION` macro to determine the correct path.

The downside is that it will break existing projects, which will have to update their configuration to point to something like `${CINDER_PATH}/lib/${PLATFORM_TARGET}/$(CONFIGURATION}/cinder.lib`.

There's already going to be some project path breakage when #1307 is completed, so we'll make this change around the same time.

richardeakin:
At a similar time, I'd like to make an adjustment to the VS output directories, so that it is easier to clean up afterwards when you want to do a complete wipe. I've been using the following in other recent projects:

Output Directory: `$(ProjectDir)\build\$(Platform)\$(Configuration)\`
Intermediate Directory: `$(ProjectDir)\build\intermediate\$(Platform)\$(Configuration)\`

So everything ends up in a folder called `build` (which is already how its done with Xcode).

vinjn:
I love the idea of `build` folder.

chaoticbob:
I think the remove of the _d will also make Android and Linux easier as well. I think there's still a few loose ends on Android and Linux where we still have to do the add manually. Linux (and maybe Android) already do the pathed Debug/Release. Would be nice to stick them into a build folder.

FWIW: The reason the build scripts on Linux and Android is called cibuild instead of build was to leave room for a build directory. The initial versions of the build systems did have a build directory, can't remember why it was removed.

richardeakin:
Closed by #1561

Issue #1307, opened 2016-02-07
reorganize project files into `proj` folder


richardeakin:
This is to prevent a growing number of folders in the base path of the cinder repo or tinderbox-built projects. Proposed structure will look like:

```
cinder_repo
-- blocks/
-- docs/
-- include/
-- proj/
---- android
---- cmake
---- xcode
---- vc2013
---- vc2015_winrt
-- lib/
-- samples/
-- src/
-- test/
-- tools/
```

Tinderbox should also create projects with this layout, although its TBD if we're going to update existing samples for this new layout.

richardeakin:
Closed by #1307

richardeakin:
Reopening after I noticed that the android folder is still in the base folder, which should be moved to `proj/android`.

richardeakin:
`proj/android` move completed with 66ddc2445c36c453c286569669e3ec925bbed26a, this is resolved.

PR #1561, opened 2016-08-15
Relocating static library, Xcode / VC projects

andrewfb:
_Please note that this a breaking change for existing projects._

This address #1429 and #1307.
• macOS library now lives at `lib/macosx/$(CONFIGURATION)/libcinder.a`
• iOS library now lives at `lib/ios/$(CONFIGURATION)/libcinder.a` & `lib/ios-sim/$(CONFIGURATION)/libcinder.a`
• MSW library now lives at `lib/msw/$(PlatformTarget)/$(Configuration)/$(PlatformToolset)/cinder.lib`
• WinRT/UWP library now lives at `lib/msw-uwp/$(PlatformTarget)/$(Configuration)/$(PlatformToolset)/cinder.lib`

The primary motivation for this change is to allow projects to share the same values in both Debug and Release configurations linker settings, by leaning on the _$(Configuration)_ and similar build macros.

In addition, this PR moves `xcode/`, `vc2013/` and `vc2015_winrt/` into `proj/`. This should reduce clutter in the root of the repository.

This change modifies samples and TinderBox templates accordingly.


paulhoux:
Looks good, makes sense, will be a pain to fix in current projects.

PR #1565, opened 2016-08-23
Fixing small memory leak in DirectShow capture implementation

sroske:
The surface cache uses a custom deleter for the Surface shared_ptr objects, but that custom deleter doesn't free up the allocated Surface objects. This small change cleans that that memory leak.


richardeakin:
Thanks for the fix!

richardeakin:
Actually, thinking about it - that delete shouldn't happen, it will destroy the data in the SurfaceCache (it keeps a separate shared_ptr for those).

It is a bit confusing as its using shared_ptr to store the internal data, which was written before C++11 so it couldn't use a unique_ptr..

I think this needs to be reverted.

richardeakin:
... and staring at it even longer, I think this delete is indeed necessary. It doesn't delete the actual memory block, it deletes a non-owning Surface8u that is wrapping the memory block..

So verdict - PR is good. :)

PR #1566, opened 2016-08-24
PostProcessAA sample: use Batch instead of slower gl::drawSolidRect()

richardeakin:
Also fixed a typo.


paulhoux:
These look good to me.

PR #1569, opened 2016-08-27
Trim

paulhoux:
Added utility functions to trim a string for both ```std::string```, ```std::u16string``` and ```std::u32string```.


paulhoux:
Might need to use ```std::iswspace``` for the unicode version, let me check.

paulhoux:
I think I'll use ```boost/algorithm/string.hpp``` instead, because I'm not sure about the ```u16string``` and ```u32string``` versions. Had hoped to move away from boost, but now may not be the time.

PR #1571, opened 2016-08-30
Added Arcball missing includes

gaborpapp:
If you include `cinder/Arcball.h` first in your project it does not compile because `CameraPersp` and `WindowRef` are not defined, only forward declared. Tested with clang on OS X.

### DIFF
--- a/include/cinder/Arcball.h
+++ b/include/cinder/Arcball.h
@@ -24,10 +24,12 @@
 
 #pragma once
 
+#include "cinder/Camera.h"
 #include "cinder/Quaternion.h"
 #include "cinder/Vector.h"
 #include "cinder/Sphere.h"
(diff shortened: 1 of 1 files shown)


richardeakin:
Makes sense to me, good catch. Ideally though I would think that the Arcball's implementation should live in Arcball.cpp to help reduce include clutter (app/Window.h pulls in quite a lot itself).

gaborpapp:
Should I move the implementation to Arcball.cpp instead?

richardeakin:
My vote is for creating an Arcball.cpp, but that is more work (needs to be added to every cinder project file) and also depends on what the author (@andrewfb) thinks. Going to merge this for now, we can remove the Windows.h include if that happens.

PR #1573, opened 2016-09-03
Fix for Timeline crash.

paulhoux:
Caching the iterator::end will not prevent it from becoming invalidated, causing a crash when the call to stepTo() changes the items. We now cache the items themselves instead and use these to iterate over them. Since the keys and items are (shared pointers), there isn't much overhead when you copy them.

### DIFF
--- a/src/cinder/Timeline.cpp
+++ b/src/cinder/Timeline.cpp
@@ -63,11 +63,11 @@ void Timeline::stepTo( float absoluteTime )
 	
 	eraseMarked();
 	
-	// we need to cache the end(). If a tween's update() fn or similar were to manipulate
+	// we need to cache the items. If a tween's update() fn or similar were to manipulate
 	// the list of items by adding new ones, we'll have invalidated our iterator.
 	// Deleted items are never removed immediately, but are marked for deletion.
-	s_iter endItem = mItems.end();
-	for( s_iter iter = mItems.begin(); iter != endItem; ++iter ) {
+	auto items = mItems;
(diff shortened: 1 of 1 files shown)


richardeakin:
Wouldn't this be a significant performance change? It requires making a copy of a multimap for each tween, each frame.  Copying `shared_ptr`s can actually be a performance problem as it has to be done atomically.

Maybe the situation that causes a crash can be clarified (or reproduced), and we can find a way to avoid it without making that copy.

paulhoux:
Yeah, you're probably right about the overhead. It's pretty easy to reproduce this crash: just call ```Timeline::reset()``` or ```Timeline::clear()``` from a callback. I've seen the comment saying this is not allowed, but I really need this functionality to work for reasons.

The problem is similar to how object can unsubscribe from receiving events *while* receiving said event. That should (and AFAIK in Cinder does) just work. It's too easy to crash the app this way and it isn't immediately obvious to the average user what's going on or why it isn't allowed.

Would you know an alternative way of fixing this?

andrewfb:
I've considered some large-scale surgery to Timeline which would defer & queue modifications when it's iterating. However I think that's a bigger change than we'd have appetite for at this stage of a release cycle. Can you not set some piece of app-level state in your callback that you then act on after the Timeline update?

MikeGitb:
The original comment in the code was wrong anyway: The end iterator never gets invalidated and doesn't need to be cached. The loop iterator is what gets invalidated when you remove the element it is referring to.

There should also be no need to queue insertions (unless deferred insertion is exactly the semantic you want).

PR #1576, opened 2016-09-10
ExtrudeSpline fixes and thickness function addition

simongeilfus:
This addresses several issues with the current implementation and adds a new function that allows to create extrusion with a non-uniform thickness.

It also add uv-mapping as the current uv-mapping is not usable.

I'm still investigating an issue where normals are not generated properly (bug already present in master) and will add a test project soon.


andrewfb:
Thanks for this feature + matching test

PR #1584, opened 2016-09-20
symlink assets directory to be next to executable

richardeakin:
Opening this for discussion. This is a solution for the problem mentioned in [this forum post](http://discourse.libcinder.org/t/rfc-cross-platform-cmake-support/102/25), where the assets fold can't be build when you build a cinder sample with something like:

```
cmake -DCINDER_BUILD_SAMPLE=Geometry ..
```

In order for this to work, the executable has to be placed in a subfolder, I used used the app name for this.

It looks for the assets folder in its common place by default (next to proj folder), but this can be overridden by passing in an argument for `ASSETS_PATH` to the `ci_make_app()` function. The symlink is currently always named `assets`.

Still to-do:
- [x] add cmake option that forces the assets to be copied instead of symlinked.
- [x] test on mac

### DIFF
--- a/proj/cmake/modules/cinderMakeApp.cmake
+++ b/proj/cmake/modules/cinderMakeApp.cmake
@@ -129,6 +129,11 @@ function( ci_make_app )
 		get_filename_component( ASSETS_SYMLINK_PATH "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/assets" ABSOLUTE )
 
 		if( EXISTS "${ARG_ASSETS_PATH}" AND IS_DIRECTORY "${ARG_ASSETS_PATH}" )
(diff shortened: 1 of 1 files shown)


richardeakin:
This is complete. You can now build, for example, both cinder and the Geometry sample in one go by doing in the base cinder folder:

```
mkdir build
cd build
cmake -DCINDER_BUILD_SAMPLE=Geometry ..
make
```

The Geometry app will then live in the build/Debug/Geometry folder (on Linux, it will be a binary in that folder, on OS X it will be the Geometry.app bundle). There will be an assets symlink next to it. If you also define `-DCINDER_COPY_ASSETS=ON`, then the assets folder will be hard copied over instead.

Issue #997, opened 2015-06-24
iOS: Audio spikes/cuts when the screen is locked.


notlion:
If the screen is locked manually or the device goes to sleep while sound is playing a tone is played at full volume. After this happens all audio is muted on the device until it is rebooted.

I've tested this with the NodeBasic sample on an iPad Air 1st gen (iOS 8.3, current cinder/master), but have also seen the same behavior on an iPad Mini 1st gen.

Perhaps AVFoundation emits some callback we are expected to respond to by reducing the volume?

richardeakin:
I'm probably not handling the iOS notifications right for when to pause / resume audio. It's long been on my TODO list, but I'd love it if someone using iOS more oftern than myself (almost never) were to look into it.

Here's what I think it is: I couldn't ever get [these notifications](https://github.com/cinder/Cinder/blob/master/src/cinder/audio/cocoa/DeviceManagerAudioSession.mm#L260-L285) to fire, once they are working, we can add some signal callbacks to DeviceManager, perhaps `getSignalInturruptionBegan()` and `getSignalInterruptionEnd()`. Then the audio::OutputDeviceNode can hook into these to pause / resume processing.

richardeakin:
Oh, and we don't need to worry about the pre-iOS 6 route, we're bumping the min runtime target to 6 so that code should just be removed.

richardeakin:
I finally had a chance to look at this and I can't say I have a solution. `AVAudioSessionInterruptionNotification` doesn't seem to fire when you hit the power button (although it does when you get a phone call), so I don't think the AVAudioSession notifications are any good to us.

Trying a different angle, using the `getSignalDidEnterBackground()` / `getSignalDidBecomeActive()`, which correspond to UIApplicationDelegate's `applicationDidEnterBackground:` and `applicationWillEnterForeground` appear to be too late - I get a sudden pop when I hit the power button, and when I come back there is no audio. Afterwards, the only way to get audio for any app is to restart the device, just as you found.

I'm curious what other projects are doing. Frankly, my knowledge of iOS is far outdated as I don't do much on that platform these days (and I dread having to try, since it usually takes more than an hour just to get an app to deploy to the device from Xcode). I'd like to find a solution to this nasty problem but I don't know of any.

richardeakin:
[...]

PR #1585, opened 2016-09-23
[iOS] Nix audio glitches during sleep and interruptions

notlion:
This adds handling for iOS interruptions (phone call, alarm, etc) and adds a fix for #997, a bug that was causing audio to glitch and drop out when the device slept, or another application was opened while audio was playing. In my tests audio now fades out smoothly when the app is backgrounded. There is no fade-in, but this is something users can add on a per-app basis.

I should mention that the interruption handling uses the iOS >= 6.0 API, as I've been told that 6.0 is going to be the new baseline.

### DIFF
--- a/include/cinder/audio/cocoa/CinderCoreAudio.h
+++ b/include/cinder/audio/cocoa/CinderCoreAudio.h
@@ -60,7 +60,7 @@ inline void copyToBufferList( ::AudioBufferList *bufferList, const Buffer *buffe
 		memcpy( bufferList->mBuffers[i].mData, buffer->getChannel( i ), bufferList->mBuffers[i].mDataByteSize );
 }
 
-inline void copyToBufferList(::AudioBufferList *bufferList, const Buffer *buffer, size_t startFrame, size_t frameCount )
+inline void copyToBufferList( ::AudioBufferList *bufferList, const Buffer *buffer, size_t startFrame, size_t frameCount )
 {
 	for( UInt32 i = 0; i < bufferList->mNumberBuffers; i++ ) {
 		Float32 *startPtr = static_cast<Float32 *>( bufferList->mBuffers[i].mData ) + startFrame;
--- a/include/cinder/audio/cocoa/CinderCoreAudio.h
+++ b/include/cinder/audio/cocoa/CinderCoreAudio.h
@@ -60,6 +60,14 @@ inline void copyToBufferList( ::AudioBufferList *bufferList, const Buffer *buffe
 		memcpy( bufferList->mBuffers[i].mData, buffer->getChannel( i ), bufferList->mBuffers[i].mDataByteSize );
 }
 
+inline void copyToBufferList(::AudioBufferList *bufferList, const Buffer *buffer, size_t startFrame, size_t frameCount )
+{
+	for( UInt32 i = 0; i < bufferList->mNumberBuffers; i++ ) {
+		Float32 *startPtr = static_cast<Float32 *>( bufferList->mBuffers[i].mData ) + startFrame;
+		memcpy( startPtr, buffer->getChannel( i ), frameCount * sizeof( Float32 ) );
+	}
+}
+
 inline void copyFromBufferList( Buffer *buffer, const ::AudioBufferList *bufferList )
 {
 	for( UInt32 i = 0; i < bufferList->mNumberBuffers; i++ )
@@ -70,6 +78,14 @@ inline void zeroBufferList( const ::AudioBufferList *bufferList )
 {
 	for( UInt32 i = 0; i < bufferList->mNumberBuffers; i++ )
 		memset( bufferList->mBuffers[i].mData, 0, bufferList->mBuffers[i].mDataByteSize );
+}
+
+inline void zeroBufferList( const ::AudioBufferList *bufferList, size_t startFrame, size_t frameCount )
+{
+	for( UInt32 i = 0; i < bufferList->mNumberBuffers; i++ ) {
+		Float32 *startPtr = static_cast<Float32 *>( bufferList->mBuffers[i].mData ) + startFrame;
+		memset( startPtr, 0, frameCount * sizeof( Float32 ) );
+	}
 }
 
 // ----------------------------------------------------------------------------------------------------
--- a/src/cinder/audio/OutputNode.cpp
+++ b/src/cinder/audio/OutputNode.cpp
@@ -96,7 +96,7 @@ OutputDeviceNode::OutputDeviceNode( const DeviceRef &device, const Format &forma
 	mDidChangeConn = mDevice->getSignalParamsDidChange().connect( bind( &OutputDeviceNode::deviceParamsDidChange, this ) );
 
 	mInterruptionBeganConn = Context::deviceManager()->getSignalInterruptionBegan().connect( [this] { disable(); } );
-	mInterruptionBeganConn = Context::deviceManager()->getSignalInterruptionEnded().connect( [this] { enable(); } );
+	mInterruptionEndedConn = Context::deviceManager()->getSignalInterruptionEnded().connect( [this]( bool shouldResume ) { enable(); } );
 
 	size_t deviceNumChannels = mDevice->getNumOutputChannels();
 
--- a/src/cinder/audio/OutputNode.cpp
+++ b/src/cinder/audio/OutputNode.cpp
@@ -95,6 +95,9 @@ OutputDeviceNode::OutputDeviceNode( const DeviceRef &device, const Format &forma
 	mWillChangeConn = mDevice->getSignalParamsWillChange().connect( bind( &OutputDeviceNode::deviceParamsWillChange, this ) );
 	mDidChangeConn = mDevice->getSignalParamsDidChange().connect( bind( &OutputDeviceNode::deviceParamsDidChange, this ) );
 
+	mInterruptionBeganConn = Context::deviceManager()->getSignalInterruptionBegan().connect( [this] { disable(); } );
+	mInterruptionBeganConn = Context::deviceManager()->getSignalInterruptionEnded().connect( [this] { enable(); } );
+
 	size_t deviceNumChannels = mDevice->getNumOutputChannels();
 
 	// If number of channels hasn't been specified, default to 2 (or 1 if that is all that is available).
--- a/src/cinder/audio/OutputNode.cpp
+++ b/src/cinder/audio/OutputNode.cpp
@@ -26,6 +26,8 @@
 #include "cinder/audio/Utilities.h"
 #include "cinder/audio/Exception.h"
 
+#include "cinder/app/App.h"
+
 #include <string>
 
 using namespace std;
@@ -96,7 +98,21 @@ OutputDeviceNode::OutputDeviceNode( const DeviceRef &device, const Format &forma
 	mDidChangeConn = mDevice->getSignalParamsDidChange().connect( bind( &OutputDeviceNode::deviceParamsDidChange, this ) );
 
 	mInterruptionBeganConn = Context::deviceManager()->getSignalInterruptionBegan().connect( [this] { disable(); } );
-	mInterruptionEndedConn = Context::deviceManager()->getSignalInterruptionEnded().connect( [this]( bool shouldResume ) { enable(); } );
+	mInterruptionEndedConn = Context::deviceManager()->getSignalInterruptionEnded().connect( [this]( bool resumeImmediately ) {
+		if( resumeImmediately ) {
+			enable();
+		}
+		else {
+			// Resume output once the app becomes active again.
+			auto app = app::App::get();
+			if( app ) {
+				mAppDidBecomeActiveConn = app->getSignalDidBecomeActive().connect( [this] {
+					mAppDidBecomeActiveConn.disconnect();
+					enable();
+				} );
+			}
+		}
+	} );
 
 	size_t deviceNumChannels = mDevice->getNumOutputChannels();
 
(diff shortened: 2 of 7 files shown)


src/cinder/audio/cocoa/DeviceManagerAudioSession.mm
```diff
@@ -24,61 +24,55 @@
 #include "cinder/audio/cocoa/DeviceManagerAudioSession.h"
 #include "cinder/audio/Exception.h"
 #include "cinder/CinderAssert.h"
+#include "cinder/Log.h"
```
notlion: Remove unused include

src/cinder/audio/cocoa/DeviceManagerAudioSession.mm
```diff
@@ -219,19 +233,22 @@ - (void)notifyInterrupted:(NSNotification *)notification;
 
 void DeviceManagerAudioSession::activateSession()
 {
-	AVAudioSession *globalSession = [AVAudioSession sharedInstance];
+	CI_LOG_V("Activating Session");
```
notlion: Remove stray log

src/cinder/audio/cocoa/DeviceManagerAudioSession.mm
```diff
@@ -219,19 +233,22 @@ - (void)notifyInterrupted:(NSNotification *)notification;
...
+
+	AVAudioSession *globalSession = [AVAudioSession sharedInstance];
 
 	NSError *error = nil;
 	bool didActivate = [globalSession setActive:YES error:&error];
-	CI_ASSERT( ! error );
+	CI_ASSERT( !error );
 
-	if( ! didActivate )
+	if( !didActivate )
```
notlion: Fix bang formatting

include/cinder/audio/cocoa/CinderCoreAudio.h
```diff
@@ -60,6 +60,14 @@ inline void copyToBufferList( ::AudioBufferList *bufferList, const Buffer *buffe
 		memcpy( bufferList->mBuffers[i].mData, buffer->getChannel( i ), bufferList->mBuffers[i].mDataByteSize );
 }
 
+inline void copyToBufferList(::AudioBufferList *bufferList, const Buffer *buffer, size_t startFrame, size_t frameCount )
```
notlion: Add space after open paren

notlion:
It seems there is still a bit of work to do on this PR. `AudioOutputUnitStart` errors when enabling the OutputDeviceNode after receiving, accepting, and then hanging up a phone call. Also, audio is not resumed when switching to, and then back from another app that plays audio.

I need to do a bit more investigating into what exactly the `AVAudioSessionInterruptionOptionShouldResume` option means.

notlion:
Apple says [here](https://developer.apple.com/reference/avfoundation/avaudiosessioninterruptionoptions/avaudiosessioninterruptionoptionshouldresume?language=objc) that "Apps that do not require user input to begin audio playback (such as games) can ignore this flag and always resume playback when an interruption ends." So perhaps always re-enabling is the correct thing to do, but we should still pass this flag through for apps where it makes sense to have play/pause functionality.

notlion:
So, one issue with interruption remains. If a phone call is declined, the interruption ended notification will fire while the app is still in the background. If background audio is disabled (this is the default case, when audio input has not been disabled and the AVAudioSession category is `AVAudioSessionCategoryAmbient`) the call to `AudioOutputUnitStart` will fail with the error `AUIOClient_StartIO failed (561015905)`.

I've found a solution to this, which is to wait until `signalDidBecomeActive` to re-enable the output. This works, but if background audio is enabled the correct course of action is to re-enable the output immediately when the interruption ended notification is received.

We don't currently (AFAIK) have a way of knowing if background audio is enabled in the context, since this is only exposed to the platform specific DeviceManager via the AVAudioSession category, and we can't simply check if input is enabled because certain apps (like mine) will want to enable background audio without enabling input.

So.. there's a couple tricky decisions here.. Do we want to expose specific capabilities (like background audio enablement) to the audio context? Also, how best would we queue enabling of the output node until didBecomeActive when background audio is not allowed? We could reach for the global app instance and connect to the signal there, but it seems a bit spaghetti-ish.. My feeling is that this should be handled internally by the DeviceManager.

richardeakin:
Here's some thoughts after reading your research:

First on the `shouldResume` bool flag, I'm not sure of any reason we'd want to expose this to the user. Either a) you always resume upon interruption ended or b) you use it as an indication to wait for something else to happen and then resume. The flag doesn't help here, in all the use cases I can think of.

I started off writing that I thought the user should have to re-enable the context directly, but the truth of the matter is that any hardware-based audio api needs to be aware of the current app for reasons like this. The `audio::Context` already needs to listen to when the app is going to shut down so that it can stop the audio thread ahead of time (else possibly crash). We will, however, want to check that the app exists before using it as there are cases (like plug-ins) where it will be a null pointer.

That said, it would be pretty nice if DeviceManagerAudioSession could handle when audio needs to be resumed internally, by checking the session category. If it is currently `AVAudioSessionCategoryAmbient`, then it knows that it can call `signalInterruptionEnded` immediately. If it isn't, then it should wait until `signalDidBecomeActive` has fired and then only call `signalInterruptionEnded` after that.  To me, this would give ci::audio users a 'just works' experience. We wouldn't handle anything like ramping in the gain, they'll have to do that on their own (and can do so by registering with the signals).

If needed (don't think we do right now), we can add something like `DeviceManager::setResumeAudioAfterInterruptionEnabeld()` so users can disable this auto-resume functionality. Users would then use the app and DeviceManager signals directly to handle when audio is resumed.

If the current session is `AVAudioSessionCategoryAmbient`, then `DeviceManagerAudioSession` can pass a true value through to the interruption ended signal to indicate that audio should be immediately resumed.

By the way, currently the only way to manually set the audio session is to use the obj-c AVAudioSession API directly. If we were going to expose the audio session it could be done directly via DeviceManagerAudioSession (would save you from having to compile your source as obj-c++):

```
auto deviceManager = dynamic_cast<audio::cocoa::DeviceManagerAudioSession *>( audio::master->getDeviceManager() );
deviceManager->setAudioSessionCategory( "AVAudioSessionCategoryAmbient" );

```

We could also think about a more generic 'setBackgroundAudioEnabled()`, but I don't yet know how this will pan out on other platforms and don't have the time to work on it personally. I think either of these features could be done separately though, as this PR already has some serious bugfixes that will be great to get into master branch.

notlion:
Well I've been trying to get my iPhone into a state where an interruption ends but the resume flag is not set.. and haven't been able to. Seems like when you aren't meant to resume, you just don't get the interruption ended notification. There's probably some subtlety here that I'm not understanding.

Anyway, always resuming on interruption end seems like an ok course of action until that subtlety reveals itself..

Your suggestion to handle resuming after an interruption via a `resumeImmediately` bool passed to the signal seems good. I'll swap that in.

As for setting the category. I was going to do it an a platform specific .mm file. The `AVAudioSession` instance is globally accessible, plus there are a couple different methods you can use to set the category.. Seems like there is marginal value in exposing that.

notlion:
Interruptions are now working with that last commit. I've tested with `PlayAndRecord` and `Ambient` categories. Audio now resumes correctly and without error after phone calls, Siri, and timers.

One thing that might be an issue.. I had to include `UIApplication` to get the application state [here](https://github.com/cinder/Cinder/pull/1585/commits/3f5d1e373ef4983c691a94c16e760936d65ae26f#diff-4aa1d0d952cf77dcbd8f0c984ed12fe0R276). This seems to work, and I'm not sure there's ever a case where there is no UIApplication instance on iOS.. Perhaps someone else knows more about this?

src/cinder/audio/OutputNode.cpp
```diff
@@ -96,7 +98,21 @@ OutputDeviceNode::OutputDeviceNode( const DeviceRef &device, const Format &forma
 	mDidChangeConn = mDevice->getSignalParamsDidChange().connect( bind( &OutputDeviceNode::deviceParamsDidChange, this ) );
 
 	mInterruptionBeganConn = Context::deviceManager()->getSignalInterruptionBegan().connect( [this] { disable(); } );
-	mInterruptionEndedConn = Context::deviceManager()->getSignalInterruptionEnded().connect( [this]( bool shouldResume ) { enable(); } );
+	mInterruptionEndedConn = Context::deviceManager()->getSignalInterruptionEnded().connect( [this]( bool resumeImmediately ) {
```
richardeakin: I was imagining that the `DeviceManagerAudioSession` would handle this logic internally, and there would be no need to pass the `resumeImmediately` flag at all. This app signal is really only needed on the iOS platform at current.

Also it seems a bit simpler to me if `DeviceManagerAudioSession` always registered for the `signalDidBecomeActive` and then when that gets called, it checks some sort of flag like `mIsCurrentlyInterrupted` or something. If it is, then it will call the `getSignalInterruptionEnded().emit()` upon app becoming active again.

src/cinder/audio/cocoa/DeviceManagerAudioSession.mm
```diff
@@ -264,18 +264,21 @@ @implementation AudioSessionNotificationHandlerImpl
...
-		NSUInteger interruptionOption = [[notification.userInfo valueForKey:AVAudioSessionInterruptionOptionKey] unsignedIntValue];
-		if( interruptionOption == AVAudioSessionInterruptionOptionShouldResume ) {
-			mDeviceManager->endInterruption(true);
-		} else {
-			mDeviceManager->endInterruption(false);
-		}
+		UIApplication *app = [UIApplication sharedApplication];
+		AVAudioSession *session = [AVAudioSession sharedInstance];
+
+		auto resumeImmediately = app.applicationState == UIApplicationStateActive ||
```
richardeakin: Would this be necessary if `DeviceManagerAudioSession` was registered for the `signalDidBecomeActive`?  If it still is, doesn't seem like a huge deal but perhaps it is avoidable that way.

src/cinder/audio/cocoa/DeviceManagerAudioSession.mm
```diff
@@ -219,17 +227,18 @@ - (void)notifyInterrupted:(NSNotification *)notification;
...
-#endif
+		[[NSNotificationCenter defaultCenter] addObserver:mSessionNotificationHandler selector:@selector( handleInterruption: ) name:AVAudioSessionInterruptionNotification object:nil];
+	}
+
+	AVAudioSession *globalSession = [AVAudioSession sharedInstance];
 
 	NSError *error = nil;
 	bool didActivate = [globalSession setActive:YES error:&error];
-	CI_ASSERT( ! error );
+	CI_ASSERT( !error );
```
richardeakin: space should remain between `!` and `error`

src/cinder/audio/cocoa/DeviceManagerAudioSession.mm
notlion: Hmm.. Well I assume we would still need to check if the UIApplication is active when we connect to `signalDidBecomeActive`, and keep track of that variable ourselves.. Seems simpler to just get it through the `sharedInstance` since we have to assume that exists anyway.. Of course if it's guaranteed that an application will be active when the DeviceManager is created we wouldn't need to check.. But since it's lazy loaded, we don't know.

src/cinder/audio/OutputNode.cpp
notlion: Hmm.. So if we can rely that the application is instantiated before DeviceManager, and doesn't ever change, then we could just connect to the `signalDidBecomeActive` signal once.

Removing `resumeImmediately` from the API would certainly make app code simpler.. Currently I need another `signalDidBecomeActive` listener in my app, just to fade the audio up.. Would be simpler to just wait to emit `signalInterruptionEnded`.

src/cinder/audio/cocoa/DeviceManagerAudioSession.mm
notlion: I wish ClangFormat had an option for this 😢

src/cinder/audio/OutputNode.cpp
richardeakin: The only time there won't be an App will be during `prepareSettings`, I think it is fair to say that you can't be manipulating the audio graph until after at least your app's constructor has been called. So I think just registering for the app became active signal once from the DeviceManagerAudioSession's constructor is OK. Actually I'm already doing the same for the shutdown signal.

I also like that user code becomes simpler this way, not having to worry about what that bool means and whether or not functionality should change based on it.

src/cinder/audio/cocoa/DeviceManagerAudioSession.mm
richardeakin: I was imagining that DeviceManagerAudioSession kept an `mWasInterrupted` bool, and then in the case of an ambient session category, if this is true during its app activated signal callback then it would call the interruption ended signal (in the regular case of a play-and-record session category, this flag would have already been set back to false as the interruption ended callback was immediately called when the obj-c notification fired).

PM me if that is confusing, this is super close so happy to hash it out and get it into master branch.

src/cinder/audio/OutputNode.cpp
notlion: Ah. I think I understand what you mean. In my implementation the user could call `endInterruption` manually, and that could result in a double-call if the `mEndInterruptionOnceAppIsActive` was set.

notlion:
I've prefixed the interruption methods with `private` in 02b4185. This seemed a little clearer to me than a postfix `Impl`. I also simplified the interface, removing `endInterruptionOnceAppIsActive` in favor of passing `resumeImmediately` directly to `privateEndInterruption`.

PR #1586, opened 2016-09-23
Updated VboMesh functionality to allow streaming.

ryanbartley:
- Added setters for num indices and vertices 
- changed logic for drawImpls to rely on the indices pointer rather than the number of indices


richardeakin:
Is it safe to update `mNumIndices` / `mNumVertices` without actually changing the size of the internal `Vbo`s? What if there is no index buffer but you unwittingly set `mNumIndices`?

ryanbartley:
It is not safe to draw more vertices or indices that you have in the corresponding vbos. If you change the number of mNumIndices without an index vbo that value will be ignored.

PR #1590, opened 2016-09-26
Updating stb to fix jpeg problem

ryanbartley:
- fixes problem found with decoding jpg images.


ryanbartley:
This fixes a problem loading certain jpgs from the internet. Also, adds output from failure_reason to the ImageIoException. Confirmed works.

PR #1591, opened 2016-09-27
Updated stb_image* version to account for problematic jpg decompression

ryanbartley:


ryanbartley:
This should be ready to merge.

PR #1592, opened 2016-09-27
Cmake osc tuio build fix

patrickFuerst:
This fixes _unresolved linker errors_ for OSC and TUIO examples. As well as a typo fix to build the TuioListener example.


richardeakin:
Thanks for fixing this up..

PR #1593, opened 2016-09-28
(macosx) Fix window pos when on secondary display

richardeakin:
This is in addition to #1489 to fix the window position during the app's initial resize() call.

~~The position still seems to be off during `App::setup()`, I'm not sure yet if that can be fixed.~~ fixed with 1cbee56.

Test app code: http://pastebin.com/BReFbN2C


richardeakin:
I should also note that a side effect of this is that the windowMove signal fires at startup now, whereas it didn't before. It could be prevented if needed, but maybe it should be called once at startup?

richardeakin:
Update: fixed the so that it is consistent during initial setup(), resize(), and basically any time after `applicationDidFinishLaunching` has been called.

PR #1594, opened 2016-10-01
Added CMake support for Visual Studio/Xcode/Android generator

chaoticbob:
Updated Windows platform specific CMake related files to build on Windows. Corrected some of the additions of preprocessor definitions. Updated ci_make_app macro to generate the right settings for apps. Tested against Cube, CubeMapping, ShadowMapping, and Extrude on Visual Studio 2013 and Visual Studio 2015.


CMakeLists.txt
```diff
@@ -1,6 +1,15 @@
 cmake_minimum_required( VERSION 3.0 FATAL_ERROR )
 set( CMAKE_VERBOSE_MAKEFILE ON )
 
+# Configure the Android toolchain before the project start
+if( CINDER_TARGET )
```
richardeakin: I think this should be a part of [configure.cmake](https://github.com/cinder/Cinder/blob/524c705c36adbf9644209d2210dc7dec26685f98/proj/cmake/configure.cmake), which sets up the `CINDER_TARGET` along with other important variables used on all platforms.

But perhaps the `cinderAndreoidToolchain.cmake` __must__ be included before the call to `project( cinder )`? If that is the case, I think that the `project()` call can be moved to after the `include( configure.cmake )`.

proj/cmake/modules/cinderMakeApp.cmake
```diff
@@ -61,12 +61,42 @@ function( ci_make_app )
...
+					string( REGEX REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}" )
+					set( "${flag_var}" "${${flag_var}}" PARENT_SCOPE )
+				endif()
+			endforeach()
+			# Force synchronous PDB writes
+			add_compile_options( /FS ) 
+			# Force multiprocess compilation
+			add_compile_options( /MP )
+			# Add necessary platform libraries		
+			list( APPEND PLATFORM_LIBRARIES "zlib.lib;shlwapi.lib;OpenGL32.lib" )
```
richardeakin: On OS X, we handled this as part of libcinder's `target_link_libraries()` ([here](https://github.com/cinder/Cinder/blob/524c705c36adbf9644209d2210dc7dec26685f98/proj/cmake/platform_macosx.cmake#L87-L97)) Its pretty nice this way because then anyone who does a `find_package( cinder )` will get linked to the required libraries, even if they aren't using `ci_make_app()` (like if they already have a different cmake setup).

proj/cmake/modules/cinderMakeApp.cmake
```diff
@@ -61,12 +61,42 @@ function( ci_make_app )
...
+			# Add lib dirs
+			cmake_policy( PUSH )
+			cmake_policy( SET CMP0015 OLD )
+			link_directories( "${CINDER_PATH}/${CINDER_LIB_DIRECTORY}" )
```
richardeakin: I think in general @PetrosKataras was advocating that we use the `target_*` variants as much as possible.  Is there a way to do this with `link_directories()` too?

CMakeLists.txt
chaoticbob: The compiler checks seem to happen when you ```project()``` is called. I think on most platform this is fine, but on Android since we have to do a manual toolchain setup, it needs to happen before ```project()```. Certainly not married to this solution, so if we can move stuff around to make it more sane, would be great.

proj/cmake/modules/cinderMakeApp.cmake
chaoticbob: I'll try moving it to ```target_link_libraries``` - I can't remember right now if there was a specific reason for this.

proj/cmake/modules/cinderMakeApp.cmake
chaoticbob: I looked around for a bit and could not find a way. Happy to change if you guys have a suggestion.

PetrosKataras: There is not `target_*` variant for `link_directories` but what I am actually wondering is if this is needed at all. In theory ( and in practice as it happens for Linux / OS X for example ) the previous call to `find_package( cinder )` together with the `target_link_libraries( myapp cinder )` that follows should take care of finding and linking with Cinder without the need for manually specifying link directories.. But maybe that is not the case for the Visual Studio generator?

chaoticbob: I don't think this needs to be a fine grain about theory and practice across the various platforms. CMake has historically had issues with the rules imposed by MSBuild and build systems like Jam that it has to do some weird jumping around. For instance the output path or the ignoring specific library. ```link_directories``` make it easier to deal with things like this: ```libboost_filesystem-vc120-mt-sgd-1_60.lib```. This leverages the auto linking feature in Visual C++ and doesn't lock you into a specific build of Boost - especially one that's tagged with a version number. In the near future we won't have to deal with this because we won't have a dependency on Boost. But for now, short of attempting to detect the version of Boost, it's easier to let Visual C++ handle it for you by using ```link_directories```.

proj/cmake/modules/cinderMakeApp.cmake
chaoticbob: Moved this to a more explicit location: https://github.com/cinder/Cinder/pull/1594/commits/76f878111d0cb0a093d3f9dee4d9dbfedd466111

proj/cmake/modules/cinderMakeApp.cmake
chaoticbob: Just noticed that some static libs on Windows can be added into the cinder.lib and we don't have to explicitly add them to the app target. Let me try moving some of this around to see if it works.

chaoticbob: Okay, moved what I could so ```find_package( cinder )``` picks up on the deps.

proj/cmake/libcinder_target.cmake
```diff
@@ -24,11 +24,12 @@ target_include_directories( cinder SYSTEM BEFORE INTERFACE ${CINDER_INCLUDE_SYST
 target_include_directories( cinder BEFORE PRIVATE ${CINDER_INCLUDE_USER_PRIVATE} )
 target_include_directories( cinder SYSTEM BEFORE PRIVATE ${CINDER_INCLUDE_SYSTEM_PRIVATE} )
 
-target_link_libraries( cinder PUBLIC ${CINDER_LIBS_DEPENDS} )
+target_link_libraries( cinder PUBLIC ${CINDER_LIBS_DEPENDS}  )
+target_link_libraries( cinder LINK_PRIVATE Ws2_32.lib  )
```
PetrosKataras: I think this line should be removed since a) it would 'leak' to non-windows platforms and b) it has been integrated in `platform_msw.cmake` as part of `CINDER_LIBS_DEPENDS` [here](https://github.com/cinder/Cinder/pull/1594/files/ead79ef496ef382b417fe9ffe239afcc57c358d3..e8d2c322b5e3332e4004358a4ee6808b8b725716#diff-034b9f3e1e4aa1f6f60463fa1c03ba8aR65)?

proj/cmake/modules/cinderMakeApp.cmake
PetrosKataras: Personally I don't have a strong opinion regarding `link_directories` as long as we can keep the usage of it just for edge cases like this one. I just look forward to moving away from Boost as a dependency..

CMakeLists.txt
PetrosKataras: I seems that the proper ( in terms of CMake ) way to do this would be using CMake's toolchain functionality. See [here](http://www.vtk.org/Wiki/CMake_Cross_Compiling) and [here](http://www.vtk.org/Wiki/CMake_Cross_Compiling#The_toolchain_file). This functionality would also be interesting to explore in the case of Linux / RPi .

proj/cmake/libcinder_target.cmake
chaoticbob: Sorry, that was apart of my experimenting to find some better way to propagate the link flags. It's been removed now.

proj/cmake/modules/cinderMakeApp.cmake
chaoticbob: So I actually implemented the globbing and it worked. But then I realized that it would make it really difficult for the CMake generated lib to interact with the handmade project files. So it seems like ```link_directories()``` is the most reasonable thing to do for now.

proj/cmake/configure.cmake
```diff
@@ -103,14 +103,17 @@ if( CINDER_LINUX )
 elseif( CINDER_MAC )
 	set( CINDER_TARGET_SUBFOLDER "macosx" )
 elseif( CINDER_ANDROID )
-	set( CINDER_ANDROID_NDK_PLATFORM 21 CACHE STRING "Android NDK Platform version number." )
-	set( CINDER_ANDROID_NDK_ARCH "armeabi-v7a" CACHE STRING "Android NDK target architecture." )
-
-	set( CINDER_TARGET_SUBFOLDER "android-${CINDER_ANDROID_NDK_PLATFORM/CINDER_ANDROID_NDK_ARCH" )
+	#set( CINDER_ANDROID_NDK_PLATFORM 21 CACHE STRING "Android NDK Platform version number." )
```
richardeakin: Don't we want to expose these variables to the cmake cache somehow?

Also, I do like that they are in `configure.cmake` as this file is shared by both libcinder and app builds.

proj/cmake/libcinder_target.cmake
```diff
@@ -24,19 +24,83 @@ target_include_directories( cinder SYSTEM BEFORE INTERFACE ${CINDER_INCLUDE_SYST
...
 target_include_directories( cinder SYSTEM BEFORE PRIVATE ${CINDER_INCLUDE_SYSTEM_PRIVATE} )
 
-target_link_libraries( cinder PUBLIC ${CINDER_LIBS_DEPENDS} )
+target_link_libraries( cinder PUBLIC ${CINDER_LIBS_DEPENDS}  )
 
 target_compile_definitions( cinder PUBLIC ${CINDER_DEFINES} )
 
+# Visual Studio and Xcode generators adds a ${CMAKE_BUILD_TYPE} to the ARCHIVE 
+# and LIBRARY directories. Override the directories so, ${CMAKE_BUILD_TYPE} doesn't double up.
+if( CINDER_MSW )
```
richardeakin: Can this live in platform_msw.cmake?

proj/cmake/libcinder_target.cmake
```diff
@@ -24,19 +24,83 @@ target_include_directories( cinder SYSTEM BEFORE INTERFACE ${CINDER_INCLUDE_SYST
...
+        LIBRARY_OUTPUT_DIRECTORY_DEBUG			"${OUTPUT_DIRECTORY_BASE}/Debug/${PLATFORM_TOOLSET}|"
+        LIBRARY_OUTPUT_DIRECTORY_RELEASE		"${OUTPUT_DIRECTORY_BASE}/Release/${PLATFORM_TOOLSET}"
+        LIBRARY_OUTPUT_DIRECTORY_MINSIZEREL		"${OUTPUT_DIRECTORY_BASE}/MinSizeRel/${PLATFORM_TOOLSET}"
+        LIBRARY_OUTPUT_DIRECTORY_RELWITHDEBINFO	"${OUTPUT_DIRECTORY_BASE}/RelWithDebInfo/${PLATFORM_TOOLSET}"
+        STATIC_LIBRARY_FLAGS_DEBUG				"${CINDER_STATIC_LIBS_FLAGS_DEBUG} ${CINDER_STATIC_LIBS_DEPENDS_DEBUG}" 
+        STATIC_LIBRARY_FLAGS_RELEASE			"${CINDER_STATIC_LIBS_FLAGS_RELEASE} ${CINDER_STATIC_LIBS_DEPENDS_RELEASE}"
+        STATIC_LIBRARY_FLAGS_MINSIZEREL			"${CINDER_STATIC_LIBS_FLAGS_RELEASE} ${CINDER_STATIC_LIBS_DEPENDS_RELEASE}"
+        STATIC_LIBRARY_FLAGS_RELWITHDEBINFO		"${CINDER_STATIC_LIBS_FLAGS_RELEASE} ${CINDER_STATIC_LIBS_DEPENDS_RELEASE}" 
+    )    
+elseif( CINDER_MAC )
```
richardeakin: Can this live in platform_macosx.cmake?

proj/cmake/modules/cinderMakeApp.cmake
```diff
@@ -61,13 +65,41 @@ function( ci_make_app )
...
 	endif()
 
 	add_executable( ${ARG_APP_NAME} MACOSX_BUNDLE WIN32 ${ARG_SOURCES} ${ICON_PATH} ${ARG_RESOURCES} )
 
 	target_include_directories( ${ARG_APP_NAME} PUBLIC ${ARG_INCLUDES} )
 	target_link_libraries( ${ARG_APP_NAME} cinder ${ARG_LIBRARIES} )
 
+	# Ignore Specific Default Libraries
+	if( MSVC )
+		set_target_properties( ${ARG_APP_NAME} PROPERTIES LINK_FLAGS "/NODEFAULTLIB:LIBCMT /NODEFAULTLIB:LIBCPMT" )
```
richardeakin: If we set this on the cinder target, won't they get picked up as the app depends on that package? That way we can keep them grouped together.

proj/cmake/platform_macosx.cmake
```diff
@@ -114,6 +114,24 @@ source_group( "cinder\\cocoa"           FILES ${SRC_SET_COCOA} )
 source_group( "cinder\\app\\cocoa"      FILES ${SRC_SET_APP_COCOA} )
 source_group( "cinder\\audio\\cocoa"    FILES ${SRC_SET_AUDIO_COCOA} )
 
+set( MACOS_SUBFOLDER            "${CINDER_PATH}/lib/${CINDER_TARGET_SUBFOLDER}" )
+set( CINDER_STATIC_LIBS_DEPENDS "${MACOS_SUBFOLDER}/libboost_filesystem.a ${MACOS_SUBFOLDER}/libboost_system.a ${MACOS_SUBFOLDER}/libz.a" )
+
+if( NOT ( "Xcode" STREQUAL "${CMAKE_GENERATOR}" ) )
+	if(NOT CMAKE_LIBTOOL)
```
richardeakin: Add comment as to why we need to use libtool instead of ar / ranlib (partially because that's what xcode does).

richardeakin:
Merging this as it is getting long and the remaining notes are mostly clean up, can be done afterwards.

PR #1600, opened 2016-10-08
Fix for MSW MessageBox crash.

paulhoux:
On my machine (running Windows 10), when I have a bug in my app's ```setup()``` function, Windows creates a ```MessageBox``` to show the error. This generates a ```WM_PAINT``` event which in turn calls the app's ```draw()``` function. Uninitialized member variables might then throw a new error, crashing the application. This is rather confusing. As a fix, I properly check if ```setup()``` has been called before calling ```draw()```. That way, the initial error message for which a ```MessageBox``` was created is shown, making debugging a lot easier.

### DIFF
--- a/src/cinder/app/msw/AppImplMsw.cpp
+++ b/src/cinder/app/msw/AppImplMsw.cpp
@@ -961,7 +961,8 @@ LRESULT CALLBACK WndProc(	HWND	mWnd,			// Handle For This Window
 		}
 		break;
 		case WM_PAINT:
-			impl->draw();
+			if( impl->getAppImpl()->setupHasBeenCalled() )
+				impl->draw();
 		break;
 		case WM_TOUCH:
 			impl->onTouch( mWnd, wParam, lParam );


paulhoux:
```
#include "cinder/app/App.h"
#include "cinder/gl/gl.h"
#include "cinder/BSpline.h"

using namespace ci;
using namespace ci::app;
using namespace std;

class ExampleApp : public App {
  public:
	void setup() override;
	void draw() override;

private:
	std::vector<vec2> mPoints;
	BSpline2f mSpline;
};

void ExampleApp ::setup()
{
        // This will generate an index out of range error:
	mPoints[0] = vec2( 100, 100 );
	mPoints[1] = vec2( 250, 200 );
	mPoints[2] = vec2( 500, 150 );

        // Spline gets properly defined only after the error.
	mSpline = BSpline2f( mPoints, 2, false, true );
}

void PracticalSDFApp::draw()
{
	gl::clear();

	gl::color( 1, 0.5f, 0 );
	gl::begin( GL_LINE_STRIP );
	for( int i = 0; i <= 100; ++i ) {
		float t = i * 0.01f;

                // Function is called on uninitialized member variable, generating a new error!
		gl::vertex( mSpline.getPosition( t ) );
	}
	gl::end();
}

CINDER_APP( PracticalSDFApp, RendererGl )
```

paulhoux:
Without the fix:

error

With the fix applied:

fix

ryanbartley:
I can confirm that I've had this exact same problem before but had no idea what it was caused by. Good find Paul!

andrewfb:
Thanks for the fix & test case

Issue #1599, opened 2016-10-08
Format speicifier error


MikeGitb:
When building the master branch of cinder with VS2015 U3 for x64 you get a ton of warnings, mostly about possibly loss of data. I haven't had time to sift through all of them, but one particular message that cought my eyes is this:

> 1>..\src\videoInput\videoInput.cpp(367): warning C4477: 'printf' : format string '%s' requires an argument of type 'char *', but variadic argument 1 has type 'WCHAR *'
1>  ..\src\videoInput\videoInput.cpp(367): note: consider using '%ls' in the format string
1>  ..\src\videoInput\videoInput.cpp(367): note: consider using '%lls' in the format string
1>  ..\src\videoInput\videoInput.cpp(367): note: consider using '%Ls' in the format string
1>  ..\src\videoInput\videoInput.cpp(367): note: consider using '%ws' in the format string

relating to the following piece of code

	WCHAR buffer[255];

        //....

	if(verbose)printf("SETUP: removing filter %S...\n", buffer);
	hr = pGraph->RemoveFilter(pFilter);
	if (FAILED(hr)) { if(verbose)printf("SETUP: pGraph->RemoveFilter() failed. \n"); return; }
	if(verbose)printf("SETUP: filter removed %s  \n",buffer); //<- warning here

I'm not familiar with the particularities of printf format specifiers in combination with wide strings, so I'm not sure what the best speciefer to put there is (capital `%S` seems to work too 3 lines above) or I would have offered a PR, but it seems like something that someone with a little experience could fix in a few seconds.

richardeakin:
I think PRs around cleaning up some warnings would be most welcome if you find the time.

MikeGitb:
@richardeakin: I'll see if I find any warnings, I can fix confidentially. Is there some CI and/or testsuite I can test my patches against? I don't have VS2013 or XCode.

richardeakin:
Thanks! We have started moving to unit tests (you can find the suite [here](https://github.com/cinder/Cinder/tree/master/test/unit)) but they aren't extensive. Up until now we've been testing with existing samples, or something in the test folder (sometimes a bit messy..).

MikeGitb:
As I'm new to this library (and Github in general): Those fixes will most likely span multipe different and otherwise unrelated files. Is it prefered to offer such refactorings as multiple small PRs (e.g. one per module and/or error type) or rather as one big one (with multiple commits)?

richardeakin:
[...]

PR #1604, opened 2016-10-10
Fix format string for wide string buffer

MikeGitb:
Fixes bug mentioned in #1599 and removes some unnecessary code in the sourrounding function. 

The actual fix is contained in the first commit (just replacing ` s`with `S`)
The second commit is removing an unnecessary temporary string buffer


richardeakin:
Thanks, both commits look good to me.

PR #1608, opened 2016-10-11
Updated Cairo with build scripts for installation across platforms

ryanbartley:
- Added Install [bash script](https://github.com/ryanbartley/Cinder/blob/CairoUpdate/blocks/Cairo/install/install.sh) for *nix platforms and a [batch script](https://github.com/ryanbartley/Cinder/blob/CairoUpdate/blocks/Cairo/install/install.bat) for windows to control version, as well as ease building and install of other versions
- updated cinderblock.xml for new path locations
- removed old binaries and headers to not conflict with new build scripts
- included a template for the CairoConfig.cmake, this will probably change depending on final chosen paths.
- updated zlib distribution


ryanbartley:
Closing as #1648 cleaned and flattened this up.

PR #1610, opened 2016-10-11
Handle mouse wheel events for Linux

patrickFuerst:
This adds scroll events to Linux Apps. 
Tested on Ubuntu 16.


andrewfb:
Thanks for this contribution.

PR #1612, opened 2016-10-11
simplify rand

MikeGitb:
This PR simplifies the implementation of `ci::Rand` and the free functions in Rand.h. Mainly by
- Replacing the OS-specific seeding methods by std::random_device
- Replacing the hand-roled gaussian distribution reneration functions by `std::normal_distribution`.

I've run a few microbenchmarks to verify that this doesn't degrade the performance of the random number generation on my particular platform (VS2015 U3, Win10), but I can revert a particular change if someone knows for a fact that the stanard library facilities were not used for a particular reason.

### DIFF
--- a/include/cinder/Rand.h
+++ b/include/cinder/Rand.h
@@ -38,7 +38,10 @@ class Rand {
 	{}
 
 	//! Re-seeds the random generator
-	void seed( uint32_t seedValue );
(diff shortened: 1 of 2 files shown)


andrewfb:
Thanks for these improvements

PR #1613, opened 2016-10-11
Refactoring private SurfaceCache class to make more legible

richardeakin:
Using a vector<vector<uint8_t>> instead of shared_ptr<uint8_t>, added
comments to explain the ownership logic.

Did this after having a difficult time understanding #1565, hopefully these comments will help the next poor soul who comes along.

### DIFF
--- a/src/cinder/CaptureImplDirectShow.cpp
+++ b/src/cinder/CaptureImplDirectShow.cpp
@@ -74,20 +74,27 @@ class SurfaceCache {
 	SurfaceCache( int32_t width, int32_t height, SurfaceChannelOrder sco, int numSurfaces )
 		: mWidth( width ), mHeight( height ), mSCO( sco )
 	{
+		// create the data blocks, which will be reused throughout the lifetime of the Capture instance.
 		for( int i = 0; i < numSurfaces; ++i ) {
-			mSurfaceData.push_back( std::shared_ptr<uint8_t>( new uint8_t[width*height*sco.getPixelInc()], std::default_delete<uint8_t[]>() ) );
+			mSurfaceData.emplace_back( width * height * sco.getPixelInc() );
 			mSurfaceUsed.push_back( false );
 		}
 	}
 	
 	Surface8uRef getNewSurface()
 	{
-		// try to find an available block of pixel data to wrap a surface around	
+		// Try to find an available block of pixel data to wrap a surface around. We create a non-owning
+		// Surface8u that points at the data block, passing this back to the user. When the last instance
+		// is destroyed, the custom deleter marks that slot as available.
 		for( size_t i = 0; i < mSurfaceData.size(); ++i ) {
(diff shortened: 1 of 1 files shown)


MikeGitb:
Looks good to me.

Just a thought: Why not have a cache of actual `Surface8uRef` objects and use the `std::shared_ptr::unique()` method to check whether the object is currently in use or not? That would further simplify `getNewSurface()` (down to about 3-4 lines) and a custom deleter wouldn't be necessary at all.

This could possibly even improve performance: Currently, there are two dynamic memory allocation each time `getNewSurface()` is called (one for the Surface8u and one for the shared_ptr)  and two de-allocations whenever the last shared_ptr for a particular surface goes out of scope. With an internal vector of `Surface8uRef` objects, one could use `std::make_shared` and have only a single allocation/deallocation per slot upon construction / destruction of the `SurfaceCache` object  and a call to `getNewSurface()` would just bump the ref counter.

src/cinder/CaptureImplDirectShow.cpp
```diff
@@ -74,20 +74,27 @@ class SurfaceCache {
 	SurfaceCache( int32_t width, int32_t height, SurfaceChannelOrder sco, int numSurfaces )
 		: mWidth( width ), mHeight( height ), mSCO( sco )
 	{
+		// create the data blocks, which will be reused throughout the lifetime of the Capture instance.
 		for( int i = 0; i < numSurfaces; ++i ) {
-			mSurfaceData.push_back( std::shared_ptr<uint8_t>( new uint8_t[width*height*sco.getPixelInc()], std::default_delete<uint8_t[]>() ) );
+			mSurfaceData.emplace_back( width * height * sco.getPixelInc() );
 			mSurfaceUsed.push_back( false );
 		}
```
MikeGitb: One could use reserve (for mSurfaceData) and resize (for mSurfaceUsed) here.

MikeGitb:
Of course my suggestion above is only possible, if you can actually reuse a `Surface` - I didn't check if a Surface can be efficiently reset.

richardeakin:
I like your suggestion, especially that it will remove the need for keeping an extra vector of bools around, and no deleting little objects each frame.  Would you like to have a go at implementing it?

By the way reusing a `Surface` is totally valid, you can see [here](https://github.com/cinder/Cinder/blob/d122f585ccd1437bef089401af2bf8f2fdbd1851/src/cinder/CaptureImplDirectShow.cpp#L192) where a pointer to its internal data storage is handed off to be filled by the `CaptureMgr` instance

MikeGitb:
I can give it a shot, but not before next weekend. So if someone else wants to work on this in the mean time, be my guest.

MikeGitb:
@richardeakin: Ok, I've created a PR where the surface cache simply reuses the Surfaces without reseting any data.

andrewfb:
This has been superseded by #1630 so closing now.

PR #1615, opened 2016-10-16
Distance to curve.

paulhoux:
With the current attention to text rendering and signed distance fields, I thought it would be a good idea to have proper functions in Cinder that will help us calculate distances from points to lines, quadratic curves and cubic curves. There was no functionality in `Path2d` or `Shape2d` for that yet.

In this PR, I have added proper, high quality algorithms to `CinderMath.h`. The `getClosestPointCubic()` method is quite involved, but compared to the implementation in Chlumsky's MSDFGEN library it is _between three and five times_ faster and _much_ more accurate. This will allow us to create much higher quality signed distance fields, without the need for additional libraries.

New methods in `Path2d` and `Shape2d` allow easy calculation of (signed) distances and/or finding the closest point to a shape or path.

To exercise the new functions, I have added a `ShapeTest` application to the test folder. It is similar to the `Triangulation` sample, but it focuses on distance calculations instead. 

Let me know what you think.

### DIFF
--- a/include/cinder/CinderMath.h
+++ b/include/cinder/CinderMath.h
@@ -289,6 +289,39 @@ int solveCubic( T a, T b, T c, T d, T result[3] );
 //! Returns the closest point to \a testPoint on the boundary of the ellipse defined by \a center, \a axisA and \a axisB. Algorithm due to David Eberly, http://www.geometrictools.com/Documentation/DistancePointEllipseEllipsoid.pdf
 glm::vec2 getClosestPointEllipse( const glm::vec2& center, const glm::vec2& axisA, const glm::vec2& axisB, const glm::vec2& testPoint );
 
+//! Returns the closest point to \a testPoint on the linear curve defined by the 2 \a controlPoints.
+template<typename T>
+glm::tvec2<T, glm::defaultp> getClosestPointLinear( const glm::tvec2<T, glm::defaultp> *controlPoints, const glm::tvec2<T, glm::defaultp> &testPoint );
+//! Returns the closest point to \a testPoint on the linear curve defined by the control points \a p0 and \a p1.
+template<typename T>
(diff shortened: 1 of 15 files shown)


paulhoux:
Per request of @chaoticbob , here is a comparison between MSDFGEN's algorithm and mine. 

comparison

paulhoux:
I was able to optimize the code a bit more and it now runs in 9.6secs compared to 15secs earlier.

MikeGitb:
I can't really comment on the shirts themselves or their performance at the moment, but as a tiny improvement suggestion: Wouldn't it make sense to turn the two for loops in Shape2.cpp into range based for loops? 
Also: would it be possible to write a few simple tests that could be checked automatically and added to the unit tests? In case anyone wants to refactor your code in the future?

paulhoux:
@MikeGitb  : I prefer using range based loops as well, but all other methods of ```Shape2d``` use the old iterator notation as well and I think it is better to choose one or the other, but not both. 

A unit test would be cool, but I have no experience with that. Also: the code has been tuned for performance and while I am sure things can still be optimized more, refactoring the code should be done with the same attention to performance as well. Small changes to the code can have a big impact. Currently, the slowest parts of the code are due to recursion and cache misses. But ~30 microseconds for a call to ```Shape2d::calcDistance()``` isn't bad.

richardeakin:
For reasons like this, I included the original SimpleSignals benchmarks ([here](https://github.com/cinder/Cinder/blob/master/test/SignalsTest/src/SignalsBenchmark.cpp)), that way we could monitor performance changes as we added features or refactored the `ci::signals` implementation.

I do think that this is a good case for both unit tests and benchmarks, along with any other geometric operations.

MikeGitb:
@paulhoux: I completely agree with your consistency argument. I didn't have a proper look at the whole file. On the long run it might be interesting to run clang-modernize over the code base to make that transition in a consistent manner.

PR #1617, opened 2016-10-18
Mhintz: Add WireRect

paulhoux:
Fixed the WireRect implementation provided by @mhintz  .


paulhoux:
Thanks for your contribution, Mark!

mhintz:
Ah thanks for grabbing this! It's been around the middle of my todo list to crack this open again for a while, never got around to it...

PR #1618, opened 2016-10-18
Updated Texture throughout to allow for passing dataformat

ryanbartley:
This pr allows the ability to set up fast path textures with dataType = GL_UNSIGNED_INT_8_8_8_8_REV and dataFormat = GL_BGRA. This form of texture format is not available without the changes made in the pr.


paulhoux:
Hey Ryan, this PR looks clean and serves a useful function, so you have my support. One thing I would like to say is that I, personally, have a hard time remembering what the difference is between data type and data format and so I often look at the comments to see examples of them. Could you add a few more comments, too?

ryanbartley:
Absolutely. I'll go back through tonight.

PR #1621, opened 2016-10-19
Avoid infinite loop if getnameinfo fails

### LINKED ISSUE #1619: System::getNetworkAdapters() get stuck in loop on OSX with VPN connection
I was trying to run the SystemTest unit test with a VPN connection active and the test got stuck in the "Network Adapters: " section. 

```c++
result = getnameinfo( currentInterface->ifa_netmask,
                           (currentInterface->ifa_addr->sa_family == AF_INET) ? sizeof(struct sockaddr_in) : sizeof(struct sockaddr_in6),
                           subnetMask, NI_MAXHOST, NULL, 0, NI_NUMERICHOST );
				if( result != 0 )
					continue;
				adapters.push_back( System::NetworkAdapter( currentInterface->ifa_name, host, subnetMask ) );
```
https://github.com/cinder/Cinder/blob/master/src/cinder/System.cpp#L650-L655

The problem seems to be that when it gets to `ifa_name` = "utun1" the result is 5. I made some edits figure out what that meant:
```c++
				if( result != 0 ) {
					fprintf(stderr, "getaddrinfo: %s\n", gai_strerror(result));
					continue;
				}
```
Which prints: `getaddrinfo: ai_family not supported`

The problem seems to be that simply calling continue doesn't advance the to the next interface.

drewish:
System::getNetworkAdapters() was getting stuck in an infinite loop with an active VPN connection on my machine. The problem was it was returning a nonzero value getting the subnet so we call continue but didn't advance to the next interface, so the failure just keeps repeating.

I fixed this by letting it fetch both values and then only adding it to the vector if both succeeded. 

The indenting changed a bit since there was a mix of tabs and spaces before. 

Fixes #1619

### DIFF
--- a/src/cinder/System.cpp
+++ b/src/cinder/System.cpp
@@ -638,24 +638,19 @@ vector<System::NetworkAdapter> System::getNetworkAdapters()
 
 	int success = getifaddrs( &interfaces );
 	if( success == 0 ) {
-		currentInterface = interfaces;
-		while( currentInterface ) {
+		for( currentInterface = interfaces; currentInterface; currentInterface = currentInterface->ifa_next ) {
 			if( currentInterface->ifa_addr->sa_family == AF_INET ) {
 				char host[NI_MAXHOST], subnetMask[NI_MAXHOST];
-				int result = getnameinfo( currentInterface->ifa_addr,
-                           (currentInterface->ifa_addr->sa_family == AF_INET) ? sizeof(struct sockaddr_in) : sizeof(struct sockaddr_in6),
-                           host, NI_MAXHOST, NULL, 0, NI_NUMERICHOST );
-				if( result != 0 )
-					continue;
-				result = getnameinfo( currentInterface->ifa_netmask,
-                           (currentInterface->ifa_addr->sa_family == AF_INET) ? sizeof(struct sockaddr_in) : sizeof(struct sockaddr_in6),
-                           subnetMask, NI_MAXHOST, NULL, 0, NI_NUMERICHOST );
-				if( result != 0 )
-					continue;
-				adapters.push_back( System::NetworkAdapter( currentInterface->ifa_name, host, subnetMask ) );
-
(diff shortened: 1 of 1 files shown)


MikeGitb:
Looks good to me. As a minor improvement: Why not turn this into a for-loop? That would seem most natural to me.

drewish:
A for loop does seem to tighten stuff up... It would make it easier to go back to the continue since it'll handle the advancement:
```c++
for( currentInterface = interfaces; currentInterface; currentInterface = currentInterface->ifa_next ) {
    if( currentInterface->ifa_addr->sa_family == AF_INET ) {
        char host[NI_MAXHOST], subnetMask[NI_MAXHOST];
        
        int result = getnameinfo( currentInterface->ifa_addr,
            (currentInterface->ifa_addr->sa_family == AF_INET) ? sizeof(struct sockaddr_in) : sizeof(struct sockaddr_in6),
            host, NI_MAXHOST, NULL, 0, NI_NUMERICHOST );
        if( result != 0 )
            continue;

        result = getnameinfo( currentInterface->ifa_netmask,
            (currentInterface->ifa_addr->sa_family == AF_INET) ? sizeof(struct sockaddr_in) : sizeof(struct sockaddr_in6),
            subnetMask, NI_MAXHOST, NULL, 0, NI_NUMERICHOST );
        if( result != 0 )
            continue;

        adapters.push_back( System::NetworkAdapter( currentInterface->ifa_name, host, subnetMask ) );
    }
}
```
That said I haven't test out this code.

drewish:
Went ahead and switched this over to the for loop. It seemed to make for the clearest code.

MikeGitb:
Sorry if I this sounds picky, but shouldn't the definition of `currentInterface` be pulled into the loop header?
Otherwise this looks pretty good to me, although I couldn't test it.

andrewfb:
Thanks for this - I merged with a tweak per @MikeGitb's suggestion.

drewish:
Thanks!

PR #1628, opened 2016-10-23
fix audio exception

MikeGitb:
Now the error code of `AudioExc` is actually set and the unused `mDescription` member (which is already present in the base class) is removed.


richardeakin:
Good catch!

PR #1630, opened 2016-10-23
refactor CaptureImplDirectShow

MikeGitb:
This PR strives to simplify the implementation of `CaptureImplDirectShow`.

The first commit refactors `SurfaceCache` as discussed in #1613, the following ones represent the individual refactoring steps that most notably lead to the complete removal of the CaptureMgr class, which turned out to be just a singleton wrapper around a `videoInput` object.


MikeGitb:
Of course I'd be happy to rebase and squash if you are satisfied with the result.

andrewfb:
I think this is well on its way to being a nice cleanup, so thanks for contributing it. It does change a key bit of behavior though, which is how a buffer "overflow" is handled. The previous version was designed to grow the buffer to accommodate the peak usage. The rationale is around a case where a user has cause to hold on to some number of frames. To be concrete, say an application always blends the last 10 frames from a webcam. In the previous design the cache would automatically grow to match this requirement. In this PR however, the `SurfaceCache` remains at size 4, and in such a scenario must constantly allocate (and deallocate) new `Surface`s.

I think a change to `SurfaceCache::getNewSurface()` to `push_back()` on `mSurfaces` and then return `mSurfaces.back()` rather than `return Surface8u::create(...)` would restore the previous behavior however.

MikeGitb:
@andrewfb: Thanks for having a look. I have to admit that I can't see the growth behavior you are describing in the original code. As far as I can see, neither `mSurfaceData` nor `mSurfaceUsed` grow after their initial construction. 
Of course, I'd be happy to implement the functionality you describe in any case.

andrewfb:
@MikeGitb Ah - it looks like you're right. Well, if we can (all) agree that would be preferable, let's go ahead and make that behavior change. Obviously the caching strategy that's appropriate is app-specific, but as a default, I think a conservative approach to grow to peak usage is probably the right choice, especially since the user in the general case won't be choosing the cache size. Thanks again for taking a look at this.

MikeGitb:
OK, I rebased, squashed and cleaned up the commits somewhat - I hope they are now easier to review. 
I added the the growth strategy as a separeate commit at the end, so you can easily choose which version you want.

andrewfb:
Thanks for your patience and work on this. This is a nice improvement in more than one regard.

PR #1636, opened 2016-10-25
CameraUI should clear connections when changing window

### LINKED ISSUE #1635: CameraUI doesn't empty mMouseConnections when reconnecting
If you call `CameraUi::connect()` multiple times adds 5 connections each time to `mMouseConnections` but never removes them.

drewish:
Seemed cleanest to just centralize this in `disconnect()`. 

Fixes #1635 

Side note but what do people think about renaming `mMouseConnections` to just `mConnections`? I think the current name is a bit misleading since it has the window resize connection in there too.

### DIFF
--- a/include/cinder/CameraUi.h
+++ b/include/cinder/CameraUi.h
@@ -100,7 +100,7 @@ class CameraUi {
 	app::WindowRef			mWindow;
 	bool					mEnabled;
 	int						mSignalPriority;
-	std::vector<ci::signals::Connection>	mMouseConnections;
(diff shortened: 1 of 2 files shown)


richardeakin:
Thanks for this fix. `mConnections` makes sense to me, I think it got its previous name before a vector was used.  Do you want to make that change and push it to this branch, before we merge?

drewish:
I'l stick another commit in here later tonight

drewish:
Made the name change. Yeah my assumption was that the purpose had changed but the naming didn't get updated to reflect it.

PR #1639, opened 2016-10-25
fix ignored params 3

MikeGitb:
More cases of parameters that are (probably) erroneously ignored


andrewfb:
Thanks very much for these finds & fixes

PR #1641, opened 2016-10-26
Added freetype build to xcode, visual studio, cmake projects

ryanbartley:
Added freetype headers/source to each platform, i.e. ios, ios-sim, macosx, windows (linux already uses it)


ryanbartley:
The conflicts introduced by adding zlib are ridiculous so I'm going to close this and add another in just a moment.

PR #1643, opened 2016-10-27
Fix forgotton info querey in getMonitorBitsPerPixel()

MikeGitb:
The MONITORINFOEX structure never got filled with data in this function.

Disclaimer:
This fix is based on immitating the neighboring function `getMonitorName()` and what I could put together from the MSDN documentation of the other involved functions - I've no actual experience with the windows API.


andrewfb:
Thanks for this find & fix. Tested & confirmed.

PR #1646, opened 2016-10-29
Fix early termination of glfw and add cleanup signal.

PetrosKataras:
This fixes cases where calls to glfw where issued after the library had been terminated ( test with the FlickrTestMultiThreaded sample ) and also adds the missing `cleanup()` signal before app termination.


richardeakin:
Looks good to me, and good call on the cleanup signal. Merging.

PR #1647, opened 2016-10-31
Updated zlib src, binary for macosx, winrt project

ryanbartley:
Updated the zlib library dependency from version 1.2.3 to 1.2.8


ryanbartley:
Also, added is cmake and VC project updates.

andrewfb:
Thanks for this.

PR #1648, opened 2016-11-01
Cairo Update

ryanbartley:
- Cleaned up original pr - #1608 
- build scripts for windows and *nix variant
- added test
- added readme
- added cmake block
- updated cinderblock


felixfaire:
+1

This seems very useful for cinder's 2D rendering abilities. Is there a reason why it hasnt been merged in yet?

richardeakin:
Not sure about on windows, but when I used cairo on Ubuntu it was very straightfoward to install it with `apt-get` and cmake package-control. On posix, I'd recommend using OS packaging systems rather than maintaining custom build scripts.

Issue #1629, opened 2016-10-23
Unused parameters


MikeGitb:
## Summary

On higher warning levels, compiling Cinder produces 100s of warning related to unused function parameters and I'd like to get rid of them. This issue is supposed to serve as a central point to discuss the general problem and specific cases, where the bes course of action is not clear.
## Explanation

When compiling with higher warning levels (e.g. /W4 on VS or -Wall on g++/clang) the compiler warns about unused parameters. In general, this is a good thing, as it might warn you that you e.g. forgot to implement a certain functionality, are using a local / member variable with a similar name instead of the parameter, have some other bug or simply that you should rethink the interface.

However, when I'm compiling the library from the master branch with VS2015 U3, I currently get around 680 of those warnings and most of them are harmless, which is a tremendous amount of noise. Even worse: A lot of those warnings alsow appear to the users of cinder when they include the header files (Just compiling the BasicApp example I results in 12 of them). 

If there is interest in this, I'm intending to resolve most of those warnings:
- In most cases, those parameters aren't used by design, e.g. because those functions implement a general interface, but the particular implementation doesn't need some of the parameters. In those cases, I'd just comment out the parameter names:
  
  ```
  int foo_old(int i) {  return42; }
  int foo_new(int /*i*/) { return 42; }
  ```
  
  This will never introduce a new bug, but it might "hide" a bug if ignoring the parameter wasn't actually intended.
- In some cases the function in question should really be a pure abstract function instead:
  
  ```
  virtual void foo_old(int i) {}
  virtual void foo_old(int i) = 0;
  ```
  
  In addition to removing the warning, this also makes the intent more clear and derived classes can not forget to override a function. This will however (obviously) cause build errors, if the function was actually meant to be a do-nothing-default-implementation.
- Some cases will be actual bugs (e.g. see #1628) and I can either offer a PR or raise a new issue
- Unclear cases could be discussed here

Priority wise I'd first target any warnings coming from header files, as those will create new warnings for every translation unit and also leak to the users of cinder.

andrewfb:
I'd say on the whole we're definitely interested in fixing these, and as you said, the highest priority warnings would be those that find their way into user apps. Frankly, we have a number of warnings that are triggered at lower warning levels too. In the constant triage of work on Cinder, these have not found their way to the top of the priority queue, but we'd definitely welcome this kind of contribution.

[...]

PR #1650, opened 2016-11-02
Comment unused function params in cinder/app

MikeGitb:
See #1629 for details.


andrewfb:
Thanks for these too, @MikeGitb

PR #1651, opened 2016-11-02
Comment more unused function parameters to prevent compiler warnings

MikeGitb:
See #1629 for details.


andrewfb:
Thanks for going after all of these (and the whitespace fixes), @MikeGitb

MikeGitb:
@andrewfb: The whitespace changes where actually unintended. I've configured my IDE to automatically delete trailing whitespaces, which is a good thing, if there aren't any to begin with, but adds some noise if there are. If those changes don't bother you, I can keep them in future PRs.

PR #1654, opened 2016-11-05
CMake cinderMakeApp fix

ryanbartley:
Fixed small problem with feeding Blocks as absolute paths to the Blocks argument of cinderMakeApp


richardeakin:
This is the 'absolute directory' option that was recently introduced, right?

Fix seems good to me, other than the formatting should be consistent.

ryanbartley:
ah, that's brutal. I'll fix it.

ryanbartley:
Alright, should be fixed now.

richardeakin:
Thanks!

PR #1656, opened 2016-11-07
Linux Key Code Modifiers

ryanbartley:
- Added a modifiers function to extract intuitive character from glfw using native key code modifiers.
- Haven't yet addressed unicode keys. But this step is needed in the immediate.


ryanbartley:
This has been updated and works fully for a US Keyboard on Linux.

PR #1658, opened 2016-11-07
Freetype Project Addition

ryanbartley:
- Added freetype to osx, ios, ios-sim projects
- Adding freetype to windows and winrt projects


ryanbartley:
This should be good, tested on all platforms. Looks like WinRT and CMake already has freetype. This pr makes all platforms and projects symmetric.

PR #1664, opened 2016-11-16
Made GlslProg::findUniform( const std::string&, int* ) behavior closer to GlslProg::findUniform( int, int* )

simongeilfus:
Updated findUniform( const std::string &name, int *resultLocation ) to have the same behavior as const findUniform( int location, int *resultLocation ) const. Currently one of the two methods breaks when trying to pass nullptr, while the other just skip it. This PR addresses this issue.

If I understand [the method's comment](https://github.com/cinder/Cinder/blob/master/include/cinder/gl/GlslProg.h#L441) correctly, it sounds like this should be the intended/expected behavior.:
```c++
//! Returns a const pointer to the Uniform that matches \a name. Returns nullptr if the uniform doesn't exist. The uniform location (accounting for indices, like "example[2]") is stored in \a resultLocation if it's non-null.
const Uniform* findUniform( const std::string &name, int *resultLocation ) const;
```


andrewfb:
Thanks for this fix

Issue #1662, opened 2016-11-11
CameraPersp::mLensShift is uninitialized


cdwfs:
None of the CameraPersp constructors initialize the mLensShift member. This can lead to mysterious NaNs in the camera projection matrix.  I'd provide a pull request with a fix, but I'm not sure what safe defaults would be; zero?

paulhoux:
Hi, 

yeah, zeroes would be right, as it would default to no lens shift.

cdwfs:
Ah, you're right. Our local GLM build has implicit vector initialization disabled. Our problem, not yours; sorry! Please disregard.

paulhoux:
Hey @cdwfs , as a matter of fact I am going to reopen this issue because I think this is an interesting debate. I can see Cinder using uninitialized GLM classes in the future as an optimization and we would bump into this issue again. 

Can I ask you to create a PR for this nevertheless? Feel free to add more class members to the initializer list of their constructor if you run into them.

richardeakin:
I responded to that pull request, but is it really an optimization to leave a couple floating point values uninitialized? Please profile beforehand.

As an argument against, the amount of bugs that would be introduced from GLM_FORCE_NO_CTOR_INIT would be huge. It would also mean all matrix types are uninitialized and no longer default to the identity matrix.

paulhoux:
@richardeakin  : see my response to the PR.

cdwfs:
Full disclosure: I'm actually not even using all of Cinder; I just pulled the Camera class out for use in a quick prototype project. I lost an hour or so chasing the bug that was ultimately due to the uninitialized lens shift, and figured it might be worth upstreaming (or at least mentioning). I don't personally feel that the impact on performance or readability is significant either way, but it's not my project. Nor am I particularly interested in participating in a style review for a trivial three-line change. I'm going to withdraw the PR; y'all'nz can sort out amongst yourselves whether you want to make the change or not.

paulhoux:
Fair enough, @cdwfs , it's probably not worth all the fuzz.

PR #1665, opened 2016-11-16
Explicitly initialize CameraPersp::mLensShift to (0,0)

cdwfs:
See issue #1662

### DIFF
--- a/src/cinder/Camera.cpp
+++ b/src/cinder/Camera.cpp
@@ -254,14 +254,14 @@ Ray Camera::calcRay( float uPos, float vPos, float imagePlaneApectRatio ) const
 // CameraPersp
 // Creates a default camera resembling Maya Persp
 CameraPersp::CameraPersp()
-	: Camera()
+	: Camera(), mLensShift(0,0)
 {
 	lookAt( vec3( 28, 21, 28 ), vec3(), vec3( 0, 1, 0 ) );
 	setPerspective( 35, 1.3333f, 0.1f, 1000 );
 }
 
 CameraPersp::CameraPersp( int pixelWidth, int pixelHeight, float fovDegrees )
-	: Camera()
+	: Camera(), mLensShift(0,0)
 {
 	float eyeX 		= pixelWidth / 2.0f;
 	float eyeY 		= pixelHeight / 2.0f;
@@ -277,7 +277,7 @@ CameraPersp::CameraPersp( int pixelWidth, int pixelHeight, float fovDegrees )
 }
 
 CameraPersp::CameraPersp( int pixelWidth, int pixelHeight, float fovDegrees, float nearPlane, float farPlane )
-	: Camera()
+	: Camera(), mLensShift(0,0)
 {
 	float halfFov, theTan, aspect;
 


paulhoux:
Thanks! Since this is (or might be) your first Cinder pull request: could you stick to the [style guidelines](https://github.com/cinder/Cinder/blob/master/CONTRIBUTING.md) and add spaces between brackets and values? Also, the ```vec2``` constructor also takes a single value, which might be more readable:
```
Camera() : mLensShift( 0 )
```

richardeakin:
I'm of the opinion that we shouldn't be initializing `ci::vec2/3/4`s to zero, this is unneeded noise. There are also many places throughout the codebase where they aren't (keep in mind that the matrix types are a composition of vector types). If you've enabled `GLM_FORCE_NO_CTOR_INIT` then you are bound to run into subtle uninitialized var bugs and there is no getting around it. Honestly I don't know why you would ever want to enable that anyway.

paulhoux:
This is not about optimization, Rich. This is to prevent ```Camera``` from initializing to garbage if compiling with GLM's ```GLM_FORCE_NO_CTOR_INIT```, which some users do and may be used by Cinder in the future, too. I believe this is a very good use case where implicit initialization is required.

richardeakin:
Apologies if I was a bit blunt - the contributions are of course always welcome and I think discussion is important to a healthy open source community. This one just happens to touch on a sensitive spot, which also happens to be one that we have debated at least a few times.  My main motivation was to bring to light the problems with globally enabling `GLM_FORCE_NO_CTOR_INIT` within the libcinder codebase, especially considering that coding patterns of many of our users (ex. they don't initialize variables to zero because clang does this for you).

I see the point that @cdwfs was making: a fix so that he could easily add a class within libcinder to a pre-existing project that has configured glm differently. I do wonder if something like linking to cinder as a dynamic library would solve this issue - that way the cinder version of glm could differ in compiler flags / features with a different codebase, and you could link to that to use classes like `CameraPersp`. I'm not entirely sure if there would be any binary compatibility issues, at first thought I would think it would be fine as the memory layout would be the same.

In the meantime I believe you can call `CameraPersp::setLensShift( 0, 0 )` directly after initializing it to use with your other project. But again I'd warn that who knows what other similar bugs you'd find in other corners of the library.

MikeGitb:
Just as a technical comment: 
I prefer not to do such default initializations in the initialization list of the constructor. I instead use [Default member initializers](http://en.cppreference.com/w/cpp/language/data_members). This reduces code clutter (especially with multiple constructors) and puts the default values directly at the point of the variable definition, which reduces the chance that you forget to initilize something.
 
Regarding binary compatibility when compiling cinder as a dynamic library with different flags: I'm no expert in this, but it should work, if the affected functions (in this case the constructors) are defined in the cpp file. But in general, it is a very brittle thing that might lead to strange errors.

cdwfs:
Just as a reply to @MikeGitb's technical comment: default member initializers are great, but they're a C++14 feature. Not all compilers in common use support C++14 yet, and not all companies allow C++14 features in their codebases yet. For these reasons, I myself tend to avoid them for now.

MikeGitb:
@cdwfs: Actually, they are a c++11 feature.

MikeGitb:
But admittedly, I don't know, if they are fully supported by VS2013.

richardeakin:
Default member initializers are supported since VS2013 / vc120 and recent (last couple years) versions of both gcc and clang. I've been moving to them myself and I agree, it is cleaner IMO then initializer lists, as well as shared across all constructors.

cdwfs:
You're right, my bad; I misread the docs. I'll be quiet now.

PR #1668, opened 2016-11-18
Updated asio to 1.10.8

ryanbartley:


andrewfb:
Spoke to Ryan about this; we're working on the dev branch of ASIO, which is 1.11.0. Going to remain on that and close this.

PR #1669, opened 2016-11-19
[warnings] Comment unused parameters in cinder/gl

MikeGitb:
See #1629

### DIFF
--- a/src/cinder/gl/Fbo.cpp
+++ b/src/cinder/gl/Fbo.cpp
@@ -316,7 +316,7 @@ void Fbo::initMultisamplingSettings( bool *useMsaa, bool *useCsaa, Format *forma
 }
 
 // Iterate the Format's requested attachments and create any we don't already have attachments for
-void Fbo::prepareAttachments( const Fbo::Format &format, bool multisampling )
+void Fbo::prepareAttachments( const Fbo::Format &format, bool /*multisampling*/ )
 {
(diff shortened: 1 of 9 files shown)


src/cinder/gl/Fbo.cpp
```diff
@@ -305,7 +305,7 @@ void Fbo::initMultisamplingSettings( bool *useMsaa, bool *useCsaa, Format *forma
 }
 
 // Iterate the Format's requested attachments and create any we don't already have attachments for
-void Fbo::prepareAttachments( const Fbo::Format &format, bool multisampling )
+void Fbo::prepareAttachments( const Fbo::Format &format, bool /*multisampling*/ )
```
richardeakin: If `multisampling` isn't used, shouldn't it be removed as an argument altogether? This is a non-virtual, protected method.

MikeGitb: @richardeakin: Probably. As it is not a private function I was a bit reluctant to change the interface. and to be honest I didn't really dig into the code or the documentation to find out, why the parameter was there in the first place. If I had to guess, I'd say it is a refactoring artifact (the only call site I could find was in the `init` function).

richardeakin: So it seems like there is some future work in progress to better support attachments with multisampling and some other configurations. In light of that, I'm going to accept this and just write a TODO there, which will likely get blown away by a work in progress overhaul.

PR #1670, opened 2016-11-19
[warnings] Comment unused parameters in cinder/audio

MikeGitb:
See #1629

### DIFF
--- a/include/cinder/audio/Buffer.h
+++ b/include/cinder/audio/Buffer.h
@@ -265,7 +265,7 @@ struct FreeDeleter {
 
 //! Returns an array of \a size elements of type \a T, aligned by \a alignment.
 template<typename T>
-std::unique_ptr<T, FreeDeleter<T> > makeAlignedArray( size_t size, size_t alignment = 16 )
+std::unique_ptr<T, FreeDeleter<T> > makeAlignedArray( size_t size, size_t /* alignment */ = 16 )
 {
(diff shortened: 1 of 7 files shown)


richardeakin:
Thanks for looking into these. I'm curious, what IDE and what warning flags are you using that emits warnings for these? When I build cinder with the vc2013 Visual Studio project, I don't see warnings for unused arguments.

I'm a bit hesitant when it comes to commenting out arguments in header files, as this is the main source of documentation and it adds a slight bit of clutter. But we're still discussing, of course warning spew also causes clutter so I see both sides.

MikeGitb:
@richardeakin: I compiled it with VS2015 U3 with /W4 (Which is my default setting but not the default for newly created projects, so most users will probably not see them). I explained this a bit in the linked issue, but the main Idea is that every now and then those warnings point to actual errors (like in #1643, #1639 or  #1634) or at least design flaws, but developers will (if at all) only pay attention to them, if they are not burried in false positives.

MikeGitb:
May I add, that the documentation concern is imho only partially valid: If the parameter hasn't any effect anyways then the current "documentation" actually gives a wrong impression about what the function does. So it is not only about warnings, but also about correcting the documentation.

MikeGitb:
Rebased onto current master

richardeakin:
I've incorporated this and made a few tweaks in #1723.

PR #1671, opened 2016-11-19
Fix multiple warnings about unused parameters

MikeGitb:
See #1629 for details.

### DIFF
--- a/include/cinder/svg/Svg.h
+++ b/include/cinder/svg/Svg.h
@@ -65,47 +65,45 @@ typedef std::function<bool(const Node&, svg::Style *)> RenderVisitor;
 //! Base class from which Renderers are derived.
 class Renderer {
   public:
-	Renderer() {}
-	
-	virtual ~Renderer() {}
+	virtual ~Renderer() = default;
 
 	void	setVisitor( const std::function<bool(const Node&, svg::Style *)> &visitor );
-	
-	virtual	void	pushGroup( const Group &group, float opacity ) {}	
-	virtual void	popGroup() {}	
-	virtual void	drawPath( const svg::Path &path ) {}
-	virtual void	drawPolyline( const svg::Polyline &polyline ) {}
-	virtual void	drawPolygon( const svg::Polygon &polygon ) {}
-	virtual void	drawLine( const svg::Line &line ) {}
-	virtual void	drawRect( const svg::Rect &rect ) {}
-	virtual void	drawCircle( const svg::Circle &circle ) {}
-	virtual void	drawEllipse( const svg::Ellipse &ellipse ) {}
-	virtual void	drawImage( const svg::Image &image ) {}
-	virtual void	drawTextSpan( const svg::TextSpan &span ) {}
 
-	virtual void	pushMatrix( const mat3 &m ) {}
+	virtual	void	pushGroup( const Group & /*group*/, float /*opacity*/ ) {}
+	virtual void	popGroup() {}
+	virtual void	drawPath( const svg::Path & /*path*/ ) {}
+	virtual void	drawPolyline( const svg::Polyline & /*polyline*/ ) {}
+	virtual void	drawPolygon( const svg::Polygon & /*polygon*/ ) {}
+	virtual void	drawLine( const svg::Line & /*line*/ ) {}
+	virtual void	drawRect( const svg::Rect & /*rect*/ ) {}
+	virtual void	drawCircle( const svg::Circle & /*circle*/ ) {}
+	virtual void	drawEllipse( const svg::Ellipse & /*ellipse*/ ) {}
+	virtual void	drawImage( const svg::Image & /*image*/ ) {}
+	virtual void	drawTextSpan( const svg::TextSpan & /*span*/ ) {}
+
+	virtual void	pushMatrix( const mat3 & /*m*/ ) {}
 	virtual void	popMatrix() {}
-	virtual void	pushStyle( const svg::Style &style ) {}
-	virtual void	popStyle( const svg::Style &style ) {}	
-	virtual void	pushFill( const class Paint &paint ) {}
+	virtual void	pushStyle( const svg::Style & /*style*/ ) {}
+	virtual void	popStyle( const svg::Style & /*style*/ ) {}
+	virtual void	pushFill( const class Paint & /*paint*/ ) {}
 	virtual void	popFill() {}
-	virtual void	pushStroke( const class Paint &paint ) {}
+	virtual void	pushStroke( const class Paint & /*paint*/ ) {}
 	virtual void	popStroke() {}
-	virtual void	pushFillOpacity( float opacity ) {}
+	virtual void	pushFillOpacity( float  /*opacity*/ ) {}
 	virtual void	popFillOpacity() {}
-	virtual void	pushStrokeOpacity( float opacity ) {}
+	virtual void	pushStrokeOpacity( float  /*opacity*/ ) {}
 	virtual void	popStrokeOpacity() {}
-	virtual void	pushStrokeWidth( float width ) {}
+	virtual void	pushStrokeWidth( float  /*width*/ ) {}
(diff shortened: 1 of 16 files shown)


include/cinder/svg/Svg.h
```diff
@@ -65,60 +65,58 @@ typedef std::function<bool(const Node&, svg::Style *)> RenderVisitor;
...
-	virtual void	pushLineCap( LineCap lineCap ) {}
-	virtual void	popLineCap() {}	
-	virtual void	pushLineJoin( LineJoin lineJoin ) {}
-	virtual void	popLineJoin() {}
-	virtual void	pushTextPen( const vec2 &penPos ) {}
-	virtual void	popTextPen() {}
-	virtual void	pushTextRotation( float rotation ) {}
-	virtual void	popTextRotation() {}
+
+	virtual	void	pushGroup( const Group & /*group*/, float /*opacity*/ ) {};
```
richardeakin: The trailing semicolon here - does this prevent some warning?

src/AntTweakBar/TwMgr.cpp
```diff
@@ -6352,6 +6352,8 @@ void CTwMgr::SetCursor(CTwMgr::CCursor _Cursor)
         if( ok && (ci.flags & CURSOR_SHOWING) )
             ::SetCursor(_Cursor);
     }
+#else
+	(void)_Cursor;
```
richardeakin: Not so important in this specific case as it is part of AntTweakBar's source code (which will eventually be deprecated and removed), but I think in cases like this we want to use some sort of `#pragma` or macro to declare that the variable is unused. Otherwise the code looks a bit mysterious.

richardeakin:
Most of these look great to me. However when it comes to third-party code, what if we use a `#pragma warning( disable: ... )` instead, so that we can leave the actual source as untouched as possible? I did this quite extensively in our embedded oggvorbis source code to silence warnings, while keeping it easy(er) to update in the future.

MikeGitb:
Sounds reasonable - I didn't think about that while going through the warnings. 
Is all code outside of src/cinder or include/cinder external?

richardeakin:
> Is all code outside of src/cinder or include/cinder external?

This is correct, those other folders are  thirdy-party dependencies that we compile into the libcinder library (or header only libraries we depend on).

MikeGitb:
Ok, I'll then remove the changes from those files - although it might take a few days before I find time to do that

MikeGitb:
@richardeakin: Is there a way to suppress those warnings for clang++/g++ too?

richardeakin:
Yes - in Xcode you need to to add the compiler flag to the individual .cpp file under Project Settings -> Build Phases -> Compile Sources, like this:

screenshot 2016-11-25 15 12 12

There is also a way to add compiler flags to specific files in cmake too (not yet in master branch), though I can't remember exactly how at the moment.

MikeGitb:
@richardeakin: Apparently clang and gcc have their own version: `#pragma clang diagnostic ignored "-Wunused"`

So I can add the following:

	#if defined(__clang__ && __GCC__)
		#pragma GCC diagnostic push
		#pragma GCC diagnostic ignored "-Wunused-parameter"
	#elsif defined(_MSC_VER)	
		#pragma warning(push)
		#pragma warning(disable : 4100) 
	#endif

// code...

	#if defined(__clang__ && __GCC__)
		#pragma GCC diagnostic pop
	#elsif defined(_MSC_VER)	
		#pragma warning(pop)
	#endif

However, I'm beginnig to doubt if this is really better than correcting the 2-3 lines containing the unused function parameters

src/AntTweakBar/TwMgr.cpp
MikeGitb: The advantage of this technique compared to pragmas is that it works accross all compilers and I've actually seen this quite often in other c code bases.

richardeakin:
yea the `#pragma clang diagnostic` seems cleaner to me - it modifies the source but at least in an obvious way, much more obvious than hidden in the depths of xcode settings menus. I think either is fine.

Thanks again for cleaning all this up!

MikeGitb:
Ok, I'm now using the suppression pragma in all files that come from external libraries

richardeakin:
To play the devil's advocate here (and sorry for not mentioning this earlier), one advantage to disabling the warnings from the project files for third-party code is that we don't have to modify the source at all, then, so updating it becomes easier. Still, we're unlikely to ever be updating some of these deps, like AntTweakBar for example, which is already massively modified to work with `ci::gl`.

I'm personally not a fan of seeing the commented out arguments in Svg.h for virtual functions but it isn't the end of the world to me, we should remove the warnings some way or another and perhaps in this case the `/* */` is the best choice. Usually, I'd say we should move the empty virtual implementation out-of-line and comment out the unused var there.

All that said, I think this one is good to merge as it is a nice improvement. The audio (#1670) and gl ( (#1669)  ones need more thought, I think, albeit mostly philosophical.

MikeGitb:
One question up front: Do you know, why `Renderer::popStyle` takes a `const svg::Style &` parameter while all other pop functions don't have a parameter? Is this a Copy-Paste error? Should we remove it or leave a comment about it's purpose?

> one advantage to disabling the warnings from the project files for third-party code is that we don't have to modify the source at all, then, so updating it becomes easier. 

Yes, but this only works for .c/.cpp files (which are admittedly most of them) not for headers. Also, you'd have to replicate that setting for each toolchain. I can't do this, because I don't have access to XCode or VS2013 and I won't find the time to dig through the new cmake build anytime soon. But I can remove those commits from my PR.

> I'm personally not a fan of seeing the commented out arguments in Svg.h for virtual functions but it isn't the end of the world to me, we should remove the warnings some way or another and perhaps in this case the /* */ is the best choice. Usually, I'd say we should move the empty virtual implementation out-of-line and comment out the unused var there.

Another solution would be to void-cast the parameters. If you want I can pull the changes to Renderer from the PR for now or implement a different workaround (out-of-band or void-cast).
 
On a more general note: Have you considered making `svg::Renderer` a pure virtual interface (except for the visitor functionality) that doesn't provide empty default implementations for the methods at all? Advantage and disadvantage would be that a derived class would have to explicitly define all overrides, even if it (deliberately or not) doesn't implement any actual functionality for them. It would eliminate the warnings as a side effect.

richardeakin:
> why Renderer::popStyle takes a const svg::Style & parameter while all other pop functions don't have a parameter?

Good catch; from a quick search nothing is using that argument. Actually, both `SvgRendererGl` and `SvgRendererCairo` (in the blocks folder, [here](https://github.com/cinder/Cinder/blob/d0fcce0972f2030ae4a8945959fc223d2cdb79ec/blocks/Cairo/src/Cairo.cpp#L2103)) have empty implementations, but I couldn't tell you why, It appears that code was written before we began using `override`, so pre-C++11.

> Yes, but this only works for .c/.cpp files (which are admittedly most of them) not for headers. Also, you'd have to replicate that setting for each toolchain. I can't do this, because I don't have access to XCode or VS2013 and I won't find the time to dig through the new cmake build anytime soon. But I can remove those commits from my PR.

Very valid points for adding the warning silencing directly to the source.

> Another solution would be to void-cast the parameters. If you want I can pull the changes to Renderer from the PR for now or implement a different workaround (out-of-band or void-cast).

Let's see if others have any opinion here among our options, as I think the verdict will affect how we go forward with silencing the rest of the warnings. Will make a note here to discuss as well.

drewish:
I've been spending a fair bit of time with CGAL and they comment out unused params. I found it a bit annoying at first but now don't mind it. I think it's probably worth the irritation to get the extra sanity check.

richardeakin:
Added a comment around this to the mother-issue: https://github.com/cinder/Cinder/issues/1629#issuecomment-273455893

I believe @andrewfb wanted to look at the svg `popStyle()` aspect of this PR around whether the argument should be removed entirely, otherwise I think it is good to go.

PR #1675, opened 2016-12-05
A fix for an issue where indices were incompletely copied over from g…

mhintz:
…eom::Source when being drawn with gl::draw

This issue was causing certain geom::Source classes, which may specify less than 4 bytes per index in their copyIndices calls, to be incorrectly rendered when passed to gl::draw(). Namely, the Cube, Icosahedron, and Extrude classes.

gl::draw(geom::Source &) constructs a DefaultVboTarget into which the geom::Source is loaded. However, without index type conversion like in VboMeshGeomTarget, the indices are incompletely copied, ending up garbled.

This seemed to me like the most straightforward solution to the problem. It wouldn’t be the first copyIndices impl to totally ignore requiredBytesPerIndex. In fact, only VboMeshGeomTarget::copyIndices actually uses this parameter.

### DIFF
--- a/src/cinder/GeomIo.cpp
+++ b/src/cinder/GeomIo.cpp
@@ -536,7 +536,7 @@ void Target::generateIndicesForceLines( Primitive primitive, size_t numInputIndi
 
 void Target::copyIndexData( const uint32_t *source, size_t numIndices, uint32_t *target )
 {
-	memcpy( target, source, numIndices * sizeof(float) );
+	memcpy( target, source, numIndices * sizeof(uint32_t) );
 }
 
 void Target::copyIndexData( const uint32_t *source, size_t numIndices, uint16_t *target )
--- a/src/cinder/gl/draw.cpp
+++ b/src/cinder/gl/draw.cpp
@@ -544,7 +544,7 @@ class DefaultVboTarget : public geom::Target {
 			return;
 
 		mIndexType = GL_UNSIGNED_INT;
-		mElementVbo->bufferSubData( 0, numIndices * requiredBytesPerIndex, sourceData );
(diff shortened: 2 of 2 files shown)


andrewfb:
Thanks for taking a look at this Mark. Do you have a concrete use case where we can reproduce the broken behavior? It doesn't have to be a full sample - just a C++ "phrase" that invokes it is plenty.

mhintz:
Sure thing. I got it when doing `gl::draw(geom::Cube())` or `gl::draw(geom::Icosahedron())`. I have a minimal app setup that replicates it.

Worth noting that the Geometry sample doesn't pick up this problem, because there the geom::Source objects are first passed through a TriMesh, which copies the indices correctly.

mhintz:
I was just looking through the implementation of `VboMeshGeomTarget::copyIndices`, which does the index bytes adjustment which the function I've edited here does not do, and I think there's a slight error in the `geom::Target::copyIndices(const uint32_t *, size_t, uint32_t *)` implementation:

```
void Target::copyIndexData( const uint32_t *source, size_t numIndices, uint32_t *target )
{
	memcpy( target, source, numIndices * sizeof(float) );
}
```

shouldn't that be `sizeof(uint32_t)`?

richardeakin:
> shouldn't that be sizeof(uint32_t)?

Nice catch!

mhintz:
I added the sizeof fix here; didn't seem big enough to warrant a separate PR...

mhintz:
Bumping this, since it's been a while, I think this should be fixed, and it's blocking me from updating cinder with the latest changes from master.

andrewfb:
Apologies for the delay, and thanks for this find & fix.

PR #1676, opened 2016-12-05
Added signals::ConnectionList

richardeakin:
Utility to easily maintain a bunch of connections, disconnecting them when container is destroyed. Tying connections to an object's scope is vital in making sure runtime reloads are safe, and this makes it much easier to do so as compared to keeping a bunch of `ScopedConnections` (the dreaded mConnTouchBegan, mConnTouchMove, mConnBlah... anti-pattern).

Convenient syntax looks like:

```cpp
ci::signals::ConnectionList mConnections;

...

mConnections += getSignalXXX().connect( [this] { ... }; );
mConnections += getSignalYYY().connect( [this] { ... }; );
```

Right now it only accepts an R-value reference to a `Connection`, I'm not sure if we need anything more than this.

- [x] Add unit test.

### DIFF
--- a/src/cinder/Signals.cpp
+++ b/src/cinder/Signals.cpp
@@ -83,11 +83,6 @@ ScopedConnection::~ScopedConnection()
 	disconnect();
 }
 
-ScopedConnection::ScopedConnection( const Connection &other )
-	: Connection( other )
-{
-}
-
 ScopedConnection::ScopedConnection( ScopedConnection &&other )
 	: Connection( std::move( other ) )
 {
@@ -96,13 +91,6 @@ ScopedConnection::ScopedConnection( ScopedConnection &&other )
 ScopedConnection::ScopedConnection( Connection &&other )
 	: Connection( std::move( other ) )
 {
-}
-
-ScopedConnection& ScopedConnection::operator=( const Connection &rhs )
-{
-	disconnect();
-	Connection::operator=( rhs );
-	return *this;
 }
 
 ScopedConnection& ScopedConnection::operator=( ScopedConnection &&rhs )
--- a/src/cinder/Signals.cpp
+++ b/src/cinder/Signals.cpp
@@ -125,6 +125,8 @@ void ConnectionList::clear()
 {
 	for( auto &conn : mConnections )
 		conn.disconnect();
+
+	mConnections.clear();
 }
 
 namespace detail {
--- a/src/cinder/Signals.cpp
+++ b/src/cinder/Signals.cpp
@@ -80,7 +80,8 @@ ScopedConnection::ScopedConnection()
 
 ScopedConnection::~ScopedConnection()
 {
-	disconnect();
+	if( mShouldDisconnect )
+		disconnect();
 }
 
 ScopedConnection::ScopedConnection( ScopedConnection &&other )
@@ -95,6 +96,9 @@ ScopedConnection::ScopedConnection( Connection &&other )
 
 ScopedConnection& ScopedConnection::operator=( ScopedConnection &&rhs )
 {
+	disconnect(); // first disconnect from existing
+	rhs.mShouldDisconnect = false; // mark that rhs shouldn't disconnect as it is about to be moved to this
+
 	Connection::operator=( std::move( rhs ) );
 	return *this;
 }
--- a/src/cinder/Signals.cpp
+++ b/src/cinder/Signals.cpp
@@ -118,7 +118,7 @@ ConnectionList::~ConnectionList()
 
 void ConnectionList::add( Connection &&target )
 {
-	mConnections.emplace_back( target );
+	mConnections.emplace_back( std::move( target ) );
 }
 
 void ConnectionList::clear()
--- a/src/cinder/Signals.cpp
+++ b/src/cinder/Signals.cpp
@@ -23,6 +23,8 @@
 
 #include "cinder/Signals.h"
 
+using namespace std;
+
 namespace cinder { namespace signals {
 
 Connection::Connection()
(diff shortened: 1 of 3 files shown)


drewish:
Would it make sense to have CameraUI use this?

richardeakin:
I think so, although `CameraUi` currently disconnects its connections at the proper time (destruction or whenever you reconnect it).

I mainly use it in objects that have signal connections and are often destroyed / reloaded.

src/cinder/Signals.cpp
```diff
@@ -111,6 +111,22 @@ ScopedConnection& ScopedConnection::operator=( ScopedConnection &&rhs )
...
+}
+
+void ConnectionList::add( Connection &&target )
+{
+	mConnections.emplace_back( target );
+}
+
+void ConnectionList::clear()
+{
+	for( auto &conn : mConnections )
```
MikeGitb: Not particularly important, but shouldn't clear also remove the vectors from the list? And couldn't this further be simplified by `ScopedConnection`

src/cinder/Signals.cpp
```diff
@@ -111,6 +111,22 @@ ScopedConnection& ScopedConnection::operator=( ScopedConnection &&rhs )
...
 }
 
+ConnectionList::~ConnectionList()
+{
+	clear();
+}
+
+void ConnectionList::add( Connection &&target )
+{
+	mConnections.emplace_back( target );
```
MikeGitb: Aren't you missing a move here?

src/cinder/Signals.cpp
richardeakin: Good point, yes should have cleared the vector too..

Keeping a vector of `ScopedConnection`s was problematic last time I tried it, but I can try it again as it seems like there is no reason it shouldn't work if we're `std::move()`ing things into it.

MikeGitb: Hm, I wonder if that is an indication of an error in the move/copy constructor/assignement operator in `Connection` or `ScopedConnection`.

MikeGitb: Actually, ScopedConnection look pretty broken to me: It derives from noncopyable, but then defines a copy constructo/assignment operator. The resulting semantics is that whichever copy is destructed first, will disconnect the link. Also: If the move constructor / assignment operator is not marked as noexcept, std::vector will copy and not move it's elements when reallocating.

Together, this would probably explain the problems you had in the past.

richardeakin: @MikeGitb thanks for your great review. I've added all of you suggestions but switching to `ScopedConnection` for `ConnectionList`'s container type, as `noexcept` isn't available to us until we switch to Visual Studio 2015 (opened [this issue](https://github.com/cinder/Cinder/issues/1681)). It sounds like using `vector<ScopedConnection>` won't be safe until then.

Everything look good to you?

richardeakin: Hm, removing the copy constructor and assignment operators that took a `Connection` seems to have exposed an issue with `ScopedConnection`'s move assignment operator. Basically the old (moved) ScopedConnection ended up still disconnecting the signal. Tentative fix [here](https://github.com/cinder/Cinder/pull/1676/commits/c5cac01d293b9a0a210ba263e13327acac97d0aa) (with a new unit test for it as well).

MikeGitb: I don't have a dev-environment at hand right now, but I think there is a cleaner solution: Imho move constructor and move assignment operator should properly reset the object variable you move from to a null state, such that a call to `disconnect` on a moved from object simply doesn't have any effect. 

Do those problems only affect VS2013? Because I'd expected the compiler to automatically create a move constructor for `Connection` that has exactly that effect, but VS2013 seems to be pretty limited in that regard.

src/cinder/Signals.cpp
```diff
@@ -95,6 +96,9 @@ ScopedConnection::ScopedConnection( Connection &&other )
 
 ScopedConnection& ScopedConnection::operator=( ScopedConnection &&rhs )
 {
+	disconnect(); // first disconnect from existing
```
MikeGitb: Shouldn't that disconnect also be conditioned on `mShouldDisconnect`?

src/cinder/Signals.cpp
richardeakin: @MikeGitb great advice, that makes sense to me so I'm in the middle of implementing it this way. One question comes up though, does this seem appropriate?

```cpp
Connection& Connection::operator=( Connection &&rhs )
{	
	mDisconnector = move( rhs.mDisconnector );
	rhs.mDisconnector.reset(); // <-- why is this needed?

	mLink = move( rhs.mLink );
	rhs.mLink = nullptr;

	mPriority = move( rhs.mPriority );
	rhs.mPriority = 0;

	return *this;
}
```
This seems wrong to me according to the weak_ptr [docs](http://en.cppreference.com/w/cpp/memory/weak_ptr/weak_ptr), which state that once I've moved a weak_ptr then it will be empty. However If I don't explicitly `.reset()` mDisconnector there, then it still has a valid weak reference, and its `use_count()` reports 1 (in my unit test).

Here's an [SO post](http://stackoverflow.com/a/26581337/506584) explaining the lack of implicit generation of move constructors and move assignment operators in VS2013. I think we have to make this work for that compiler, so I'm fine with adding in the extra constructors for now, just want to make sure to get them right.

richardeakin: So talking with @andrewfb here, turns out that `std::weak_ptr` doesn't have a move constructor or assignment operator until C++14, and vc120 must not support this yet. So I think the interim solution is to leave that `.reset()` in there and mark it with a TODO. The rest of the operator has no effect as `rhs` is about to die, so I'll remove those.

MikeGitb: @richardeakin: Sorry, for not answering earlier. I read the post, but then it got swept under the ruggs during the holidays. 

In any case, the implementation sounds good to me, but you probably also have to remove the copy constructor and copy assignment operator or specify the move constructor and move assignment operator as noexcept (at least on platforms other than VS2013). Otherwise, a c++11 conforming implementation of `std::vector` will use the copy assignment/constructor instead of the move operators when it moves things around (e.g. when using `push_back`). I believe VS2013 and VS2015 will still use the move operations , but e.g. gcc on linux will use the copy operators.

richardeakin: @MikeGitb thanks for the continual suggestions and helping to make this implementation better. 

About `std::vector`, I think we're in the clear for now as I left the vector type in `ConnectionList` as `Connection`, thinking we'd move to it using `ScopedConnection` once our bottom end compilers catch up. Or are you foreseeing some other problem? If the issue is concerning users keeping their own vector of `ScopedConnection`s, I think we can address that in a separate pull request if you see a good way to do it. It won't work across all platforms though, so I'm happy with waiting a bit on that considering we'll be switching to vc140 first thing in the next release cycle.

I pushed the recently discussed changes in c4c736f, so barring the current question this should be good.

richardeakin:
Going to merge this, if we've got more work to do around `ScopedConnection` let's bring that over to the #1681 discussion.

Issue #1611, opened 2016-10-11
Add continuous integration 


MikeGitb:
As the title says: Please consider adding support for CI. For open source projects this is usually free and a basic configuration is set up in just a few minutes (at least when you already know how to build from the command line).

Especially for code that is supposed to compile and work on multiple different platforms and compiler versions, I think it is very helpful if the developer has a simple way to check that a change that works well on the local machine doesn't break the program on a different platform.

Of course there are limits on how far you can automate testing of a GUI framework on a console based infrastructure, but you get at least some feedback.

richardeakin:
Thanks for opening this. We've been actively looking into CI using both travis for POSIX and appveyor for Windows.  The POSIX part depends on work done in the android_linux branch, which will soon end up in master / 0.9.1 release.

It would however be nice if we could find one solution that worked across all platforms, but unfortunately I don't think there is one that will also provide us machines (even if virtual) to run the builds on.  Still, open to suggestions.

MikeGitb:
I also don't believe there is a free CI service that supports both POSIX and Windows and I haven't checked about android, but setting up Travis and appveyor for the master branch's Mac and non-winrt windows version took me only a couple of minutes each so I think the overhead is acceptable (I didn't try unit tests thought).

richardeakin:
- [X] Travis-CI for Linux and Mac
- [ ] AppVeyor for Windows.

richardeakin:
AppVeyor added with https://github.com/cinder/Cinder/pull/1791, currently in dev_0.9.2 branch and will be merged into master after 0.9.1 release is out.

PR #1679, opened 2016-12-09
Adds Travis CI support for OS X (various xcode versions) and Ubuntu Trusty

### LINKED ISSUE #1263: Fft unit test failing on windows
Appears to be with OOURA backend when sizeFft = 4, although this works fine on mac. Also, `MonitorSpectralNode` is working fine so it's probably not a huge problem but need to look into it.

lucasvickers:
Addresses part of #1611, bandaid fixes #1263 

Adds Travis CI support for OS X (various xcode versions) and Ubuntu Trusty (14)
Specific changes
- Added an ios specific build script to proj/xcode.
  - Wrapped this build in xcpretty since the verbosity of ios build breaks Travis (logs > 4MB)
- Modified the JsonTest to use the /tmp directory rather than the documents dir.  In a Travis env there is no documents dir.
- Disabled unimplemented system tests in SystemTest for linux
- Disabled FFT tests for non CINDER_MAC platforms
- Added scripts to tools/ for the following
  - tools/ci/cmake_test.sh - walks through building debug and release versions of cinder via cmake
  - linux/scripts/trusty-install-deps.sh - installs Trusty specific dependencies
  - tools/osx/scripts/cmake-brew-install.sh - installs cmake on OS X platform if not already installed

Results of this PR's build can be seen [here](https://travis-ci.org/lucasvickers/Cinder/builds/182710825).
To enable Travis-CI on Cinder, you'll want settings similar to these:

screen shot 2016-12-09 at 3 40 31 pm

### DIFF
--- a/test/unit/src/audio/FftUnit.cpp
+++ b/test/unit/src/audio/FftUnit.cpp
@@ -50,9 +50,11 @@ TEST_CASE( "audio/Fft" )
 
 SECTION( "round trip error" )
 {
+#if defined( CINDER_MAC )
 	CI_LOG_I( "... Fft round trip max acceptable error: " << ACCEPTABLE_FLOAT_ERROR );
 	for( size_t i = 0; i < 14; i ++ )
 		computeRoundTrip( 2 << i );
+#endif
 }
 
 } // "audio/Fft"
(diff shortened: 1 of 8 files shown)


test/unit/src/audio/FftUnit.cpp
```diff
@@ -50,9 +50,11 @@ TEST_CASE( "audio/Fft" )
 
 SECTION( "round trip error" )
 {
+#if defined( CINDER_MAC )
```
richardeakin: Could you move this define to the top, replacing [this line](https://github.com/lucasvickers/Cinder/blob/15b051deb7b044b6dd923f06bd87c5ab437269e5/test/unit/src/audio/FftUnit.cpp#L4)? I'm hoping to look at this soon but at least that will keep things tidy in the meantime.

richardeakin:
Awesome work on getting CI going!

PR #1686, opened 2016-12-18
Reallocate memory when shrinking an audio::DynamicBuffer to fit

notlion:
This fixes `audio::DynamicBufferT::shrinkToFit()`. Previously the buffer's internal vector was resized, preserving the capacity. This change ensures that the extra memory is freed.

The swap trick was suggested by @chaoticbob. I would have rather used vector::shrink_to_fit(), but that function might not actually free memory. It's apparently implementation dependent.


MikeGitb:
Looks good to me.

Though I wonder, why there is a separate mAllocatedSize Variable that is just an imperfect representation of `vector::capacity()`. Why not use `vector::capacity()` in the first place?

@chaoticbob & @notlion: Correct me if I'm wrong, but afaik the swap method doesn't provide any more guarantees than `shrink_to_fit` does. I don't think anything in the standard prevents an implementation where a newly created `std::vector` has a bigger capacity than it's size. 
Also, I believe that at least one reason why `shrink_to_fit` is nonbinding is that the actual heap allocation often only happens in chunks of certain sizes. So if you can only have a chunk of either 128 or 256 bytes anyway, it doesn't make sense to shrink a vector from 200 to 180 - you'd still block a chunk of size 256 (numbers are just randomly picked).

notlion:
I was also curious about the `mAllocatedSize` variable. It does seem to represent the same thing, however I suppose it would become out of sync if the vector implementation did in fact allocate a larger size that was specified. Perhaps @richardeakin could shed some light on the reasoning for that.

If the swap trick doesn't provide any more guarantees then it should probably be replaced with `shrink_to_fit` since that method is much clearer. I'd be happy to make both these changes if you guys feel it's appropriate.

MikeGitb:
I did a bit more searching on that and particular the post of the VisualC++ STL implementer on reddit (https://www.reddit.com/r/cpp/comments/1r1s58/did_the_committee_blow_it_with_shrink_to_fit/) convinced me that using `std::vector::shrink_to_fit` is indeed the right thing to do here.

As far as `mAllocatedSize` is concerned I get the impression that this might be a refactoring artifact. Maybe the buffer was a pointer to a dynamically allocated array in the past which required to keep track of it's length (or someone jsut didn't know vector too well). But I agree that you should wait for feedbrack from one of the maintainers before changing it.

notlion:
Mm. This makes sense to me:

> Either way, you're asking the Standard Library to determine the ideal capacity for N elements. Smart implementations will exhibit the same behavior for both.

If we used `vector::capacity` then `BufferDynamic::shrinkToFit` becomes a passthrough to `vector::shrink_to_fit`, which.. I suppose there's not reason for it to be more than that.

richardeakin:
Apologies for a delayed response (project deadline approaching) and thanks for the insightful conversation.

As some background information: when the `audio::Buffer` classes were originally designed, it was only a temporary plan to use `std::vector` for the internal data store.  However, it remained because they are easier to debug compared to manually allocated heap memory (most IDEs provide an easy way to see the element values).

It does seem like I missed the call to `shrink_to_fit()` in `BufferDynamicT::shrink_to_fit()`, and I do agree with some points in that reddit article that at least `shrink_to_fit()` is much more legible. Now, if we are to move forward with using std::vector's `capacity()` instead of the extra `mAllocatedSize` that is carted around, that means we're ditching any future plan to use a custom allocated memory store. I'd like to know what you'all think about this. For me, that we're at the mercy of this much debated spec seems like a disadvantage to me over using a raw memory store, but then I do like its ease of use when debugging.

One other thing that needs to be thought about: we could very much use something like an `audio::BufferView`, which I have been intending to add ever since the initial design but haven't found the time. This class would be similar in spirit to C++17's `std::string_view`, in that it provides you with a 'window' into an existing `audio::Buffer` that doesn't require any copying. It would also subclass `audio::Buffer`, as `BufferDynamic` does, so that it could be passed into any methods that can handle `audio::Buffer *` (or a `const &`).  At one point I was thinking that use `std::vector` as the internal data store might complicate this, especially when resizing, but I haven't completely thought this through. Another minor issue would be that `BufferView` would essentially inherit a `std::vector` that it doesn't use, but instead keeps a raw pointer to some memory, which would be similar as in the case if we ditched the std::vector and used `malloc()` / `free` throughout.

I just realized there are no unit tests for BufferDynamic, shame on me... :/ We should probably add that to [`BufferUnit.cpp`](https://github.com/cinder/Cinder/blob/master/test/unit/src/audio/BufferUnit.cpp).

MikeGitb:
can`t you just reintroduce `mAllocatedSize` if and when std::vector gets ditched?

In VS, debug ability could be addressed by providing a natvis file. I assume xcode has something similar?

Concerning the view: In most examples I know of, the view is either the base class or a completely independent datastructure that can constructed cheaply from the corresponding containers.

richardeakin:
I was thinking the same for VS debugging, just haven't gotten around to learning how to do it. But if we could view an `audio::Buffer` in the debugger similar to `std::vector`, that would be really slick and basically make me lean towards ditching the vector in favor of having more control over when and what is allocated.

I do think that `BufferView` should subclass `Buffer`, for the sake of keeping virtual methods that need to be implemented to minimum when subclassing things like `audio::Node` or `audio::Source`.  We can't do the same tricks that STL does as most of ci::audio is based on runtime polymorphism.

Well, sorry @notlion for taking your PR for a roller coaster ride. :) So far as I can tell, the right thing to do here for now is to add a call to `this->mData.shrink_to_fit()` after resizing, and I'll make a note to come back to this when I (or someone else) get to `BufferView`. I do think that would be a very welcome addition, as it will cut down on much of the per-processing-block copying that is currently necessary.

notlion:
Just got back from mini-vacation. Sorry for my lack of input on this. I think I agree that using `shrink_to_fit` is the best thing to do here. I made that change in the last commit. Swapping `mAllocatedSize` for `vector::capacity()` and other changes can happen in another PR if you guys think that's the way to go.

MikeGitb:
Looks good to me

richardeakin:
Thanks again for the reviewing and discussion, which will surely help in making better decisions during future `ci::audio` work. I've created #1701 for planning BufferView.

PR #1689, opened 2016-12-21
Fix GstPlayer compilation on older linux.

PetrosKataras:


richardeakin:
Looks like this fixes the trusty build, the new failure is due to log overflow which we'll soon address.

PR #1690, opened 2016-12-21
android_linux merge

cinder:
Merges the `android_linux` branch with `master`, which includes support for Linux and CMake, as well as experimental support for Android.


drewish:
For anyone else who's wondering why the can't open the Xcode project after this got merged, it moved into the `proj` directory ;)

PR #1691, opened 2016-12-23
Allow 3d triangulation

drewish:
I needed to do some basic 3d triangulation and noticed that libtess2 supports it but the Triangulator wrapper did not. It wasn't too hard to hook it up but the mesh output is wonky. I've added some really basic unit tests to demonstrate what it's doing.


include/cinder/TriMesh.h
```diff
@@ -200,7 +200,7 @@ class TriMesh : public geom::Source {
 	size_t		getNumIndices() const override { return mIndices.size(); }
 	//! Returns the total number of triangles contained by the TriMesh.
 	size_t		getNumTriangles() const { return mIndices.size() / 3; }
-	//! Returns the total number of indices contained by the TriMesh.
+	//! Returns the total number of verticies contained by the TriMesh.
```
drewish: I was referring to these docs and this seemed like a bit of copy paste from a couple lines above.

include/cinder/Triangulate.h
```diff
@@ -54,11 +56,15 @@ class Triangulator {
...
+	void		addPolyLine( const PolyLine3f &polyLine );
+	//! Adds a PolyLine defined as a series of vec3's
+	void		addPolyLine( const vec3 *points, size_t numPoints );
 
 	//! Performs the tesselation, returning a TriMesh2d
-	TriMesh		calcMesh( Winding winding = WINDING_ODD );
+	TriMesh		calcMesh( Winding winding = WINDING_ODD, u_int8_t positionDims = 2 );
 	//! Performs the tesselation, returning a TriMesh2d
-	TriMeshRef	createMesh( Winding winding = WINDING_ODD );
+	TriMeshRef	createMesh( Winding winding = WINDING_ODD, u_int8_t positionDims = 2 );
```
drewish: I'm not a fan of this interface but couldn't think of anything better.

src/cinder/Triangulate.cpp
```diff
@@ -105,23 +111,43 @@ void Triangulator::addPolyLine( const vec2 *points, size_t numPoints )
...
-	TriMesh result( TriMesh::Format().positions( 2 ) );
+	TriMesh result( TriMesh::Format().positions( positionDims ) );
 	
-	tessTesselate( mTess.get(), (int)winding, TESS_POLYGONS, 3, 2, 0 );
-	result.appendPositions( (vec2*)tessGetVertices( mTess.get() ), tessGetVertexCount( mTess.get() ) );
+	tessTesselate( mTess.get(), (int)winding, TESS_POLYGONS, 3, positionDims, 0 );
+	if( positionDims == 2 ) {
+		result.appendPositions( (vec2*)tessGetVertices( mTess.get() ), tessGetVertexCount( mTess.get() ) );
+	} else {
+		result.appendPositions( (vec3*)tessGetVertices( mTess.get() ), tessGetVertexCount( mTess.get() ) );
```
drewish: I'd love suggestions on better ways to make this type conversion conditional.

paulhoux:
Will this PR allow people to add both 2D and 3D vertices to the Triangulate instance? Because I think that would be problematic.

drewish:
It seems like libtess2 uses 3 dimensions internally and leaves the z at 0 if you don't provide a value but I think it could be clearer to have specialized classes for each. I'll take a swing at that interface and see how it looks.

drewish:
I took a swing at putting the 3d version in it's own class. I'm not sure if there's a good way to share code between the two classes. Since this is really just a proof of concept I set it up as a `Triangulator3d` class that used `Triangulator::Winding` and `Exception`.  Ended up having to duplicate a fair bit of code but open to see what people think.

vinjn:
How about converting into a templated class.

```C++
template<typename T>
class Triangulator {
};

typedef Triangulator<vec2>		Triangulator2;
typedef Triangulator<vec3>		Triangulator3;
```

src/cinder/Triangulate.cpp
```diff
@@ -108,22 +108,87 @@ void Triangulator::addPolyLine( const vec2 *points, size_t numPoints )
...
+void Triangulator3d::allocate()
+{
+	mAllocated = 0;
+
+	TESSalloc ma;
+	memset( &ma, 0, sizeof(ma) );
+	ma.memalloc = stdAlloc;
+	ma.memfree = stdFree;
+	ma.userData = (void*)&mAllocated;
+	ma.extraVertices = 2560; // realloc not provided, allow 256 extra vertices.
```
MikeGitb: Comment doesn't fit the code (256 <-> 2560 ) - is the 0 there deliberate?

src/cinder/Triangulate.cpp
```diff
@@ -108,22 +108,87 @@ void Triangulator::addPolyLine( const vec2 *points, size_t numPoints )
...
+
+	mTess = shared_ptr<TESStesselator>( tessNewTess( &ma ), tessDeleteTess );
```
MikeGitb: This seems a bit fishy to me. I don't know how exactly the tess library works, but you are passing the address of a local variable (`ma`) to a function, that will take that address and store it in a heap allocated `TESStesselator` object which will then will be used it long after the local object is gone.

src/cinder/Triangulate.cpp
drewish: Yeah been noticing some other weird behavior since that structure has an [`outOfMemory`](https://github.com/cinder/Cinder/blob/master/src/libtess2/tess.h#L53) member that isn't explicitly initialized to 0. So sporadically it seems to fail once it reuses memory the OS hasn't zeroed out.

src/cinder/Triangulate.cpp
```diff
@@ -108,22 +108,87 @@ void Triangulator::addPolyLine( const vec2 *points, size_t numPoints )
...
+{
+	if( numPoints > 0 )
+		tessAddContour( mTess.get(), 3, &points, sizeof(vec3), (int)numPoints );
+}
+
+TriMesh Triangulator3d::calcMesh( Triangulator::Winding winding )
+{
+	TriMesh result( TriMesh::Format().positions( 3 ) );
+
+	tessTesselate( mTess.get(), (int)winding, TESS_POLYGONS, 3, 3, 0 );
```
drewish: This returns a 1 for success but we're not checking it leading to very strange results when you hit the outOfMemory bug I mentioned in the other comment.

src/cinder/Triangulate.cpp
drewish: Or rather, the TESStesselator's outOfMemory value wasn't being initialized. It seems like we could switch the malloc to a calloc in our stdAlloc.

But to your original point I'm not sure why that number changed or if it was accidental.

src/cinder/Triangulate.cpp
drewish: yeah agreed it seem suspect. it looks like `tessNewTess` [stores a reference](https://github.com/cinder/Cinder/blob/master/src/libtess2/tess.c#L485) to it in 
 the mTess we get back. i'm not sure why it doesn't cause more obvious problems.

drewish:
@vinjn yeah that was what i'd looked at initially but i wasn't clear how you'd get a scalar `2` or `3` to pass as a parameter out of the template type.

vinjn:
```C++
int getVecDim(vec3) { return 3;}
int getVecDim(vec2) { return 2;}
```
@drewish

drewish:
I'm thinking this might be hard to generalize for random 3d input. I've been getting really strange behavior with multiple contours where adding a second contour causes it to return no output. I think this is due to the way libtess builds the arrangement by projecting everything onto a single sweep plane. It looks like you can influence the choice of plane by passing in a normal. So I think the only way it would reliably work in 3d would be to either require a normal be provided when calculating the mesh or just limit it to a single input contour.

src/cinder/Triangulate.cpp
drewish: @MikeGitb what seems like a sane way to handle the TESSalloc? it seems like it should be an instance variable but then you'd need to include `tesselator.h` in `Triangulate.h`. so pointer or shared_ptr?

MikeGitb: @drewish: If backwards compatibility was not an issue, I'd wrap the `mTess` pointer, the allocator and the used `tess...` functions and in particular the `allocate` function into a - forward declared - class and store a single `std::unique_ptr`. As things stand just creating a member variable for the allocator as you suggested seems to be the best solution. 
However, as it is a private/protected variable that isn't shared with anything, I'd not use a `shared_ptr` but a `unique_ptr`.

src/cinder/Triangulate.cpp
```diff
@@ -108,22 +108,87 @@ void Triangulator::addPolyLine( const vec2 *points, size_t numPoints )
...
+	return result;
+}
+
+TriMeshRef Triangulator3d::createMesh( Triangulator::Winding winding )
+{
+	TriMeshRef result = make_shared<TriMesh>( TriMesh::Format().positions( 3 ) );
+
+	tessTesselate( mTess.get(), (int)winding, TESS_POLYGONS, 3, 3, 0 );
+	result->appendPositions( (vec3*)tessGetVertices( mTess.get() ), tessGetVertexCount( mTess.get() ) );
+	result->appendIndices( (uint32_t*)( tessGetElements( mTess.get() ) ), tessGetElementCount( mTess.get() ) * 3 );
```
MikeGitb: that cast from `int*` (returned by tessGetElements) to `uint32_t*` is actually undefined bahavior / completely illegal

src/cinder/Triangulate.cpp
```diff
@@ -108,22 +108,87 @@ void Triangulator::addPolyLine( const vec2 *points, size_t numPoints )
...
+void Triangulator3d::addPolyLine( const PolyLine3f &polyLine )
+{
+	if( polyLine.size() > 0 )
+		tessAddContour( mTess.get(), 3, &polyLine.getPoints()[0], sizeof(vec3), (int)polyLine.size() );
+}
+
+void Triangulator3d::addPolyLine( const vec3 *points, size_t numPoints )
+{
+	if( numPoints > 0 )
+		tessAddContour( mTess.get(), 3, &points, sizeof(vec3), (int)numPoints );
```
MikeGitb: Is the `&` in front of `points` correct here? I don't think we want a pointer to a pointer here

src/cinder/Triangulate.cpp
```diff
@@ -66,15 +73,16 @@ Triangulator::Triangulator()
...
-	TESSalloc ma;
-	memset( &ma, 0, sizeof(ma) );
-	ma.memalloc = stdAlloc;
-	ma.memfree = stdFree;
-	ma.userData = (void*)&mAllocated;
-	ma.extraVertices = 2560; // realloc not provided, allow 256 extra vertices.
-
-	mTess = shared_ptr<TESStesselator>( tessNewTess( &ma ), tessDeleteTess );
+
+	mAlloc = shared_ptr<TESSalloc>( new TESSalloc() );
```
drewish: @MikeGitb I tried doing a shared_ptr on this but got a rather cryptic error:
> /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/memory:2523:27: Invalid application of 'sizeof' to an incomplete type 'TESSalloc'

src/cinder/Triangulate.cpp
```diff
@@ -102,28 +110,70 @@ void Triangulator::addPolyLine( const PolyLine2f &polyLine )
...
 {
 	if( numPoints > 0 )
-		tessAddContour( mTess.get(), 2, &points, sizeof(vec2), (int)numPoints );
+		tessAddContour( mTess.get(), 2, points, sizeof(vec2), (int)numPoints );
+}
+
+void Triangulator::addPolyLine( const PolyLine3f &polyLine )
+{
+	if( polyLine.size() > 0 )
+		tessAddContour( mTess.get(), 3, &polyLine.getPoints()[0], sizeof(vec3), (int)polyLine.size() );
```
drewish: Any reason not to just do `polyLine.getPoints().data()` here?

src/libtess2/tess.c
```diff
@@ -482,6 +482,7 @@ TESStesselator* tessNewTess( TESSalloc* alloc )
 	if ( tess == NULL ) {
 		return 0;          /* out of memory */
 	}
+	tess->outOfMemory = 0;
```
drewish: Since there was already one other edit to this lib this seemed like the cleanest way to make sure this value was initialized.

drewish:
I went ahead and folded this back into one class. The only difference is flow is really in the mesh outputting so I just added a parallel set of calc/createMesh3d functions. The 3d version takes a normal and I think in the docs we just need to make it clear that you only want to tesselate things on the same plane unless you know what you're doing.

src/cinder/Triangulate.cpp
MikeGitb: I assume the problem is that Triangulate has an implicitly defined destructor. That means that the compiler tries to generate the destructor in every translation unit that includes the header file. However, in most translation units, it doesn't have access to the Definition of TESSalloc which it apparently needs  for the destructor of mAlloc. 
Try defining an empty destructor in Triangulate.cpp

src/cinder/Triangulate.cpp
MikeGitb: Nope, that also seems more logical to me (probably pre c++11 code when vector didn't have the `data()` member function). I'd usually also not write `sizeof(vec3)`, but `sizeof(*polyLine.getPoints().data())` as this makes the code more generic and less prone to refactoring erors (but admittedly also more ugly)

src/cinder/Triangulate.cpp
MikeGitb: If you really want to use a shared_ptr instead of a unique_ptr may I suggest using `std::make_shared`?

drewish: sorry that was a typo, i meant to write unique_ptr. i agree that makes more sense so i'll try adding a destructor and see if that helps.

drewish: At least with the destructor I'm getting an error that's understandable:
> /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/memory:2523:27: Invalid application of 'sizeof' to an incomplete type 'TESSalloc'

MikeGitb: Isn't that the same message as you got before?

MikeGitb:
I was wrong btw.: The cast from `int*` to `uin32_t*` seems to be legal (at least on systems, where `uint32_t` is the same as `unsigned int`

drewish:
Since over on #1698 it sounded like PolyLine3 might not be around long I switched this interface to just use `std::vector<vec3>` instead.

src/cinder/Triangulate.cpp
```diff
@@ -96,34 +104,72 @@ void Triangulator::addPath( const Path2d &path, float approximationScale )
...
+		tessAddContour( mTess.get(), 3, points, sizeof(vec3), (int)numPoints );
 }
 
 TriMesh Triangulator::calcMesh( Winding winding )
 {
 	TriMesh result( TriMesh::Format().positions( 2 ) );
-	
-	tessTesselate( mTess.get(), (int)winding, TESS_POLYGONS, 3, 2, 0 );
+
+	float normal[3] = { 0, 0, 1 };
```
drewish: So I'd started passing this in to avoid computing it, but I need to test it out. I think the old behavior would flip the normal to use a (0,0,-1) if the winding order was backwards. I need to add test cases for that.

MikeGitb:
Any reason this didn't get merged?

drewish:
Not that I'm aware of. Happy to rebase it or make any changes that are necessary.

andrewfb:
Sorry for the unreasonable delay on this. Two quick questions as my head is no longer in this code unfortunately. First, should we consider moving to the latest here? https://github.com/memononen/libtess2 And second, does the tesselator do the right thing if I supply a mixture of 2D and 3D contours as input?

PR #1693, opened 2016-12-27
fix include utilities

MikeGitb:
cinder/utilities.cpp should include the `fstream` header (leads to compilation eror on VS2015 U3)

This pr also makes a few small formatting changes:
- delete some trailing whitespaces
- the opening `{` of a function is always on a new line

### DIFF
--- a/src/cinder/Utilities.cpp
+++ b/src/cinder/Utilities.cpp
@@ -76,7 +76,7 @@ std::vector<std::string> split( const std::string &str, const std::string &separ
 {
 	vector<string> result;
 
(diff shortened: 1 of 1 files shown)


MikeGitb:
Please note, that the build failure in travis is due to a timeout and not due to an actual build error

richardeakin:
Thanks for this, and also for keeping the bug fix separate to the whitespace improvements.

Yes we need to get the freetype warnings reduced to prevent the Xcode build logs from overloading, hopefully soon. Also hoping to get CI on windows so that things like this missing include get caught (seems to be specific to vc140 as it builds fine for me using vc120).

PR #1695, opened 2016-12-28
Style and indentation fixes.

PetrosKataras:
Moving between Linux / OS X style got mixed up. This should bring things together in preparation for the scrubbing fixes that will follow.


MikeGitb:
This PR adds/keeps a lot of training whitespaces. Could you please remove them too?

MikeGitb:
Thank you very much!

PetrosKataras:
Thanks for reviewing!

MikeGitb:
Doesn't take care of line ending changes automatically? On windows it will always check-in with unix style line endings and checkout with whatever type you specify.

PetrosKataras:
I am not entirely sure about this to be honest but `git diff` was showing me  CRLF endings ( i.e ^M ) which I managed to get around it with `git config --global core.autocrlf true` but I m not sure if this would display it properly on someone else's editor and since Vim was recognizing these files as DOS I decided to reformat them to Unix which should be the preferred option for us.

richardeakin:
I do think we want to try to use Unix line endings across the board, and its better earlier than later IMO as it  marks the entire file as modified. Very nice that it is in a separate commit though, that way you can look before and after the whitespace change if needed.

PR #1698, opened 2016-12-31
PolyLine3d doesn't really work

drewish:
Demonstrates a linking bug I've encountered:
```
Undefined symbols for architecture x86_64:
  "cinder::PolyLineT<glm::tvec3<float, (glm::precision)0> >::reverse()", referenced from:
      ____C_A_T_C_H____T_E_S_T____11() in PolyLineTest.o
  "cinder::PolyLineT<glm::tvec3<float, (glm::precision)0> >::reversed() const", referenced from:
      ____C_A_T_C_H____T_E_S_T____11() in PolyLineTest.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

[According to Travis](https://travis-ci.org/cinder/Cinder/builds/187942442) this only affects the Xcode builds.

If anyone know why that's happening I'd be happy to add the fix to this PR.

### DIFF
--- a/src/cinder/gl/draw.cpp
+++ b/src/cinder/gl/draw.cpp
@@ -391,7 +391,7 @@ void draw( const PolyLine2 &polyLine )
 	ctx->popVao();
 }
 
-void draw( const PolyLine3 &polyLine )
+void draw( const std::vector<vec3> &points, bool isClosed )
 {
 	auto ctx = context();
 	const GlslProg* curGlslProg = ctx->getGlslProg();
@@ -400,7 +400,6 @@ void draw( const PolyLine3 &polyLine )
 		return;
 	}
 	
-	const vector<vec3> &points = polyLine.getPoints();
 	VboRef arrayVbo = ctx->getDefaultArrayVbo( sizeof(vec3) * points.size() );
 	arrayVbo->bufferSubData( 0, sizeof(vec3) * points.size(), points.data() );
 
@@ -415,7 +414,7 @@ void draw( const PolyLine3 &polyLine )
 
 	ctx->getDefaultVao()->replacementBindEnd();
 	ctx->setDefaultShaderVars();
-	ctx->drawArrays( ( polyLine.isClosed() ) ? GL_LINE_LOOP : GL_LINE_STRIP, 0, (GLsizei)points.size() );
+	ctx->drawArrays( ( isClosed ) ? GL_LINE_LOOP : GL_LINE_STRIP, 0, (GLsizei)points.size() );
 	ctx->popVao();
 }
 
--- a/test/unit/src/PolyLineTest.cpp
+++ b/test/unit/src/PolyLineTest.cpp
@@ -179,3 +179,4 @@ TEST_CASE("PolyLine2f")
 		CHECK( poly.calcCentroid() == vec2( 0.5, 0.5 ) );
 	}
 }
+
--- a/include/cinder/PolyLine.h
+++ b/include/cinder/PolyLine.h
@@ -99,8 +99,5 @@ class PolyLineT {
 typedef PolyLineT<vec2>		PolyLine2;
 typedef PolyLineT<vec2>		PolyLine2f;
 typedef PolyLineT<dvec2>	PolyLine2d;
-typedef PolyLineT<vec3>		PolyLine3;
-typedef PolyLineT<vec3>		PolyLine3f;
-typedef PolyLineT<dvec3>	PolyLine3d;
 
 } // namespace cinder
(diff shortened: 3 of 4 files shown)


test/unit/src/PolyLineTest.cpp
```diff
@@ -8,89 +8,128 @@ using namespace ci;
 using namespace ci::app;
 using namespace std;
 
-TEST_CASE("PolyLine", "Orientation")
```
drewish: I didn't really understand how catch worked when I set this up. It seems better now to make Orientation it's own section rather than a top level case.

drewish:
Ignoring whitespace is probably the best way to view these changes: https://github.com/cinder/Cinder/pull/1698/files?w=1

drewish:
Ah, today I learned about `template class`. Now I see down at the bottom of Polyline.cpp:
```c++
template class PolyLineT<vec2>;
template class PolyLineT<dvec2>;
```
https://github.com/cinder/Cinder/blob/master/src/cinder/PolyLine.cpp#L347-L348

Adding in:
```c++
template class PolyLineT<vec3>;
template class PolyLineT<dvec3>;
```
Triggers some build errors. A couple of there were easy to fix but `calcCentroid()` and `contains()` are a bit more complicated.

richardeakin:
Hm, it appears that I goofed when doing the port from old cinder-math to GLM - [here](https://github.com/cinder/Cinder/blob/0bde389eac5aed343706fc56914b7a4f3fb3fefb/include/cinder/PolyLine.h#L77-L78) you can see that in 0.8.6 we only had `PolyLine2f` and `PolyLine2d`.

That said, what are your use cases for a `vec3` version of PolyLine? I'm not sure if the math in all of the methods holds up in some of PolyLine's methods, especially the boolean operators, but I think we could break those out into standalone methods that only operate on the 3d version if we wanted.

Aside from that, improving unit tests is always welcome. I also see that in some places in `vec2` is used in PolyLine.cpp, which will either be wrong or won't compile when using `PolyLine2d`, and we should in any event fix those cases.

drewish:
Yeah my needs are pretty minimal--as demonstrated by the fact that I didn't notice the bug until I'd tried to reverse one. If there's any easy way to make some methods 2d only I think that'd be a great start.

drewish:
I'm also happy to split this up into two PRs so we could get the `vec2` -> `T` fixes and 2d unit tests in, then look at getting the 3d parts working later.

richardeakin:
> I'm also happy to split this up into two PRs so we could get the vec2 -> T fixes and 2d unit tests in, then look at getting the 3d parts working later.

This would be great, that way we can get in the non-controversial part and let the discussion on whether PolyLineT should support `vec3` / `dvec3`. At minimum, I think that would take some breaking changes to break out the boolean operators, so not at all sure there.

andrewfb:
To my mind `PolyLine` was always a 2D use-case. I think the question is, what kind of functionality would we be able to provide over a simple `vector<vec3>` for a 3D case? If we did not assume coplanarity, I think the answer is not very much. If we did assume that, there's probably more we could provide, but it might warrant a separate class, since much of the current functionality wouldn't make sense.

drewish:
I was using it along with #1691 but you're right I could ditch the PolyLine3f side in favor of simple vectors. I'd been doing that after realizing it wasn't actually supported.

If the consensus is to ditch the 3d part I'm happy to close this and open up another PR to remove it.

richardeakin:
Seems we should just remove the stray vec3 typedefs at the bottom of PolyLineT for now, until we have a good reason to implement something like that.

drewish:
That sounds good to me. I'll change this around tonight.

drewish:
I updated this PR. We'll have to make some change to `void draw( const PolyLine3 &polyLine )` I opted to remove it but it could be nice to offer `void draw( const std::vector<vec3> &polyLine );` instead. Open to feedback.

richardeakin:
Ah, hadn't realized that `PolyLine3` was used anywhere. I think switching that convenience draw method to take a `const std::vector<vec3> &` makes sense to me, instead of removing it.

src/cinder/gl/draw.cpp
```diff
@@ -391,34 +391,6 @@ void draw( const PolyLine2 &polyLine )
...
-	ScopedBuffer bufferBindScp( arrayVbo );
-	int posLoc = curGlslProg->getAttribSemanticLocation( geom::Attrib::POSITION );
-	if( posLoc >= 0 ) {
-		enableVertexAttribArray( posLoc );
-		vertexAttribPointer( posLoc, 3, GL_FLOAT, GL_FALSE, 0, (const GLvoid*)nullptr );
-	}
-
-	ctx->getDefaultVao()->replacementBindEnd();
-	ctx->setDefaultShaderVars();
-	ctx->drawArrays( ( polyLine.isClosed() ) ? GL_LINE_LOOP : GL_LINE_STRIP, 0, (GLsizei)points.size() );
```
richardeakin: Hm, might also require an extra argument to determine if it is `isClosed` or not (why not add a default while there).

drewish:
@richardeakin that makes sense, I've updated it accordingly.

PR #1699, opened 2017-01-01
Linux: Recompile boost libs on desktop to enable support for Cinder as a shared lib.

PetrosKataras:
The following command was used to recompile boost system and boost filesystem v1.60 :  

`./b2 cxxflags="-fPIC -D_GLIBCXX_USE_CXX11_ABI=0" link=static debug-symbols=off optimization=space --with-system --with-filesystem`

The `D_GLIBCXX_USE_CXX11_ABI` define is necessary in order to allow building on systems with gcc versions < 5.1 . See [here](https://gcc.gnu.org/onlinedocs/libstdc%2B%2B/manual/using_dual_abi.html) and [here](http://stackoverflow.com/questions/34571583/understanding-gcc-5s-glibcxx-use-cxx11-abi-or-the-new-abi) for more info.

Ideally though we want to move away from disabling this option ASAP. This can be achieved either when we remove boost as a dependency or if we set a minimum supported gcc version thats greater /equal than 5.1 on all Linux platforms that we support.

There is a separate PR coming that actually enables the support for building Cinder as a shared lib and depends on this one.


PetrosKataras:
This seems to fail due to different binutils versions between 16.04 and 14.04 as far as I understand so there might be more issues with trying to maintain this backwards compatibility..

PetrosKataras:
Closing this one then until we figure out what is the proper way to go forward.

MikeGitb:
Why is it necessary to touch this option at all? Isn't it possible to just use whatever abi version a particular distribution chose as it's default? Also, I wonder, if boost should be part of cinder on Linux systems at all, as it is usually provided via the system packet manager anyway.

MikeGitb:
Sorry, for my previous post. I didn't realize this was about the binary distribution of boost library. 
However, my main question remains: Why should the boost library be distributed along cinder at all? Pretty much all Linux distros I'm aware of provide the boost library as a system package.

PetrosKataras:
The reasons its necessary are described in the links that are provided but long story short we need it due to the fact that boost is shipped as a precompiled static lib with Cinder and this makes it dependent on the gcc version that it was compiled with ( i.e gcc < 5.1 or gcc >= 5.1 ). 

You cannot for example build boost with 4.9 and then try to build a Cinder app that links to boost with 5.1 because of ABI changes. This won't  work except if you explicitly ask for the old ABI through the above macro.

Boost  comes with Cinder on Linux as it does on the other platforms too because ( at least to my understanding ) of having finer control on what components are used exactly but also because its easier to patch/update in this case. 

That said, personally, I can understand that this might seem a bit controversial for Linux platforms and it can actually cause a lot of headaches ( like this one for example ) but on the other hand I can also see the merit on maintenance and freedom behind this approach. 

I think ideally boost should be removed as a dependency all together as soon as possible and AFAIK this is the current plan as soon as compilers catch up.

MikeGitb:
@PetrosKataras: That is precisely my point, if we just use the system provided boost version, we get the correctly configured one for free and there is no danger of version conflicts if the application that uses cinder wants to use boost itself.

PetrosKataras:
In the case of boost I totally agree. The way it is right now causes a lot of problems on Linux especially in the case it happens to have boost already installed ( not uncommon since its a dependency for some bins in Linux ) and also in the case that you want to link your app with Cinder but also with another library that depends on boost. In both of those cases most likely things will fail with the current configuration. It happened to me for example when linking a Cinder app with CGAL and I had to remove Cinder's boost to make things work.

MikeGitb:
@PetrosKataras: Thanks, I wasn't aware of that option. One possiblity would be to treat boos like the msw libraries like zlib, which are also only used on windows.

PR #1700, opened 2017-01-01
Linux fix shared lib option. 

PetrosKataras:
With this and #1699 we have support for building/linking Cinder as a shared lib on Linux desktop.

For gcc < 5.1 we disable the new ABI since this is necessary for older systems. See comments on #1699 for more info.

EDIT: Its seems that #1699 needs more thinking so I closed it for now but this one should still be relevant and ok to merge in preparation for proper support of Cinder as a shared lib on Linux desktop.


richardeakin:
Thanks, the discussion around #1699 is ongoing, especially what we'll be doing for boost libs on linux in the future. At the same time, it is looking like a dll option of shared lib, on both POSIX and Windows, will be a focal point during the next release cycle, so we can circle around then.

Merging this as it seems that you can get a dylib working on linux now if you specify to use the system boost binaries.

PR #1706, opened 2017-01-06
PolyLine template cleanups

drewish:
I split some of the 2D fixes out of #1698 so they can go in while we discuss the 3D parts. I add some quick tests for the methods I was touching.

The tests has a bunch of space indenting that I cleaned up so probably easiest to see what's changing there by [ignoring whitespace](https://github.com/cinder/Cinder/pull/1706/files?w=1)


richardeakin:
Thanks for breaking this out.

PR #1707, opened 2017-01-06
Paulhoux distance to curve update

andrewfb:
This incorporates and supersedes #1615, which is @paulhoux's distance functions for `Path2d` and `Shape2d`. That work revealed some numerical instabilities in our implementation of `Path2d::contains()` which this PR replaces with an implementation derived from Skia. It also adds a new visual test, `test/ShapeTest`, and adds some preliminary Path2d unit tests.


include/cinder/Path2d.h
```diff
@@ -109,8 +109,27 @@ class Path2d {
...
-	bool	contains( const vec2 &pt ) const;
+	//! Returns whether the point \a pt is contained within the boundaries of the Path2d. If \a evenOddFill is \c true (the default) then Even-Odd fill rule is used, otherwise, the Winding fill rule is applied.
+	bool	contains( const vec2 &pt, bool evenOddFill = true ) const;
+
+	//! Returns the minimum distance from point \a pt to the Path2d
+	float	calcDistance( const vec2 &pt ) const;
+	//! Returns the minimum distance from point \a pt to segment \a segment
+	float	calcDistance( const vec2 &pt, size_t segment ) const { return calcDistance( pt, segment, 0 ); }
+	//! Returns the minimum distance from the Shape2d to point \a pt. For points inside the Shape2d, the distance is negative.
+	float	calcSignedDistance( const vec2 &pt ) const
```
richardeakin: I think moving these type of methods, and the similar one in `Shape2d`, would clean up the interface a tad, which is already quite a bit to take in as they are fairly complex classes and there isn't much to be gained by in-lining them.

include/cinder/Path2d.h
```diff
@@ -141,6 +160,14 @@ class Path2d {
 	void	arcHelper( const vec2 &center, float radius, float startRadians, float endRadians, bool forward );
 	void	arcSegmentAsCubicBezier( const vec2 &center, float radius, float startRadians, float endRadians );
 	
+	//! Returns the minimum distance from point \a pt to segment \a segment. The \a firstPoint parameter can be used as an optimization if known, otherwise pass 0.
+	float	calcDistance( const vec2 &pt, size_t segment, size_t firstPoint ) const;
+
+	int calcWinding( const ci::vec2 &pt, int *onCurveCount ) const;
```
richardeakin: This guy looks so lonely without dox. :)

test/unit/vc2013/unit.vcxproj
```diff
@@ -271,7 +270,8 @@
     <ClCompile Include="..\src\SystemTest.cpp" />
     <ClCompile Include="..\src\TestMain.cpp" />
     <ClCompile Include="..\src\UnicodeTest.cpp" />
-    <ClCompile Include="..\src\Utilities.cpp" />
+    <ClCompile Include="..\src\PolyLineTest.cpp" />
```
drewish: Thanks for picking adding this to the Windows projects

richardeakin: Hm, just noticed that this somehow removed Utilities.cpp from the unit.vcxproj file and .filters, perhaps when resolving a merge conflict?

andrewfb: Thanks - this should be addressed in recent commits, as well as missing units on Linux cmake project (which I committed blindly; if I missed something there Travis should catch it)

PR #1708, opened 2017-01-06
Use mStartDrawFn and mFinishDrawFn for CINDER_COCOA_TOUCH.

clinthidinger:
The CINDER_COCOA_TOUCH section of RendererGl is not using mStartDrawFn or mFinishDrawFn.  I'm assuming that it should be.  The fix makes it identical to the CINDER_MAC section.


andrewfb:
Thanks for this find & fix

PR #1710, opened 2017-01-09
Expose native touch event

clinthidinger:
Access is already exposed to the native touch (UITouch*) on iOS through TouchEvent::Touch::getNative().  This is useful for getting the force/pressure data for the apple pencil and other data.  In addition, I'd like to be able to get the UIEvent as well in order to be able to call predictedTouchesForTouch and coalescedTouchesForTouch at the application level.  Multi-touch is very laggy on iOS, which breaks the illusion of direct manipulation.  The predictive touches help out a lot.

### DIFF
--- a/include/cinder/app/TouchEvent.h
+++ b/include/cinder/app/TouchEvent.h
@@ -76,10 +76,7 @@ class TouchEvent : public Event {
 	TouchEvent()
 		: Event()
 	{}
-	TouchEvent( const WindowRef &win, const std::vector<Touch> &touches )
-		: Event( win ), mTouches( touches ), mNative( nullptr )
-	{}
-	TouchEvent( const WindowRef &win, const std::vector<Touch> &touches, void *native )
+	TouchEvent( const WindowRef &win, const std::vector<Touch> &touches, void *native = nullptr )
 		: Event( win ), mTouches( touches ), mNative( native )
 	{}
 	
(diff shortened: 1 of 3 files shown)


richardeakin:
Sounds good to store the actual UIEvent, however not sure about adding an extra constructor that is only used on iOS.  Seems a bit of a shortcut to add a new TouchEvent constructor, though at the same time the entire Event system is bound for a rewrite to be more cross-platform. I'd suggest keeping only one constructor, and changing it to:

```
TouchEvent( const WindowRef &win, const std::vector<Touch> &touches, void *native = nullptr )
```

(or going through and updating each platform to pass in what they see as fit for the native pointer, which is of course more work).

clinthidinger:
I am currently passing in the  NSEvent* for OS X in CinderViewMac.mm.  

Should I go ahead and do another push with the default param instead of the extra ctor?

richardeakin:
Ah I see, missed that. Yes IMO one constructor here is better than two, otherwise it is a nice and easy way to expose the platform functionality you're looking for so you have my vote.

clinthidinger:
I made the push to remove the extra ctor in favor of default param.

notlion:
👍  I didn't know about `predictedTouchesForTouch` before. Seems like this could be really useful when the touch event rate drops below 60Hz.

andrewfb:
Thanks for this

Issue #1158, opened 2015-10-10
Disable /Gm (minimal rebuild) for newly created projects.


paulhoux:
The /Gm switch disables multi-processor building. In practice, building on all available processors reduces compile times more than any minimal rebuild could.

richardeakin:
It would be nice to see numbers around this, especially including when you modify a file and recompile. I've found that in that case, the minimal rebuild is faster and ideal for user app projects. But I haven't done any close measurements with VS 2015 and alot has changed around compile times in that version.

paulhoux:
OK, did some profiling. I cleanly rebuilt Cinder Release Win32 several times, using the build timer:

Minimal Rebuild, No Multiprocessor | No Minimal Rebuild, Multiprocessor
--------------------|------------------
200.095 sec|67.376 sec
202.173 sec|65.190 sec
201.552 sec|66.915 sec

Single file (AppBase.cpp) forcing it to rebuild be adding a space

Minimal Rebuild, No Multiprocessor | No Minimal Rebuild, No Multiprocessor | No Minimal Rebuild, Multiprocessor
--------------------|------------------|-------
3.692 sec|3.582 sec|3.306 sec
3.524 sec|3.273 sec|3.329 sec
3.511 sec|3.326 sec|3.294 sec

If these tests are representative, it is clear that enabling multiprocessor compilation outweighs minimal rebuild, especially when compiling more than just a few files.

Tested on Intel Core i7-4720HQ @ 2.60GHz (4 cores, hyperthreading)

richardeakin:
Hm, looks like `/Gm` isn't giving us anything over `/MP`. Though your numbers are for rebuilding libcinder, so I tried with the FallingGears sample to see what they look like for a user app (which point at a similar conclusion). Tested on an Intel i7-4790 4gz with ssd drive, VS2015 running v120 toolset:

```
FallingGears sample

/Gm rebuild:
16.9s
20.3s
20.0s

/Gm build /w added line to SceneController.cpp
3.76s
4.3
4.2

/MP rebuild:
8.8s
8.7s
8.5s

/MP build /w added line to SceneController.cpp
4.0
3.5
3.45
```

A full rebuild is nearly twice as fast 

For libcinder, I am seeing that neither `/MP` nor `/Gm` are set. We should probably add `/MP` for all targets.

[...]

PR #1711, opened 2017-01-09
Visual Studio project file improvements

richardeakin:
These are a bunch of small improvements to both `cinder.vcxproj` files and VS tinderbox templates, for both desktop and UWP.  The primary thing I addressed, as proposed in the #1429 comments, was that all build outputs for all configurations and platforms end up in the following folders:

Output Directory: $(ProjectDir)build\$(Platform)\$(Configuration)\
Intermediate Directory: $(ProjectDir)build\$(Platform)\$(Configuration)\intermediate\

This way, all of your build files can be easily blown away by deleting the folder called `build`, which is in line with how things usually happen on OS X and with CMake.

The next was to make the optimization settings a bit more consistent, as it was proposed in #1158 to always use `/MP` and disable `/Gm` flags. This was already happening, although in some places (vc2013/cinder.vcxproj`) I believe that `/MP` was implicit, so I marked it explicitly so that things are consistent.

Lastly, in some places we were disabling Whole Program Optimization (building libcinder) and in others (templates) it was enabled for some targets while not others. I've disabled it throughout, user apps can enable it as they see fit though I don't think it is improving much at current, from my brief tests.

As always, open to suggestions, while I'm looking at this concerning a couple other ongoing projects.

### DIFF
--- a/blocks/__AppTemplates/__Foundation/vc2013/foundation.vcxproj
+++ b/blocks/__AppTemplates/__Foundation/vc2013/foundation.vcxproj
@@ -37,14 +37,14 @@
     <UseDebugLibraries>false</UseDebugLibraries>
     <PlatformToolset>v120</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
-    <WholeProgramOptimization>true</WholeProgramOptimization>
+    <WholeProgramOptimization>false</WholeProgramOptimization>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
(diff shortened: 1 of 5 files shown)


MikeGitb:
Just wanted to mention that I did a short test and compiled the desktop solution in vc2013 with VS2015 (all build configurations) just fine. The unittests also compile. When I compile the samples I get some warnings about *"'/Gm' is incompatible with multiprocessing; ignoring /MP switch"* though.

richardeakin:
@MikeGitb thanks for testing!  We haven't updated the samples or tests and a couple of us have discussed waiting until 0.9.2 to do that, and just generate fresh new projects at that point so they are all in line.

What would be good to test now are the [TinderBox](https://libcinder.org/docs/guides/tinderbox/index.html) templates - I made some modifications to those so that projects created in the future follow this new pattern of output directories and optimizer flags. Shouldn't see warnings there, I hope.

Issue #1306, opened 2016-02-07
ShaderPreprocessor doesn't work on android


richardeakin:
Not a fault of android, just that android doesn't use direct file access like other platforms. Problem is that `ShaderPreprocessor` tries to directly access files with `ifstream( fullFilePath )`, instead of getting routed through the android assets mechanisms.  I propose we add a std::function based 'filepath resolver' callback that returns either a `ci::DataSourceRef` or `ci::IStreamRef`, providing a workable default that uses `ci::loadFile` but also allows users to override when needed.

richardeakin:
include path handler added in #1721, GlslProg (or something else) should add a custom handler to load strings via android's asset file system stuff.

PR #1721, opened 2017-01-18
ShaderPreprocessor features and bug fixes

### LINKED ISSUE #850: ShaderPreprocessor debug line wrong
When debugging a shader with compile errors the exception description will tell you what the offending line is, however the newly introduced `ShaderPreprocessor` is mucking this up.

~~Firstly, the line number is greater by one than it should be when you give it glsl that already has a `#version` string and no includes (like [this simple shader](https://github.com/cinder/Cinder/blob/glNext/samples/_opengl/Cube/assets/shader.frag)). I'm not sure exactly which part, but it seems to be picking it up the off-by-one in the [ShaderPreprocessor::parseDirectives()](https://github.com/cinder/Cinder/blob/glNext/src/cinder/gl/ShaderPreprocessor.cpp#L69-L105) pass.~~ fixed with #895.

Secondly, any additions made in `parseDirectives()` do not correct the line number via the `#line` directive. Also for this to work in conjunction with the #line directives added for includes ([here](https://github.com/cinder/Cinder/blob/glNext/src/cinder/gl/ShaderPreprocessor.cpp#L123)), `parseDirectives()` would have to happen before parsing includes, whereas it is currently performed afterwards.

richardeakin:
This is a hefty pass on `ci::gl::ShaderPreprocessor` to address a number of issues:

* Fixes debug error lines, which were wrong in many scenarios (included files, custom defines, `#version 150` vs `330`, lots of reasons). Fixes #850.
* Adds support for a custom include handler by way of `signals::Signal`. This should make it possible to handle #1306 without direct modification to the `ShaderPreprocessor`, as well as let you load in include sources from a library that provides them as embedded strings, for example.
* Made it easier to reuse your own custom `ShaderPreprocessor` when creating `GlslProgs`. Now you can pass a shared_ptr to your own preprocessor to the `GlslProg::Format` and it will use that instead of its own.
* Adds some initial unit tests
* Refactoring: cleaned up the duplicate while( getline() ) loops, reduced the workload that parsing directives is doing.

There's also one more neat feature added, which AFAIK only works on NVidia. If you set `ShaderPreprocessor::setUseFilenameInLineDirectiveEnabled( true )`, then the debug lines will contain the name of the currently processed source file. This isn't according to spec, but NVidia seems to have implemented the `#line` directive like ANSI C anyway. So, by default you would see something like:

```
|error  | ShaderPreprocessorTestApp::testSeparateShaderPreprocessor[144] exception caught, exception type: class cinder::gl::GlslProgCompileExc, what: FRAGMENT: 1(3) : error C0159: invalid char 'b' in integer constant suffix
1(11) : error C0000: syntax error, unexpected identifier, expecting ',' or ';' at token "a"
```

And if you enable that setting, instead you get:

```
|error  | ShaderPreprocessorTestApp::testSeparateShaderPreprocessor[144] exception caught, exception type: class cinder::gl::GlslProgCompileExc, what: FRAGMENT: commonSimple.glsl(3) : error C0159: invalid char 'b' in integer constant suffix
simple.frag(11) : error C0000: syntax error, unexpected identifier, expecting ',' or ';' at token "a"
```

Much nicer. Since it isn't according to spec though, it is disabled by default. I'm curious if there are any Linux or AMD people out there, if they can give it a try and let me know what you see with this (you can run the [dev test app](https://github.com/richardeakin/Cinder/blob/shaderpp_improvements/test/_opengl/ShaderPreprocessorTest/src/ShaderPreprocessorTestApp.cpp) and hit 'n' then 'r' to reload to test).


richardeakin:
Thanks for the review @MikeGitb, will address those concerns in the next couple days. Also looks like some of the unit tests are failing on POSIX that are likely to be because filepaths are different there, will have to do something about that as well.

richardeakin:
Updates from feedback and unit test fixes up, I believe this is good to go. Current unit test errors are due to timeouts, that problem has been fixed in the master branch

drewish:
(Probably obvious but you could rebase on master to pick up those fixes)

PR #1724, opened 2017-01-20
Fix for ComputeBasic sample freezing on window resize

MarcinKantochAMD:
ComputeBasic sample in Vulkan Explicit group is freezing on window resize. This commit fixes the issues. 

The issue is caused by draw() method being called before update() when a window resize happens. draw() method submits a command buffer to a graphics queue, but the rendering operation is configured to wait on mComputeDoneSemaphore being signaled. The signal operation for this semaphore is not scheduled at all if update() method is not called before draw(). The fix enforces to call update() before draw() during a window resize situation.

It fixes also another issue with waiting on mComputeDoneSemaphore semaphore in the wrong pipeline stage. Currently the code is waiting on mComputeDoneSemaphore in stage specified by bit VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT. This stage is not present at all in the set of graphics operations submitted to the graphics queue. The correct bit should be VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT as we are trying to sample a texture in fragment shader that ought to be filled in with data in a previous compute pass.

### DIFF
--- a/include/cinder/app/RendererVk.h
+++ b/include/cinder/app/RendererVk.h
@@ -117,7 +117,7 @@ class RendererVk : public Renderer {
 		uint32_t							mSwapchainImageCount = 2;
 		VkSampleCountFlagBits				mSamples = VK_SAMPLE_COUNT_1_BIT;
 		VkFormat							mDepthStencilFormat = VK_FORMAT_D16_UNORM;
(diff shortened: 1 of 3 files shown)


vinjn:
Looks good.

@andrewfb I can help review Vulkan PRs.
I'd like to have Vulkan renderer in master branch.

PR #1730, opened 2017-01-22
Dox markdown script

gregkepler:
Includes a script to convert a markdown file into an html file, which can then be integrated into the rest of the html documentation.

Includes:
* md2html.py script for doing markdown to html conversion
* minor css updates
* updated footer copyright to 2017
* inclusion of cherry-picked commits to fix the git guide

To use:
`python md2html.py path/to/inputfile.md path/to/outputfile.html`
ex:
`python md2html.py cmake.md htmlsrc/guides/cmake.html`


drewish:
Oh yeah it'll be nice to be able to use markdown in the docs. Much easier to focus on the content and less on the formatting.

andrewfb:
This looks good to me. I just tested it on some wiki .md and it seemed to do the right thing. A future iteration might optionally generate the full page based on an official template. I'd also reiterate that for now, we aren't integrating .md into the larger docs system - this is just designed to jumpstart HTML based on the .md of the Wiki in particular, though we can revisit that in the future. This seems like a great 1.0.

PR #1735, opened 2017-01-25
Implements fullscreen on Linux via prepareSettings.

PetrosKataras:
This allows to launch fullscreen on Linux via `prepareSettings`. Runtime fullscreen toggle remains to be implemented.

### DIFF
--- a/src/cinder/app/linux/WindowImplLinuxGlfw.cpp
+++ b/src/cinder/app/linux/WindowImplLinuxGlfw.cpp
@@ -30,6 +30,7 @@ namespace cinder { namespace app {
 WindowImplLinux::WindowImplLinux( const Window::Format &format, RendererRef sharedRenderer, AppImplLinux *appImpl )
 	: mAppImpl( appImpl )
 {
+	mFullScreen = format.isFullScreen();
 	mDisplay = format.getDisplay();
 	mRenderer = format.getRenderer();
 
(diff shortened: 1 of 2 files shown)


src/cinder/app/linux/WindowImplLinuxGlfw.cpp
```diff
@@ -74,7 +75,12 @@ WindowImplLinux::WindowImplLinux( const Window::Format &format, RendererRef shar
     ::glfwWindowHint( GLFW_SAMPLES, options.getMsaa() );
 
 	auto windowSize = format.getSize();
-	mGlfwWindow = ::glfwCreateWindow( windowSize.x, windowSize.y, format.getTitle().c_str(), NULL, NULL );
+	if( mFullScreen ) {
+		mGlfwWindow = ::glfwCreateWindow( windowSize.x, windowSize.y, format.getTitle().c_str(), ::glfwGetPrimaryMonitor(), NULL );
```
richardeakin: Is there a way to make this so it isn't always the primary monitor, but the monitor / `ci::Display` that is associated with this `Window`?

PetrosKataras:
This should do the trick although I don't currently have a multi-monitor setup at hand to test.

andrewfb:
Thanks for this @PetrosKataras I also threw in a first pass at runtime toggle-able fullscreen at a175b2af6850a930680d31f13a60b475d6cb5525

PetrosKataras:
👍

PR #1737, opened 2017-01-25
Fix shader memory leak in GlslProg

notlion:
Calling `glDeleteShader` on a shader after it is attached to a program flags it for deletion once it is no longer attached (e.g. When the program is deleted).

### DIFF
--- a/src/cinder/gl/GlslProg.cpp
+++ b/src/cinder/gl/GlslProg.cpp
@@ -482,25 +482,26 @@ GlslProg::GlslProg( const Format &format )
 {
 	mHandle = glCreateProgram();
 
-	if( ! format.getVertex().empty() )
+	GLuint vertexHandle = format.getVertex().empty() ? 0 :
 		loadShader( format.getVertex(), format.mVertexShaderPath, GL_VERTEX_SHADER, format.getPreprocessor() );
-	if( ! format.getFragment().empty() )
+	GLuint fragmentHandle = format.getFragment().empty() ? 0 :
 		loadShader( format.getFragment(), format.mFragmentShaderPath, GL_FRAGMENT_SHADER, format.getPreprocessor() );
 #if defined( CINDER_GL_HAS_GEOM_SHADER )
-	if( ! format.getGeometry().empty() )
+	GLuint geometryHandle = format.getGeometry().empty() ? 0 :
 		loadShader( format.getGeometry(), format.mGeometryShaderPath, GL_GEOMETRY_SHADER, format.getPreprocessor() );
 #endif
 #if defined( CINDER_GL_HAS_TESS_SHADER )
-	if( ! format.getTessellationCtrl().empty() )
+	GLuint tessellationCtrlHandle = format.getTessellationCtrl().empty() ? 0 :
 		loadShader( format.getTessellationCtrl(), format.mTessellationCtrlShaderPath, GL_TESS_CONTROL_SHADER, format.getPreprocessor() );
-	if( ! format.getTessellationEval().empty() )
+	GLuint tessellationEvalHandle = format.getTessellationEval().empty() ? 0 :
 		loadShader( format.getTessellationEval(), format.mTessellationEvalShaderPath, GL_TESS_EVALUATION_SHADER, format.getPreprocessor() );
 #endif
 #if defined( CINDER_GL_HAS_COMPUTE_SHADER )
-	if( ! format.getCompute().empty() )
+	GLuint computeHandle = format.getCompute().empty() ? 0 :
 		loadShader( format.getCompute(), format.mComputeShaderPath, GL_COMPUTE_SHADER, format.getPreprocessor() );
-#endif    
-    auto & userDefinedAttribs = format.getAttributes();
+#endif
+
+	auto &userDefinedAttribs = format.getAttributes();
 	
 	bool foundPositionSemantic = false;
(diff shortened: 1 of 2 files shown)


richardeakin:
This seems good and thorough to me, according to [this](https://www.opengl.org/discussion_boards/showthread.php/181432-Correct-way-to-delete-shader-programs) and [this](http://gamedev.stackexchange.com/questions/47910/after-a-succesful-gllinkprogram-should-i-delete-detach-my-shaders).

I'm inclined to think we should merge it into dev_0.9.2 branch for now though, since I haven't really read the spec on this across all gl versions and it appears to be a formality more than anything else. Unless others think This Is The Way.

notlion:
Oh interesting. So according to these posts the correct thing to do would be to detach shaders after linking the program. That would prevent loading a specific, say, fragment shader and relinking with a previously attached set of vertex, geometry, etc shaders. So we probably don't want to do that, unless it's through an explicit user-callable method. Maybe something like `GlslProg::detachShaders()` that detaches every shader returned from `glGetAttachedShaders()`.

notlion:
Sorry for the base branch noise. Took me a second to realize I had to rebase my change onto `dev_0.9.2`.

notlion:
I added a couple more commits here. The first detaches shaders immediately after linking a program, since they are no longer needed and can't be used further. This saves a small amount of memory for the lifetime of the GlslProg. My test app uses ~3MB less memory when 1000 pretty simple shaders are allocated.. only around a 0.5% change.

The second commit prevents memory leaking when GlslProg throws an exception in it's constructor. This can happen when compiling a shader, or linking. Previously, the compiling shader and program were not cleaned up. This change is useful in the case where you don't want the application to hard crash on exception, for example when loading shaders from user input.

src/cinder/gl/GlslProg.cpp
```diff
@@ -697,10 +718,20 @@ void GlslProg::loadShader( const string &shaderSource, const fs::path &shaderPat
...
+		// owned GL objects here to avoid leaking. Any other attached shaders will be cleaned up
+		// when the program is deleted
+		glDeleteShader( handle );
+		glDeleteProgram( mHandle );
+
 		throw GlslProgCompileExc( log, shaderType );
 	}
+
 	glAttachShader( mHandle, handle );
+	glDeleteShader( handle );
```
richardeakin: While I'm reading from the [docs](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glDeleteShader.xhtml) that this `glDeleteShader` is essentially flagging the shader for deletion once it is detached, I think this is confusing here (had to go look it up :) ).  Would there be any harm in moving these calls to lines 564-583 above, directly after the calls to `glDetatchShader`? This would mean you need one call to `glDeleteShader` per stage, but at least it would be clear what is going on.

If the extra calls seems ugly, I'd suggest adding a new helper function that checks if the shader handle is non-zero and if so, calls both `glDetachShader` followed by `glDeleteShader`.

src/cinder/gl/GlslProg.cpp
```diff
@@ -697,10 +718,20 @@ void GlslProg::loadShader( const string &shaderSource, const fs::path &shaderPat
 			}
 			log += "\n" + ss.str();
 		}
-#endif		
+#endif
+		// Since the GlslProg destructor will not be called after we throw, we must delete all
+		// owned GL objects here to avoid leaking. Any other attached shaders will be cleaned up
+		// when the program is deleted
+		glDeleteShader( handle );
+		glDeleteProgram( mHandle );
```
richardeakin: Nice catch, I suppose this doesn't happen in the case of an exception, as `GlslProg`'s destructor isn't called.

notlion: Yeah it's a bit of a shame, since this bails on the whole program and wouldn't allow someone catching the exception to do anything with it after that.. like say attempting to recompile a single shader. But that's not possible anyway with the current GlslProg implementation, and I don't see another way around it, unless we move compilation and linking to an init function outside the constructor.

src/cinder/gl/GlslProg.cpp
notlion: Moving this to a utility seems fine. Also, I'd kind of like to get those ternary operators out of the constructor.. How would you feel about checking the source string for `.empty()` int `loadShader()` and just returning 0 there? Seems cleaner.

src/cinder/gl/GlslProg.cpp
richardeakin: That thought had actually crossed my mind after writing the previous comment. I think it would be fine to pass in an empty string to loadShader(), and if so, it just returns 0. That method is internal, so sometimes for things like this I just write a little comment above it in the cpp for the next fellow who comes along.

src/cinder/gl/GlslProg.cpp
richardeakin: That thought had actually crossed my mind after writing the previous comment. I think it would be fine to pass in an empty string to loadShader(), and if so, it just returns 0. That method is internal, so sometimes for things like this I just write a little comment above it in the cpp for the next fellow who comes along.

notlion: Ah. Actually there's a problem with waiting to delete the shaders. If a subsequent shader fails to compile or linking the program fails they will not be deleted along with the program. Deleting them ahead of time scopes their lifetime to the program.

Issue #1740, opened 2017-02-02
OSC simple sender/receiver samples not working


daanvanhasselt:
The samples in blocks/OSC/samples/SimpleSender and blocks/OSC/samples/SimpleReceiver seem to be broken. Multithreaded variants are working though.

Windows 10, VS2015 with v140 platform toolset.

richardeakin:
What isn't working.

We don't yet ship v140 platform binaries, so unless you're building everything from scratch, expect linker errors. We'll be upgrading that very soon.

daanvanhasselt:
Building and linking is fine (of course after rebuilding Cinder), but the SimpleSender doesn't seem to be sending any OSC messages. For now I reverted back to the old OSC block which doesn't have any issues. I will investigate a bit more to see what could be causing this.

daanvanhasselt:
So the sender *is* actually sending messages (I'm using an OSC monitor to sniff for OSC messages) but the receiver is not responding to them.

daanvanhasselt:
Osc.cpp, line 1341 `mSocket->async_receive_from(...)`
Breakpoint at line 1343, inside the callback, is never hit. I'm trying to step through and debug this but I have to say I'm having a hard time understanding any of the asio code.

ryanbartley:
Are you running both from the same computer?

ryanbartley:
Actually nevermind, it looks like i somehow swapped out the `localPort` variable for an incorrect value of 3000. If you change the lines...
```
SimpleReceiverApp::SimpleReceiverApp()
: mReceiver( 3000 )
{
}
```
...to...
```
SimpleReceiverApp::SimpleReceiverApp()
: mReceiver( localPort )
{
}
```
...can you confirm that that addresses the problem? I must have been testing something other than the interaction between these two projects.

daanvanhasselt:
Yep, that fixes it. Closing the issue.

richardeakin:
Reopening until we fix the sample.

PR #1741, opened 2017-02-02
Incorrect localPort value in sample

ryanbartley:


ryanbartley:
Tried using the "connect with an issue" feature but doesn't seem to work. Anyway, this addresses #1740.

PR #1742, opened 2017-02-02
Use localPort variable

daanvanhasselt:
Instead of hardcoding the port to listen on to 3000 the variable localPort is used.


daanvanhasselt:
Duplicate of #1741

PR #1746, opened 2017-02-13
fix capture cube vs2013 project file

patrickFuerst:
The project file for visual studio had the wrong name for the source file.


andrewfb:
Thanks for this - I also cherry-picked it into `master` and added it to the _AllSamples project.

PR #1747, opened 2017-02-14
add missing Fbo::Format::setLabel implementation

patrickFuerst:
This was just throwing a linking error when used in an App.


richardeakin:
Thanks!

PR #1755, opened 2017-02-23
adding VC2015 build folder

fschliem:
Adding a VC2015 solution so that I don't need to clobber the 2013 one.

### DIFF
--- a/proj/vc2015/cinder.sln
+++ b/proj/vc2015/cinder.sln
@@ -1,0 +1,40 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Express 2013 for Windows Desktop
+VisualStudioVersion = 12.0.21005.1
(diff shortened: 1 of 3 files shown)


richardeakin:
Thanks for this, we're actually holding off VS 2015 projects until the 0.9.2 release cycle (really soon!), at which point we'll be updating all projects in the repo and other in-progress work like shared library support.

fschliem:
I figured you'd might.   I got Cinder only recently and by then VS2015 community was the default.  I believe 2017 is in beta as well.

Still new to your contribution process and whatnot so I figured I'd start with a very innocuous change.

richardeakin:
Welcome! We're also thinking about how to support VS 2017 / v141 as well (from what I've read, it should be released next month) in the next release without having to modify the cinder.vcxproj, still doing research around that.

PR #1756, opened 2017-02-23
Rbge file read

fschliem:
I was looking for HDR files, and stumbled upon this great repository:
https://hdrihaven.com/hdri.php?hdri=hansaplatz

... but when I tried loading them into Cinder apps, the Radiance file parser would fail on a header check: it was expecting "#RADIANCE" but got "#RGBE".  The rest of the file structure is exactly the same.

This pull request considers that header valid as well.


andrewfb:
Thanks for this find & fix - I cherry-picked 334809b55884e12ed73bd952b44969b234f1df9b to keep the repo history a little cleaner.

PR #1760, opened 2017-02-24
Fix radiance parser off-by-one

fschliem:
**TL;DR;** The fix is to check for EOF at the start of the line scanning method instead of the end so we don't skip the last line.

**Details:**

The core of the RADIANCE/HDR/RBGE file decoder does this:

```c++
	for( int y = height - 1; y >= 0; y-- ) {
		if( ! decrunchScanline( scanline.get(), width, stream.get() ) )
			break;  /// <<< if we hit this, we miss a line!
		workOnRgbeScanline( scanline.get(), width, cols );
		cols += width * 3;
	}
```

... where ```decrunchScanline(...)``` deals with unpacking the RLE bytes that are compressed into a non-compressed format that ```workOnRgbeScanline``` then converts into a 3-float value scanline.

```c++
bool decrunchScanline( RgbePixel *scanline, int len, IStreamCinder *stream )
{
      ...
      // while we have bytes to read from the stream ...
      while( ... ) {
            ...
            // ... read bytes form the stream into the scanline
            stream->read(&scanline[j++][i]);
      }

      // and finally this: 
      return ! stream->isEof();
     // ... that is a problem on the last line !!!
}
```
The problem with the check at the end means that for the last line of a perfectly valid file, we read in the last line, parse all the data, but as we consume the input stream to the last byte, we reach EOF, report it at the end and so we skip the part where we call ```workOnRgbeScanline(...)``` in the top loop.
The texture now has one line worth of unprocessed memory.

Apologies for the multiple commits, trying to wrangle VS2015 into mangling the spacing and style as little as possible.


richardeakin:
Thanks for these! A couple requests to ease reviewing and getting your contributions merged.

- Please follow the [Contributing.md's styleguide](https://github.com/cinder/Cinder/blob/master/CONTRIBUTING.md#style-guide), I see your if statements are formatted differently.
- In order to keep the history clean, I'd suggest creating a new branch off of the current master, and then cherry pick the specific commits you want into that. This way, you won't have the 'adding VC2015', 'removing VC2015' commits, but you can continue to work from your own dev branch with those in it if you choose (this is usually how I get changes I make back into cinder core).

Once you've done those, you can update this PR by doing a `git push --force` to your fix_radiance_off_by_one branch, but be aware that will blow away the current history of that branch.

src/cinder/ImageSourceFileRadiance.cpp
```diff
@@ -156,6 +156,10 @@ bool decrunchScanline( RgbePixel *scanline, int len, IStreamCinder *stream )
 {
 	char i;
 
+	// Early out if there's nothing more to read from the stream
+	if (stream->isEof())
```
richardeakin: whitespace should be:

```
if( stream->isEof() )
    return false;
```

fschliem:
Closing this, opened a cleaner version here: https://github.com/cinder/Cinder/pull/1761

I reset my fork back to origin and re-edited the change.  It turns out that VS2015 wants to reformat things on paste....  sorry about that.

PR #1764, opened 2017-02-26
Conversion error

MikeGitb:
I was going over a few conversion warnings in VS and afaik came about an actual error:
If 

    Target::copyIndexDataForceTriangles

is called with a `uint16_t` array as a target and an non-zero offset, bytes are memcopied from a `uint32_t` array to a uint16_t array. I'm PR ing the whole branch, which silences a few more warnings, but the fix for that error is contained in the last commit. So if you don't want to merge the whole PR just cherry pick that one.

### DIFF
--- a/include/cinder/GeomIo.h
+++ b/include/cinder/GeomIo.h
@@ -844,7 +844,7 @@ class BSpline : public Source {
 	template<typename T>
 	void init( const ci::BSpline<4,T> &spline, int subdivisions );
 
-	int						mPositionDims;
+	uint8_t					mPositionDims;
 	size_t					mNumVertices;
 	std::vector<float>		mPositions;
 	std::vector<vec3>		mNormals;
@@ -1306,7 +1306,7 @@ class Constant : public Modifier {
   protected:
 	geom::Attrib	mAttrib;
 	vec4			mValue;
-	int				mDims;
+	uint8_t				mDims;
 };
 
 //! Maps an attribute as a function of another attribute. Valid types are: float, vec2, vec3, vec4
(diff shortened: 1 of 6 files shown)


andrewfb:
All of these warnings fixes look good to me - thanks again for your help in addressing those.

With regards to the narrowing issues, I think there's three things going on here.
* The `memcpy()` is definitely the wrong thing to do when `T != uint32_t` - thanks for this catch.
* The `indexOffset` is actually meant to be a constant added to the indices themselves. It's used when concatenating two meshes, for example, where we want to "slide" all of the values of the second mesh in order to account for the vertices of the first. As such, we'll need a different solution than the proposed `std::copy_n()` when `indexOffset != 0`
* I made a conscious choice to allow the risk of narrowing (which would generally only come up when using 16-bit indexed geometry, for example on some iOS devices) for the sake of performance. That does create the possibility of a potentially hard-to-find bug, and I think your `assert()` approach could be a nice balance. The way I would propose to solve it would be something like:
```
template<typename S>
bool indicesInRange( const uint32_t *indices, size_t numIndices )
{
    return true;
}
template<>
bool indicesInRange<uint16_t>( const uint32_t *indices, size_t numIndices )
{
    for( size_t i = 0; i < numIndices; ++i )
        if( indices[i] > 65535 )
            return false;
    return true;
}

template<typename T>
void copyIndexDataForceTrianglesImpl( Primitive primitive, const uint32_t *source, size_t numIndices, T indexOffset, T *target )
{
    ...
    CI_ASSERT( indicesInRange<T>( source, numIndices ) );
}
```

I will submit a PR to that effect shortly.

MikeGitb:
Luckily you are paying attention - sorry for that buggy PR. Shall I close this for now and offer the other changes as part of more targeted but bigger PRs? Or shall I just remove the commits that are superseded by yours?

andrewfb:
No problem at all - thanks for these warnings fixes and also identifying the problem. I manually merged the former, and then merged my recent PR for the latter. Thanks again.

PR #1765, opened 2017-02-26
Fix some warnings

MikeGitb:
Again working on some warnings that VS puts out under W4
- Some unused local variables
- Calling the constructor of a virtual base class from a abstract class doesn't have any effect
- use `std::vector` to implement internal buffer of IStreamUrlImplWinInet (makes functions exception safe)
- Fix a bug in `getClosestPointQuadratic`, where number of solutions wasn't considered.


andrewfb:
Thanks for these - merged.

PR #1766, opened 2017-02-27
copyIndexDataForceTriangles bugs

andrewfb:
This addresses a bug found by @MikeGitb in #1764 in `Target::copyIndexDataForceTriangles()` which would manifest when using GL_TRIANGLES primitive with 16-bit indices. Additionally, this PR asserts that indices are within range in the 16-bit index case. Unfortunately I don't see an easy way to unit test that this assert fires using Catch. The code I used to exercise it is below.

```
	vector<uint16_t> idx16( 10 );
	vector<uint32_t> idx32;
	
	for( size_t i = 0; i < idx16.size() - 1; ++i )
		idx32.push_back( (uint32_t)i );
	idx32.push_back( 65536 ); // failure case
	
	geom::Target::copyIndexDataForceTriangles( geom::Primitive::TRIANGLES, idx32.data(), idx32.size(), 0, idx16.data() );
```

### DIFF
--- a/src/cinder/GeomIo.cpp
+++ b/src/cinder/GeomIo.cpp
@@ -322,8 +322,27 @@ void copyData( uint8_t srcDimensions, const float *srcData, size_t numElements, 
 
 namespace { 
 template<typename T>
+bool indicesInRange( const uint32_t *indices, size_t numIndices )
+{
+    for( size_t i = 0; i < numIndices; ++i )
+        if( indices[i] > std::numeric_limits<T>::max() )
+            return false;
+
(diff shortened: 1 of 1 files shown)


MikeGitb:
Thanks for fixing this. 
You might want to consider keeping the memcpy optimization where source and target are of the same type and the offset is zero. Not quite sure if all compilers can do that transformation by themselves. 
One way to acheive this without additional checks would be to use `std::copy(_n)`, in the offset == 0 branch which usually uses memmove/cpy automatically when possible.

PR #1770, opened 2017-03-02
Added VS2015 targets to build Cinder as a .dll/.lib

simongeilfus:


richardeakin:
Also worth noting, that this upgrades and moves the vc2013 / v120 cinder vcxproj and sln files to vc2015 / v140. Right after we merge this, we'll be upgrading all sample and test projects to use vc2015 too.

Looks like glload on linux takes a slightly different include path, causing the [Ubuntu build](https://travis-ci.org/cinder/Cinder/jobs/206852977) to fail. This is the error I'm seeing on travis:

```
In file included from /home/travis/build/cinder/Cinder/include/glload/glx_all.h:5:
/home/travis/build/cinder/Cinder/include/glload/_int_glx_exts.h:83:1: fatal error: 
      unknown type name 'CI_API'
CI_API extern PFNGLXCREATECONTEXTATTRIBSARBPROC _funcptr_glXCreateContex...
```

Probably need to add an extra `#include "cinder/Export.h" to [glload/gl_core.h](https://github.com/cinder/Cinder/blob/master/include/glload/gl_core.h).

MikeGitb:
Does that mean cinder officially drops vs2013 support?

richardeakin:
@MikeGitb I believe that for 0.9.2, we'll be officially moving to VS 2015 / v140 as our minimum target on windows desktop. 0.9.1 will be released soon that will require VS 2013 / v120.

andrewfb:
Can confirm that OpenGL CubeMapping sample still works under Ubuntu 16.10 with this branch.

PR #1773, opened 2017-03-02
Fix some more warnings

MikeGitb:
Another set of changes to address warnings from VS.

There are a few whitespace changes inbetween that I can remove and I can squash the whole thing into fewer commits if you prefer. 

No actual bugs this time ;)

### DIFF
--- a/include/cinder/ChanTraits.h
+++ b/include/cinder/ChanTraits.h
@@ -38,14 +38,14 @@ struct CI_API CHANTRAIT<uint8_t>
 	typedef uint32_t Sum;
 	typedef uint32_t Accum;
 	typedef int32_t SignedSum;
-	static uint8_t max() { return 255; }
-	static uint8_t convert( uint8_t v ) { return v; }
-	static uint8_t convert( uint16_t v ) { return v / 257; }
-	static uint8_t convert( half_float v ) { return static_cast<uint8_t>( glm::clamp( halfToFloat( v ), 0.0f, 1.0f ) * 255 ); }
-	static uint8_t convert( float v ) { return static_cast<uint8_t>( glm::clamp( v, 0.0f, 1.0f ) * 255 ); }
-	static uint8_t grayscale( uint8_t r, uint8_t g, uint8_t b ) { return ( r * 54 + g * 183 + b * 19 ) >> 8; } // luma coefficients from Rec. 709
-	static uint8_t premultiply( uint8_t c, uint8_t a ) { return a * c / 255; }
-	//static uint8_t premultiply( uint8_t c, uint8_t a ) { uint16_t t = c * a + 0x80f; return ( ( t >> 8 ) + t ) >> 8; }
+	static uint8_t max()										{ return 255; }
+	static uint8_t convert( uint8_t v )							{ return v; }
+	static uint8_t convert( uint16_t v )						{ return static_cast<uint8_t>( v / 257 ); }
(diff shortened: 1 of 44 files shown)


MikeGitb:
Regarding the second commit: Is it a safe assumption, that distance squared will never exceed the range of a 32 bit integer under "normal" conditions? Otehrwise this introduces an error.

MikeGitb:
One thing that came to my mind during the last round of changes: There are a lot of places in the library where a narrowing cast is performed. Would it make sense to add a gernal `narrow` and/or `narrow_cast` function simliar to [those in the gsl](https://github.com/Microsoft/GSL/blob/master/include/gsl/gsl_util#L90-L127) that might even do some range checks in debug mode?

richardeakin:
Only skimmed at the changes so far, which seem mostly whitespace, but can we switch this to be against the [dev_0.9.2](https://github.com/cinder/cinder/tree/dev_0.9.2) branch? Trying to freeze master branch for all things except bug fixes and docs at this point. There are also some very large over-arching changes coming in with #1770, not sure if that will cause conflicts but should think about getting all that synced together.

MikeGitb:
Sure. 
However, if 0.9.1 get's released soon anyway, I can also just wait until #1770 gets merged into master and then rebase - should be rather simple. 

Regarding the whitespaces: My IDE deletes trailing whitespaces on save and a lot of files have tabs on the empty lines between two indentated blocks. However, I put those changes into separate commits, so I can easily remove them to reduce the noise.

richardeakin:
I think that'd work too, we'll look to get this in either after #1770 is merged into dev or dev is merged into master (post 0.9.1 release).

The whitespace fixes are OK, you're keeping them in separate commits which is great. I think there is a way we can view the changeset without whitespace mods? Can't recall how to do that at the moment, but that way we wouldn't need to look commit by commit to review things that could potentially change functionality.

Thanks again for helping to tidy up things! It'll be great once we have a relatively quiet build with minimal to no warnings.

MikeGitb:
Ironically, I just pulled the whitespace commits, because I didn't like signal to noise ratio ;) 
If you are interested in those I can also open a different PR. But it might be more efficient, if you'd just run a script that does this accross the whole project yourself. Looking at a PR without the whitespace changes is possible by appending `?w=1` to the web address.

richardeakin:
Cool, thanks will try that next time. Yes I believe we'll do an entire sweep at some point, since we began establishing a styleguide a couple years after the project was open sourced.

MikeGitb:
Ok I rebased the PR against the dev_0.9.2  branch. 
One additional remark: when building for x64 the compiler complains

   `cl : Command line warning D9002 : ignoring unknown option '/arch:SSE2'`

The reason being that sse2 is activated by default on x64 (as sse2 is part of the x64 instruction set) If you are giving the project settings an overhaul anyway you might want to get rid of that option or maybe even switch to `/arch:AVX2`

richardeakin:
Awesome, hoping to look at getting this in soon.

re. `ignoring unknown option '/arch:SSE2'`, yes we need to get that sorted out, I think it is quite vestigial now but perhaps still useful on the 32-bit target? I haven't dug into this enough to myself to make any changes but I agree, now is a good time since we're revamping all VS project files. Perhaps we should move this to a new issue along with information about our options, such as enabling AVX2 and where that is and isn't possible.

andrewfb:
Thanks again for going after all these. Cherry-picked into master.

Issue #1776, opened 2017-03-08
Cmake asset folders on windows 


m1keall1son:
symlink creation fails on windows, so without explicitly copying the files via CINDER_COPY_ASSETS, assets wont be found.

symlinks should be created on windows as: 

`cmd.exe /c mklink /D ${link} ${target}`

but it requires the building terminal be run as Administrator since symlink creation is protected by defuault.  

alternatively it could use `runas.exe`

`runas.exe /user:administrator /savecred "cmd.exe /c mklink /D ${link} ${target}"`

which will pause the project generation and prompt the user for a password, but I've not been able to get this to work even with the correct password...sigh...

cmake should put out a warning when the symlink fails on windows and if the error is because of privileges, then put out another warning to re-run the cmake command as Administrator

richardeakin:
Fixed with #1780.

PR #1777, opened 2017-03-08
updated ci_make_app to symlink assets on windows

m1keall1son:
addresses #1776 

To test this, generate a project with the cube sample before and after this addition.  

NOTE: must run from a terminal run As Administrator
If you don't, it will emit a warning that you should run as administrator and try again.

```
$ mkdir build
$ cd build
$ $ cmake -DCINDER_BUILD_SAMPLE=_opengl/Cube -G"Visual Studio 14 2015 Win64" ..
```


andrewfb:
One nice solution to the Administrator issue might be to teach Cinder to follow Windows shortcuts when searching for assets directories. Then we could use `mklink` to simply create a shortcut, which I believe does not require privilege escalation, and that seems like a nice feature in its own right.

m1keall1son:
After a little more digging into the documentation, i found if i use the junction flag `/J` instead of the soft directory symlink flag `/D` then mklink will create a junction, which i guess is a symlink but for local files only--no possibility of a directory on the network somewhere,  and doesn't require permissions.  

To your point though, i think there would be value to being able to generate and follow shortcuts (*.lnk) from within cinder msw, there's probably a use-case for it.

m1keall1son:
closing this PR because I included #1779 in this PR by accident

Issue #1778, opened 2017-03-08
cmake ignores LIBCRT and LIBCPRT in release 


m1keall1son:
Comparing the provided sample projects to a cmake generated project in visual studio 2015, i see that the LIBCRT.lib and LIBCPRT.lib are specifically ignored in the generated version, causing the Release build to fail to link to the runtime library and all hell to break loose.

```cmake
#platform_msw.cmake
# Static library flags
    set( CINDER_STATIC_LIBS_FLAGS_DEBUG     "/NODEFAULTLIB:LIBCMT /NODEFAULTLIB:LIBCPMT" )
    set( CINDER_STATIC_LIBS_FLAGS_RELEASE   "/NODEFAULTLIB:LIBCMT /NODEFAULTLIB:LIBCPMT" )
```

removing these ignores from the release properties restore balance to the universe.

m1keall1son:
Also it appears here:

```cmake
#cinderMakeApp.cmake
# Ignore Specific Default Libraries
	if( MSVC )
		set_target_properties( ${ARG_APP_NAME} PROPERTIES LINK_FLAGS "/NODEFAULTLIB:LIBCMT /NODEFAULTLIB:LIBCPMT" )
	endif()
```

richardeakin:
Fixed in dev_0.9.2 branch, see comment in #1779 for details.

PR #1779, opened 2017-03-08
Set cmake windows linker flags to fix release build

m1keall1son:
addresses #1778 

Flags for specifically ignoring LIBCRT and LIBCPRT were being added to the release configurations breaking the builds on windows. 

Now they are only ignored in debug, which is how all other cinder projects are setup.


richardeakin:
Thanks Mike, manually cherry-picked this into the dev_0.9.2 branch in 25758fa270e8fad429beebbb133e8a6a016b6ee2.

Anyone who's interested, please see #1801 for remaining cmake + VS issues.

PR #1791, opened 2017-03-19
Appveyor CI working.  Broken into two groups, one for winRT with no u…

lucasvickers:
## AppVeyor CI support for Cinder

This PR provides CI support for Windows VS 2015 via AppVeyor.
I broke it down into **Cinder** and **Cinder-WinRT** projects.  Additionally the unit tests were fixed for the ANGLE targets.

### Cinder
Cinder will build cinder, then build our unit tests, and finally run the unit tests.  This project covers the following configurations and platforms:

**Configurations:**
- Debug
- Release
- Debug_ANGLE
- Release_ANGLE

**Platforms:**
- Win32
- x64

### Cinder-WinRT
Cinder-WinRT will build cinder RT, but it will not run unit tests.  This project covers the following configurations and platforms:

**Configurations:**
- Debug
- Release

**Platforms:**
- x86
- x64
- ARM

### Instructions
To integrate this into the Cinder repo, set up AppVeyor, then in the config do the following:
- Set the _Custom configuration .yml file name_ to be:
  - `.appveyor_winrt.yml` or `.appveyor.yml` (depending on project)
- Select the _Always build closed Pull Requests_ checkbox (optional).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,34 @@
+#version: 1.0.{build}
+
+build:
+  parallel: true
+  project: proj/vc2015/cinder.sln
+
+platform:
+    - x64
+    - Win32
+
+configuration:
+  - Debug
+  - Release
+  - Debug_ANGLE
+  - Release_ANGLE
+
+image: 
+  - Visual Studio 2015
+
+install:
+  - git submodule update --init --recursive
+
+clone_depth: 10
+
+before_test:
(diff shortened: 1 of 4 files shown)


MikeGitb:
Great!

richardeakin:
As far as Appveyor integration goes, I think we also want to select "Skip branches without appveyor.yml", since there will be a release branch in there that won't have the appveyor.yml file for quite some time.

Do we also want to set the git clone depth, to speed up the fetch?

lucasvickers:
clone depth is set to 10 in the file, which is arbitrary.  I guess we can set it to 1?

PR #1793, opened 2017-03-23
Add initial FileWatcher class

richardeakin:
This introduces a new class `ci::FileWatcher`, to facilitate the task of watching the filesystem for modifications, typically used in asset managers and live coding applications, among other applications. It is meant to be a fundamental layer for these purposes, in that it is the guy who speaks directly to the filesystem.

The focus of this initial work is on the public facing interface, implemented in a simple manner with a secondary thread that polls for file changes. The API is callback (`std::function`) based, which is using `ci::signals` under the hood so that watches can easily be scoped to avoid stale pointers in heavily dynamic applications. The following example shows typical usage:

```cpp
mConnection = ci::FileWatcher::instance().watch( "someFile.txt", [this]( const ci::WatchEvent &event ) {
	// load / reload file...
} );
```

Two things to note: First, when `mConnection` goes out of scope (typically when `this` is destroyed), the callback will be automatically disconnected, so it is generally safe to use `this` and its member data within the callback. Second, the callback will, by default, fire initially on watch (before the call to `watch()` returns). This is important in ensuring the order in which data loads in typical setup situations. If you don't want the callback to fire on initial load, for example if you are watching include paths on a file that has already been watched, you can pass in the `callOnWatch( false )` option as such:

```cpp
mConnection = ci::FileWatcher::instance().watch( "someFile.txt", FileWatcher::Options().callOnWatch( false ), [this]( const ci::WatchEvent &event ) {
	// this will only fire when the file is next modified
} );
```

By default `FileWatcher` will register for updates with the App's update signal (if the App exists), however in an appless situation or in the case that you want / need to manage the update loop yourself, you can disable that with `setConnectToAppUpdateEnabled( false )` and then calling the public `update()` method.

~~The above examples use the 'default shared instance' of FileWatcher for convenience, however you can also create and manage your own with the `FileWatcher::create()` method.~~ edit: we've decided to do away with the create() method, instead you can construct but not copy a `FileWatcher`, and the `instance()` method returns a `FileWatcher&`.

Some initial unit tests have been added, and usage has been added to the _opengl/CubeMapping sample to show how you can use the FileWatcher for live shader editing.

Still open to debate various aspects of the API (better methods names included), though there will be more rounds for additional features. Some of these planned features that we've discussed for later include:

- Using platform-specific APIs to get notifications when files change, rather than polling the filesystem.
- Watching folders.
- Information about the type of modification event (ex. add, delete, modify).


MikeGitb:
Thanks for that handy feature. 

One thing I was wondering when looking at the Implementation: It seems there are a lot of accesses to `mWatchList`, that are not protected by the mutex. Are you sure this is safe?

richardeakin:
>  It seems there are a lot of accesses to mWatchList, that are not protected by the mutex. Are you sure this is safe?

Not at all, the (newer) `watch()` and `enable()` methods look like they all need lock guards. I'll give that a look during the next pass, open to all feedback otherwise before then.

include/cinder/FileWatcher.h
```diff
@@ -0,0 +1,157 @@
...
+
+	void	configureWatchPolling();
+	void	stopWatchPolling();
+	void	threadEntry();
+
+	std::list<std::unique_ptr<Watch>>	mWatchList;
+	std::recursive_mutex				mMutex;
+	std::unique_ptr<std::thread>		mThread;
+	std::atomic<bool>					mThreadShouldQuit;
+	std::atomic<double>					mThreadUpdateInterval		= { 0.02 };
```
richardeakin: Currently we use doubles to represent time throughout the codebase. This is also a private variable so I think double is fine here, and more precision than it'd ever need.

include/cinder/FileWatcher.h
```diff
@@ -0,0 +1,157 @@
...
+
+  private:
+	FileWatcher();
+
+	void	configureWatchPolling();
+	void	stopWatchPolling();
+	void	threadEntry();
+
+	std::list<std::unique_ptr<Watch>>	mWatchList;
+	std::recursive_mutex				mMutex;
```
richardeakin: both foreground and background threads can be adding watches at the same time (think async loaders with dependencies).

src/cinder/FileWatcher.cpp
```diff
@@ -0,0 +1,412 @@
...
+			break;
+
+		watch->emitCallback();
+	}
+}
+
+const size_t FileWatcher::getNumWatchedFiles() const
+{
+	size_t result = 0;
+	for( const auto &w : mWatchList ) {
```
richardeakin: Probably

src/cinder/FileWatcher.cpp
```diff
@@ -0,0 +1,412 @@
...
+	auto fullPath = findFullFilePath( filePath );
+
+	for( auto &watch : mWatchList ) {
+		watch->setEnabled( false, fullPath );
+	}
+}
+
+void FileWatcher::configureWatchPolling()
+{
+	if( mConnectToAppUpdateEnabled && ! mConnectionAppUpdate.isConnected() && app::App::get() )
```
richardeakin: There are app less situations, such as in the unit tests, or if you made a standalone program that did some sort of file post processing or something.

src/cinder/FileWatcher.cpp
```diff
@@ -0,0 +1,412 @@
...
+void FileWatcher::setWatchingEnabled( bool enable )
+{
+	if( mWatchingEnabled == enable )
+		return;
+
+	mWatchingEnabled = enable;
+	if( enable )
+		instance()->configureWatchPolling();
+	else
+		instance()->stopWatchPolling();
```
richardeakin: Good catch, this was from when that function was static, in a previous version.

richardeakin:
(...and back from vacation...)

That for the thorough feedback! Some quick comments on our current design decisions:

static instance: this can be very handy in projects where you just need to make things work, but you don't want to be passing around a FileWatcher instance through many layers (think 3 weeks, 5 people, 12 hour days and angry client).  So it is for convenience. However we do want to be able to provide the more idealistically correct way of using your own FileWatcher instance, which you can use in your own larger asset management system, if desired.

Reason that `std::list` is used is the way I've implemented the modified list, basically all Watches that have modified files that need callbacks are moved to the front on the background thread, and processed from in main update() loop. Once we hit a Watch that doesn't need a callback, we break.

As far as using asio::io_service, we'll most likely move there eventually, however for this initial implementation we discussed getting it in there with using a simple `std::thread` polling mechanism.  We're spending 10mb of memory for simplicity, and well polling the filesystem is never going to be that fast. :) There will be later work that uses system level callbacks that will already use `asio`, so probably the switch will happen then.  From what I know about `std::async`, on some platforms (ex. Windows) it will spin up a new thread each time you call it, so even worse than just creating one for each FileWatcher.

Thanks for looking at all the mutex places, those all need to be looked at.

richardeakin:
One last comment on @ryanbartley's feedback: I'm not against making the constructor public but non-copyable. We'll almost surely be better off making it non-copyable (or have to) as it is going to gain some platform-specific implementations that are hidden behind a `uniqe_ptr`. @lucasvickers any opinions here?

src/cinder/FileWatcher.cpp
```diff
@@ -0,0 +1,412 @@
...
+	mWatchItems.reserve( filePaths.size() );
+	for( const auto &fp : filePaths ) {
+		auto fullPath = findFullFilePath( fp );
+		mWatchItems.push_back( { fullPath, fs::last_write_time( fullPath ), true } );
+	}
+
+	if( needsCallback ) {
+		// mark all files as modified, using the full path we just resolved.
+		for( const auto &item : mWatchItems )
+			mModifiedFilePaths.push_back( item.mFilePath );
```
MikeGitb: Would that gain anything?

paulhoux:
I second @ryanbartley 's comments (including the "thank you for adding this" one). I believe the `std::list` isn't required, even if you like to move items to the front a lot: using a `std::vector`, you can simply swap them or use `std::remove`, while benefiting from cache coherency, less memory fragmentation and maybe a simpler implementation altogether. None of this is a deal breaker, though, just a matter of preference.

The more important thing I'd like to mention, is that it is pretty wasteful to have either a single thread checking a single file, or a single thread regularly checking a bunch of files. It's accessing the file system quite often. On Windows, there is a platform specific method to efficiently check all files in a specific folder by letting the file system notify you when one or more files change, along with a description of the exact changes. Therefor, I'd opt for a `FolderWatcher`, instead of a `FileWatcher` and use the native code where applicable. See: [ReadDirectoryChangesW](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365465(v=vs.85).aspx) and [GetOverlappedResult](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683209(v=vs.85).aspx).

simongeilfus:
Hey @paulhoux . I think a small roadmap would help understand what's going on here. The idea is to roll a first version using polling and file system checks so we get that in sooner and start using it. That said @lucasvickers is already working on an improved version that uses native callbacks and so what you are suggesting is definitely the direction this code is suppose to take.

So I'd say let's get this in there so we can start refining its API by using this instead of our own versions. And let's work on making the implementation better iteratively.

MikeGitb:
One thing to be aware of when using std::vector instead of std::list is reentrancy: One can't use iterators to loop over a vector, if one of the callbacks called in the loop body might modify the list - that includes range based for loops. Instead one would have to use a classic index based for loop.

Regarding the moving shuffling of the entries - on a vector this can be very efficiently done via std::partition.

richardeakin:
I'm iterating over the list while moving around its elements - something vectors suck at. A different approach would be to use a second 'needs update' vector that you push to from the background thread and drain on the main thread. But the performance gain is negligible, when either compared to the file timestamp checks on the background thread or that only the first couple elements are ever looked at from the main thread.

As Simon wrote, this is 'initial', so that we can get a baseline file watching tool into core that we can all use and make better. For this, I tried to use straightforward solutions that required the least amount of code.

lucasvickers:
Thanks for all the comments guys.

@paulhoux once we get this initial interface and basic functionality in, I'll be refactoring it behind the scenes to look a lot closer to [Cinder-FileWatcher](https://github.com/lucasvickers/Cinder-FileWatcher).  The idea will be to move over to asio and take in an io_service, then utilize os level APIs.  When it comes time to do it on Windows I'll definitely be talking to you.  I haven't used [GetOverlappedResult](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683209(v=vs.85).aspx) and will be curious how that works.

richardeakin:
Thanks all for partaking in this one, it's a bit of a tricky one as we're trying to think forward while getting the bulk of the functionality in there now. @lucasvickers has made some updates that are merged into this PR and I made a couple tidy-up commits to follow, we believe this is ready to merge or near to it.

The largest change we discussed and decided on is to expose the constructor and make it non-copyable. Furthermore the shared `instance()` method returns a reference, which indicates that there will always be a shared instance (whereas a pointer can be null).

The watch container is left for the moment as a `list<unique_ptr<Watch>>` for simplicity, considering that a vector of pointers doesn't give you any increased performance. I'm sure this will be thought about more in the future when there are multiple backends for how the watching is done, and this is internal so that shouldn't be a problem to revisit.

PR #1802, opened 2017-04-02
VboMesh::allocateIndexVbo() incorrectly handles GL_UNSIGNED_BYTE

IntellectualKitty:
In the original code, geom::ExcIllegalIndexType() is thrown if mIndexType is GL_UNSIGNED_BYTE.


IntellectualKitty:
Note that the else clause throwing `geom::ExcIllegalIndexType()` does not appear in the small snippet shown in the change viewer.  You have to view the whole file to see the context -- otherwise this code change will look like it doesn't do anything.

PR #1805, opened 2017-04-03
Replace 2.0f * 3.14159f with float( M_PI * 2 ) in GeomIo.cpp

IntellectualKitty:
Just a minor mathematical fix.  I noticed that float( M_PI * 2 ) was used more consistently.

### DIFF
--- a/src/cinder/GeomIo.cpp
+++ b/src/cinder/GeomIo.cpp
@@ -1738,7 +1738,7 @@ void Circle::loadInto( Target *target, const AttribSet &/*requestedAttribs*/ ) c
 	normals.emplace_back( 0, 0, 1 );
 
 	// iterate the segments
-	const float tDelta = 1 / (float)mNumSubdivisions * 2.0f * 3.14159f;
+	const float tDelta = 1 / (float)mNumSubdivisions * float( M_PI * 2 );
 	float t = 0;
 	for( int s = 0; s <= mNumSubdivisions; s++ ) {
 		vec2 unit( math<float>::cos( t ), math<float>::sin( t ) );
@@ -1827,7 +1827,7 @@ void Ring::loadInto( Target *target, const AttribSet &/*requestedAttribs*/ ) con
 	float outerRadius = mRadius + 0.5f * mWidth;
 
 	// iterate the segments
-	const float tDelta = 1 / (float) mNumSubdivisions * 2.0f * 3.14159f;
+	const float tDelta = 1 / (float) mNumSubdivisions * float( M_PI * 2 );
 	float t = 0;
 	for( int s = 0; s <= mNumSubdivisions; s++ ) {
 		vec2 unit( math<float>::cos( t ), math<float>::sin( t ) );


MikeGitb:
I'm tempted to suggest the introduction of global `2pi` and `2pif`constants.

That aside I personally would prefer the usage of static_casts, but I don't know, what the maintainers' view on this is.

IntellectualKitty:
@MikeGitb I, too, generally prefer C++ style static_casts, and I abhor old C-style casts in C++ code.  However, I don't mind things like `float( M_PI * 2 )` because that is legitimate C++ since `float` is being used as a constructor in this case.

I, too, also think it's easier to code with 2-PI constants.  And I never understood why that wasn't made part of the C standard.

richardeakin:
This ancient forum post is one take on why we didn't add any sort of `M_TWO_PI` constants, and decided to stick with just the common `M_PI`: https://forum.libcinder.org/topic/rfc-math-constants

As far as `static_cast()` goes, I think it usually depends on what you're doing, though IMO for simple math types like double to float, there's no added clarity or safety benefit between using the C style cast or something like `float( M_PI * 2 )`, so I prefer to go with the most legible version. Dealing with polymorphic types is a different story.

notlion:
FWIW, GLM comes with `glm::pi<float>()` and `glm::two_pi<float>()`, etc in [constants.hpp](https://github.com/cinder/Cinder/blob/master/include/glm/gtc/constants.hpp).

MikeGitb:
Nice, so all in all one could write `one_over_two_pi<float>() / mNumSubdivisions` - without any casts.

IntellectualKitty:
@richardeakin thanks for the link detailing the issues with `M_TWO_PI` constants.  That really clears it up.  That's one of my big beefs with programming history -- using abbreviations and contractions instead of programming for clarity.  If the author had simply used `M_2_OVER_PI` instead of `M_2_PI` there wouldn't be any confusion.

I also understand @andrewfb 's concerns about "[not wanting] to define a competing definition."  The other thing that bothers me about legacy programming standards is that they never go away -- our industry often just keeps building on top of them instead of refining and correcting them.  For that reason, I think that whenever someone defines a standard it should also include an expiration date, so that the next revision can remove the mistakes and improve on what was incomplete or what wasn't well thought through.  I, of course, support keeping legacy standards around for maintaining legacy code, but I certainly wish that modern code could be created and maintained with modern, corrected standards.

As for static_casts, I agree with your point about legibility.  I find that they are a nuisance to write and to read.  I am pedantic about warnings -- pun intended :) -- and when I code, I turn on every warning switch and flag that I can find so that I can keep my code as correct and up-to-date as possible.  So, the first time I got dinged by the compiler about using "legacy C-style casts" in my C++ code, I forced myself to start using C++ casts exclusively.

IntellectualKitty:
@notlion thanks for posting the note about GLM's constants.  I'm with @MikeGitb .  I think they're pretty sweet!

@richardeakin and @andrewfb while we're on the topic, what is Cinder's preferred standard with regard to math functions?  For instance, when I've scanned the repository, I've see `math<float>::cos()`, `glm::cos()`, and `cos()` all used fairly interchangeably in Cinder's source code, tests, and samples (I of course ignore the usage in included libraries).

MikeGitb:
This might be of interest to you:
http://discourse.libcinder.org/t/purpose-of-cinder-math/
I actually intended to give the math section a little overhaul and replace its internal usage with std::.. functions where possible, once the latest "silence warnings" PR is merged.

The reason I use static casts is that they are so ugly I sooner or later try to refactor the code in a way that I don't need them anymore. Although I have to admit that that I break my own rule too often.

There is by the way a recent paper that would finally bring all those math constants into the cpp standard, but offhand I don't remember the naming scheme - but as they won't use the posix macros there is a chance it will be something sane. But that is not really relevant right now.

MikeGitb:
@IntellectualKitty: Any particular reason you closed this pr?

PR #1809, opened 2017-04-06
allow blocks to be named without Cinder- prefix in cmake

whg:
Community blocks usually have the 'Cinder-' prefix on their repository names, but the block name itself usually omits this. For blocks that are placed in the blocks directory, this patch allows for the blocks to be named without the prefix for the `BLOCKS` variable in `ci_make_app`.


richardeakin:
Sorry, I'm not sure I understand the purpose of this - is it so that you could for example specify something like `OpenCV` as the blocks name instead of `Cinder-OpenCV`? I could be misinterpreting this, but the latter actually seems clearer to me (since OpenCV is something separate to the cinderblock).

whg:
Yes, you've got it. To me it makes sense that what is passed to `BLOCKS` should be the same as the `name` attribute in the cinderblock.xml. This would make the CMake usage more similar to Tinderbox.
I can see what you're saying about clarity with the OpenCV example, but I would argue that `OpenCV` *is* the block's name!

richardeakin:
A couple of us discussed this a bit last night, and we're just not sure that the added convenience justifies adding extra logic. Currently a cinderblock's name maps to its foldername exactly, and while we don't have any hard policies about this yet, at least it leaves a bit of a crumb trail for tracking down where things come from. At the same time, we're still sussing out how cmake fits into the cinder ecosystem (which still needs some work concerning its relationship with tinderbox), so while I think adding more features makes sense, special casing things like optional prefixes are a harder sell.

On a side note, perhaps we should clarify what a cinderblock's name actually is. I was a bit unclear [here](https://libcinder.org/docs/guides/cmake/cmake.html#using-cinderblocks), perhaps I should have written 'the folder the block lives in'.

Though contributions in the build system are very much appreciated, please do let us know if you've got other ideas on how to improve these things.

PR #1816, opened 2017-04-11
change '>>' to '> >'

mofeing:
Building a cinder app gives the following error "a space is required between consecutive right angle brackets (use '> >')"


andrewfb:
Hey - what compiler are you seeing this error with?

mofeing:
@andrewfb clang++
I use `clang main.cpp -lcinder` and I got the headers and libcinder.a installed in the system's directories.

simongeilfus:
Would adding `-std=c++11` help?

mofeing:
@simongeilfus It worked!
Thanks.

PR #1826, opened 2017-04-30
Add missing ; in loadImageAsync()

vinjn:


andrewfb:
Thanks for this catch

PR #1843, opened 2017-05-14
Fix path for xcode/fullbuild.sh in docs

bzalasky:
- Path in docs is currently cinder/xcode/fullbuild.sh, but
the script is actually at cinder/proj/xcode/fullbuild.sh


richardeakin:
Thanks for this.

Issue #1483, opened 2016-06-16
Get all samples building and running with new cmake


richardeakin:
Moved over from #1481.

Most are resources that need to be added to the sample projects.

Need to test on both linux and Mac OS X.

lucasvickers:
OS X and Ubuntu Trusty Currently produces:
```
verbose (CMakeLists.txt) -- adding sample: 1
CMake Error at CMakeLists.txt:54 (add_subdirectory):
  add_subdirectory given source
  "/Cinder/samples/1/proj/cmake" which is not an existing
  directory.
```

richardeakin:
What is the full cmake command?

lucasvickers:
```
mkdir build
cd build
cmake .. -DCINDER_VERBOSE:BOOL=1 -DCMAKE_BUILD_TYPE=Release -DCINDER_BUILD_TESTS=1 -DCINDER_BUILD_SAMPLE=1
```

richardeakin:
I was afraid that this confusion might come up. You either want `-DCINDER_BUILD_ALL_SAMPLES=1` or `-DCINDER_BUILD_SAMPLE={sample name}`.

lucasvickers:
Got it.  I don't think the naming is an issue as long as we document (and I read said documentation)

ufidstudios-ch:
I can look at this again for a few days. Have people stopped arguing over the desired CMake version yet?

PR #1857, opened 2017-06-05
Svg samples CMake build

gaborpapp:
macOS only at the moment


richardeakin:
This is great! One step closer to getting all samples building with cmake (#1483). For linux, I think we're coming to the conclusion to use apt-provided cairo binaries. I've successfully used this [FindCairo.cmake](https://raw.githubusercontent.com/WebKit/webkit/master/Source/cmake/FindCairo.cmake) file on Ubuntu 16.04, however this wasn't in conjunction with cinder's cairo block but rather some completely different setup (on a private client project). Still, hopefully we can adapt that, and then have all the svg samples working on linux, too.

PR #1862, opened 2017-06-17
Fix documentation Converter::convert to match implementation

koenaad:
Matches function documentation with the actual implementation: https://github.com/cinder/Cinder/blob/master/src/cinder/audio/cocoa/CinderCoreAudio.cpp#L92


richardeakin:
Indeed, thanks!

Issue #1701, opened 2017-01-01
audio::BufferView


richardeakin:
Planning to add this during the next round of audio work. Basically it's a subclass of `audio::Buffer` that doesn't own its audio samples, but internally points to a different data store. This will reduce the amount of copying required for some things, and you'll be able to pass them through to existing methods that handle an `audio::Buffer *`.

Related discussion in #1686.

PR #1869, opened 2017-07-09
Add optional sample rate conversion to audio::TargetFile and audio::BufferRecorderNode

koenaad:
As I was discussing with @richardeakin [on the forum](http://discourse.libcinder.org/t/audio-configure-sample-rate-of-bufferrecordernode/730), I've added the abillty to configure the sample rate of ``TargetFile``.
The sample rate conversion is implemented in ``audio::TargetFile`` and exposed in ``audio::BufferRecorderNode``.

I propose the following API changes:
 - ``TargetFile::create`` has an additional optional parameter ``targetSampleRate``.
I placed it before ``extension`` since this parameter is (for some reason) not used in the implementation of ``create``.
 - this extra parameter is propagated to ``TargetFileCoreAudio``.
 - I've added a new function ``TargetFile::getSampleRateNative`` similar as in ``Source``.
 - ``BufferRecorderNode::writeToFile`` has an additional optional parameter ``destSampleRate`` which is passed to ``TargetFile::create``.

I've not implemented these changes on MSW since I work on macOS, but instead added an assert firing when sample rate conversion is required.

I feel like my implementation of ``TargetFile::write`` is bit hacky, because ``dsp::Converter`` 1) processes data in fixed chunk sizes and 2) can't handle frame offsets in buffers. So I first have to copy data into a new buffer (removing the frame offset), convert it using a second buffer before writing it to file.
So could that could be improved imo.

### DIFF
--- a/include/cinder/audio/Target.h
+++ b/include/cinder/audio/Target.h
@@ -32,30 +32,42 @@ namespace cinder { namespace audio {
 
 typedef std::shared_ptr<class TargetFile>		TargetFileRef;
 
+namespace dsp {
+	class Converter;
+}
+
 //! Base class that is used to create and write to an audio destination. Currently only supports .wav encoding.
 class CI_API TargetFile {
   public:
-	static std::unique_ptr<TargetFile> create( const DataTargetRef &dataTarget, size_t sampleRate, size_t numChannels, SampleType sampleType = SampleType::INT_16, const std::string &extension = "" );
-	static std::unique_ptr<TargetFile> create( const fs::path &path, size_t sampleRate, size_t numChannels, SampleType sampleType = SampleType::INT_16, const std::string &extension = "" );
-	virtual ~TargetFile() {}
+	static std::unique_ptr<TargetFile> create( const DataTargetRef &dataTarget, size_t sampleRate, size_t numChannels, SampleType sampleType = SampleType::INT_16, size_t targetSampleRate = 0, const std::string &extension = "" );
+	static std::unique_ptr<TargetFile> create( const fs::path &path, size_t sampleRate, size_t numChannels, SampleType sampleType = SampleType::INT_16, size_t targetSampleRate = 0, const std::string &extension = "" );
+	virtual ~TargetFile();
 
 	void write( const Buffer *buffer );
 	void write( const Buffer *buffer, size_t numFrames );
 	void write( const Buffer *buffer, size_t numFrames, size_t frameOffset );
 
+	//! Returns the user facing sample rate (input)
 	size_t getSampleRate() const	{ return mSampleRate; }
+	//! Returns the true sample rate of the target file. \note Actual input samplerate may differ. \see getSampleRate()
+	size_t getSampleRateTarget() const	{ return mSampleRateTarget; }
+
 	size_t getNumChannels() const	{ return mNumChannels; }
 
   protected:
-	TargetFile( size_t sampleRate, size_t numChannels, SampleType sampleType )
-		: mSampleRate( sampleRate ), mNumChannels( numChannels ), mSampleType( sampleType )
-	{}
+	TargetFile( size_t sampleRate, size_t numChannels, SampleType sampleType, size_t destSampleRate = 0 );
 
 	// Implement to write \a numFrames frames of \a buffer to file. The writing begins at \a frameOffset.
 	virtual void performWrite( const Buffer *buffer, size_t numFrames, size_t frameOffset ) = 0;
 
-	size_t			mSampleRate, mNumChannels;
+	// Sets up samplerate conversion if needed.
+	void setupSampleRateConversion();
+
+	size_t			mSampleRate, mSampleRateTarget, mNumChannels, mMaxFramesPerConversion;
 	SampleType		mSampleType;
+
+	std::unique_ptr<dsp::Converter>	mConverter;
+	BufferDynamic					mConverterSourceBuffer, mConverterDestBuffer;
 };
 
 } } // namespace cinder::audio
--- a/src/cinder/audio/Target.cpp
+++ b/src/cinder/audio/Target.cpp
@@ -22,6 +22,7 @@
  */
 
 #include "cinder/audio/Target.h"
+#include "cinder/audio/dsp/Converter.h"
 #include "cinder/CinderAssert.h"
 
 #include "cinder/Utilities.h"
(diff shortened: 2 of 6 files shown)


richardeakin:
Really excited to see these changes. Just wanted to let you know that I'm onsite for a project and pretty busy for the next week and a half, but looking forward to checking these out when I have a moment.

include/cinder/audio/Target.h
```diff
@@ -32,30 +32,42 @@ namespace cinder { namespace audio {
...
+	virtual ~TargetFile();
 
 	void write( const Buffer *buffer );
 	void write( const Buffer *buffer, size_t numFrames );
 	void write( const Buffer *buffer, size_t numFrames, size_t frameOffset );
 
+	//! Returns the user facing sample rate (input)
 	size_t getSampleRate() const	{ return mSampleRate; }
+	//! Returns the true sample rate of the target file. \note Actual input samplerate may differ. \see getSampleRate()
+	size_t getSampleRateTarget() const	{ return mSampleRateTarget; }
```
richardeakin: Do you think it's worth naming this `getSampleRateNative()` to be consistent with [`SourceFile`](https://github.com/cinder/Cinder/blob/649732c2868315ee5a606befdb2e9e43ff271732/include/cinder/audio/Source.h#L50)? I can't say I like one name over the other, though it is slightly ambiguous that the class is called `TargetFile` and the method has the noun 'target' in in.

src/cinder/audio/Target.cpp
```diff
@@ -48,20 +49,40 @@ std::unique_ptr<TargetFile> TargetFile::create( const DataTargetRef &dataTarget,
...
+	setupSampleRateConversion();
+}
+
+TargetFile::~TargetFile() = default;
+
+void TargetFile::setupSampleRateConversion()
+{
+	if ( ! mSampleRateTarget ) {
+		mSampleRateTarget = mSampleRate;
+	} else if ( mSampleRateTarget != mSampleRate) {
```
richardeakin: Style nit: `else` statements should begin on a new line (apologies, this wasn't in the CONTRIBUTING.md, but I'm going to add it as it is something we've all agreed upon AFAIK).

src/cinder/audio/Target.cpp
```diff
@@ -81,7 +102,29 @@ void TargetFile::write( const Buffer *buffer, size_t numFrames, size_t frameOffs
...
+		auto currFrame = frameOffset;
+		auto lastFrame = frameOffset + numFrames;
+
+		// process buffer in chunks of mMaxFramesPerConversion
+		while ( currFrame != lastFrame ) {
+			auto numSourceFrames = std::min( mMaxFramesPerConversion, lastFrame - currFrame );
+			auto numDestFrames = size_t( numSourceFrames * (float)getSampleRateTarget() / (float)getSampleRate() );
+
+			// copy buffer into temporary buffer to remove frame offset (needed for mConverter->convert)
+			mConverterSourceBuffer.copyOffset( *buffer, numSourceFrames, 0, currFrame );
```
richardeakin: Concerning this and your comment in the PR description about this being a bit hacky, I hope that in the future this problem can be alleviated once `audio::BufferView`s are implemented (see https://github.com/cinder/Cinder/issues/1701). I think for now a TODO comment linking back to that would suffice. The copy is quite cheap compared to the actual encoding, but obviously eventually we want to do as little extra copying as possible.

include/cinder/audio/Target.h
koenaad: For me both seem reasonable. I personally didn't find ``getSampleRateTarget()`` ambiguous since I thought of it as "get _the_ sample rate _of the_ target [file]". But I can definitely see the consistency argument with ``SourceFile``.

src/cinder/audio/Target.cpp
koenaad: I've just fixed this with 89ad51f.
I've also removed spaces between ``if``/``while`` and their braces. This isn't mentioned in the ``CONTRIBUTING.md`` either but is done very consistently in the code.

koenaad:
I can see how the `supportsConversion()` works, I can add this easily. But I'm not familiar with the functionality ExtAudioFile provides and currently don't have to time to figure that out.
Is it okay if I add the `supportsConversion()` function and leave the implementation within `TargetFileCoreAudio` to someone else for now?

I'll also try to adapt the `getSampleRateTarget` to `getSampleRateNative` the next couple of days.

koenaad:
This last commit adds `supportsConversion()` but does not implement it. So what's left for a complete implementation is:
 * implement sample rate conversion in MSW
 * adapt `TargetFileCoreAudio` to use resampling capabilities of `ExtAudioFile`

koenaad:
@richardeakin I don't have experience with resolving conflicts on GitHub. What's the process here? Can I do this since I don't have write access?

Besides a merge conflict in `Target.cpp`, there also have to be some changes in the constructor of `TargetFileOggVorbis`. I can make a diff if that's useful.

richardeakin:
You should probably locally merge master back into your _koenaad:target_sample_rate_ branch, resolve conflicts, then push it back to your repo - it'll update this PR.  You can also rebase and `git push --force` to your repo / branch if you're comfortable with that, either way works.

I started looking at how to use this on Windows but was only able to put a couple hours towards it. I'm hoping to revisit that soon. Basically my current thought that Windows and Linux will need to use a `dsp::Converter`, while OS X / iOS can use the conversion feature built into `ExtAudioFile`, the Core Audio layer doing all the work on those platforms.

koenaad:
Okay, makes sense. Unfortunately, my macbook broke down this week (which has my development environment) so I will only be able to continue working on this over at least a week I think.

richardeakin:
No problem, sorry to hear about your macbook. I may continue with trying to get it working on the windows side, but I can replay my mods on top of your branch once it is fixed up.

koenaad:
Hi @richardeakin, this should now be ready to be merged.

include/cinder/audio/Target.h
```diff
@@ -32,30 +32,44 @@ namespace cinder { namespace audio {
...
   protected:
-	TargetFile( size_t sampleRate, size_t numChannels, SampleType sampleType )
-		: mSampleRate( sampleRate ), mNumChannels( numChannels ), mSampleType( sampleType )
-	{}
+	TargetFile( size_t sampleRate, size_t numChannels, SampleType sampleType, size_t sampleRateNative = 0 );
 
 	// Implement to write \a numFrames frames of \a buffer to file. The writing begins at \a frameOffset.
 	virtual void performWrite( const Buffer *buffer, size_t numFrames, size_t frameOffset ) = 0;
+	//! Implementations should override and return true if they can provide samplerate conversion.  If false (default), a Converter will be used if needed.
+	virtual bool supportsConversion()	{ return false; }
```
richardeakin: I think `TargetFileCoreAudio` still needs to override this and return true.

koenaad:
Hi, sorry for the delay. I haven't had much time to look at this and won't have much in the future either (started a new job, my first actually).

About the OS X / iOS side: I don't have any experience with these services so I won't be able to help much with this. I'd propose to continue with this (non-optimal) implementation for now and opening an issue as reference for future contributors.

If you would like me to merge something, let me know and I'll help out as much as I can. But I won't be able to provide much development for now.

richardeakin:
All good, thanks for your what you've already contributed! I'll try to get to the OS X side the next time I'm on my macbook, which isn't often these days, but I'd like to try to get those changes in there before putting this down and moving on.

PR #1870, opened 2017-07-10
Added implementation of TargetFileOggVorbis

koenaad:
I've added the ability to write .ogg files from ``TargetFile``.

Some work could still be done to expose encoder settings (e.g. the currently hardcoded ``mVorbisBaseQuality``). For now I've simply copied defaults from the Ogg Vorbis examples.

Questions:
 * what's the recommended method to generate a random integer within Cinder? I've used ``Rand::randomize()`` and ``randInt()`` now, but this has some side-effects.

Be aware that this PR is not compatible with #1869. To merge both PR's, the constructor should be updated and ``getSampleRate`` should be replaced by ``getSampleRateTarget``.

### DIFF
--- a/include/cinder/audio/FileOggVorbis.h
+++ b/include/cinder/audio/FileOggVorbis.h
@@ -24,6 +24,7 @@
 #pragma once
 
 #include "cinder/audio/Source.h"
+#include "cinder/audio/Target.h"
 
 //! don't include ogg's static callbacks (we rely on cinder's stream utils instead)
 #define OV_EXCLUDE_STATIC_CALLBACKS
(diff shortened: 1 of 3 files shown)


richardeakin:
Tested and working great, thanks so much for this! Setting bit rate or quality can be added at a later time, this is an excellent addition as is.

Do you think a good order of operations is to merge this, then update it as part of  #1869?

koenaad:
If this is ready to be merged, I would merge this first and then adapt #1869 since there is still some changes to be made there.

I'm still a bit worried about the `Rand::randomize()` at [FileOggVorbis.cpp - line 215](https://github.com/cinder/Cinder/pull/1870/files#diff-4fc2ad56fe8e5b63bf40826c1115e521R215) since this could create weird side-effects if people use a fixed `Rand` seed and Ogg files.
But it's a rather small use case, is this something worth considering or not?

richardeakin:
Hm, using the static `Rand` instance does seem problematic. You could use a local instance of `Rand` instead (perhaps seed it with the current time), but I'm curious why you need a random number there at all? From what I'm reading, that value is the 'stream serial number' as described [here](https://xiph.org/vorbis/doc/framing.html), and it pertains to unique streams within the same file. If that were the case though, I would imagine it would be fine just passing in 0, as we are only using one stream. But perhaps I'm misunderstanding, do you have a better idea what that `serialno` argument is?

koenaad:
Good point.  [Wikipedia](https://en.wikipedia.org/wiki/Ogg_page) mentions:
> Each logical stream must have a unique serial number within a physical stream. It is also intended to be used like a weak hash so that a collision is very unlikely when multiplexing different streams, which eliminates the need for continuous recalculation of page headers at every multiplexing step.

Use case seems limited to multiplexed streams, which is not necessary in this case. Using a constant seems to work fine.

richardeakin:
Thanks again for this, and will hopefully get to #1869 soon.

koenaad:
Cool! Glad I could help out. I'll take a lot at the other PR.

PR #1871, opened 2017-07-11
Use correct gl define flag for RPi.

PetrosKataras:
It seems we missed to update the specific define when switching from the old linux build system to our current configuration. This should also resolve the issue discussed in [this](http://discourse.libcinder.org/t/fbo-rendering-problems-in-cinder-0-9-1-on-raspberry-pi-2/765) forum post.


CallumHoward:
Hi Petros, Callum from the forum post. Thanks for the quick response.
I checked out this PR and build Cinder on the Pi with no problems.
When I went to build the FboBasic sample I got the following error:
```
pi@raspberrypi:~/Documents/Cinder.git/samples/_opengl/FboBasic/proj/cmake/build $ cmake .. -DCINDER_TARGET_GL=es2-rpi
CMake Error at /home/pi/Documents/Cinder.git/proj/cmake/configure.cmake:49 (message):
  CINDER_TARGET not defined, and no default for platform 'Linux.'
Call Stack (most recent call first):
  /home/pi/Documents/Cinder.git/proj/cmake/modules/cinderMakeApp.cmake:23 (include)
  CMakeLists.txt:11 (ci_make_app)
```

CallumHoward:
Adding `-DCINDER_TARGET=linux` makes everything work as expected! Thanks.

PetrosKataras:
Hey, this error you got seems irrelevant to the PR actually and it shouldn't happen in theory since cmake recognizes that we are on a Linux platform so `-DCINDER_TARGET=linux` should not be necessary. I suspect that this has to do with caching of `CINDER_TARGET` somehow - Did you try a fresh build of the sample or did you just used a previously configured build directory after you applied this PR?

CallumHoward:
I tested that again deleting the build folder and you were correct, no need for the CINDER_TARGET define. All resolved, thanks!

richardeakin:
@CallumHoward  Sorry about the `CINDER_TARGET` thing, it is a problem I introduced in https://github.com/cinder/Cinder/commit/5225addafa51a8ff48b7054586ea2d8d5af2f474 - it appears that a CMakeCache.txt that existed before this commit is broken after updating. Regenerating it as you did solves the problem on my end.

PR #1887, opened 2017-08-06
Prevent adding multiple of the same display and removing unconnected displays

notlion:
I've been experiencing an issue on iOS 11 where `screenDidDisconnect` can be called for a `UIScreen` that was not previously connected. This PR checks adds checks so that displays are only added once, and only removed if they exist.


andrewfb:
Thanks for this, Ryan.

PR #1889, opened 2017-08-08
Added guideline about placing else statements on a new line.

richardeakin:
This came up in https://github.com/cinder/Cinder/pull/1869, seems like something we've all agreed on and do throughout the codebase so I added it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,6 +47,7 @@ Please make sure your code conforms to the following general guidelines. If some
   *  ex. `if( ! mEnabled ) { ...`
 * Brackets for class methods and functions begin on a new line. Brackets for everything else (class declarations, if / for / while loops, etc. ) begin on the same line following a space.
 * The contents of an `if`, `for`, or `while` statement should always be on a new line. This not only makes it easier to read but also prevents some ambiguities that come up in some debugging situations, where you can't tell if you've jumped into the body of the statement or not. 
+* `else` statements should be placed on a new line for vertical readability.
 
 #### Types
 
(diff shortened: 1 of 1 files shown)


gaborpapp:
the [google style guidelines](https://google.github.io/styleguide/cppguide.html#Conditionals) also adds this:
"if one part of an if-else statement uses curly braces, the other part must too"
the example in the docs does not fulfill this. what do you think?

richardeakin:
@gaborpapp I'd be fine with that, but I don't feel strongly on whether single line statements contain brackets or not. Curious what others think. I don't believe the current codebase attempts to adhere to this rule, but this is something that is always evolving.

MikeGitb:
I'd put the braces in there. If nothing else, to show wether the else should be put on the same line as the closing brace or on the next.

richardeakin:
@MikeGitb with this new guideline, the `else` should always be on a new line, so that it is vertically aligned with the `if`. If you mean we should add the braces to the one line `if` statement in the guideline example, I can do that, but in general I'm not sure if people will find a rule around this unnecessarily restrictive.

MikeGitb:
@richardeakin: Doesn't hve to be a rule - just an example of the prefered style. If the current style is prefered for one line blocks, then just keep it as it is. 

Personally I find that omitting braces is bad style and dangerous, but much less so between if and else than after else.

richardeakin:
Okay, I changed the example code to reflect this opinion. Preferences from the community are very much appreciated, although I'd like to avoid making hard rules unless there is a consensus among at least the majority of core contributors. I didn't hear much feedback on this one and the current codebase clearly doesn't follow the rule of always using braces for `if` / `else` statements, etc., so I think it makes sense to just leave the example code in 'best practice' form for now.

PR #1890, opened 2017-08-12
Fix FindGStreamer for cases when GStreamer_FIND_VERSION is undefined

raphaelss:
GStreamer_FIND_VERSION isn't defined anywhere for me, so I had to do this change to FindGStreamer.cmake otherwise the gstreamer-1.0 include path wouldn't be passed to the compiler and compilation would fail.

I'm running Fedora 26.


bdon:
This was also necessary for me, following https://libcinder.org/docs/guides/linux-notes/ubuntu.html on fresh Ubuntu 17.10 Artful install

richardeakin:
Thanks for this, please do let us know if anything else comes up when using Fedora (or the new Ubuntu it appears!).

PR #1896, opened 2017-09-18
Add PolyLineT ctor that takes a temporary std::vector

MikeGitb:
Nothing particularly important, but it came up in a program of mine

### DIFF
--- a/include/cinder/PolyLine.h
+++ b/include/cinder/PolyLine.h
@@ -32,22 +32,23 @@ class CI_API PolyLineT {
   public:
 	PolyLineT() : mClosed( false ) {}
 	PolyLineT( const std::vector<T> &aPoints, bool closed = false ) : mPoints( aPoints ), mClosed( closed ) {}
-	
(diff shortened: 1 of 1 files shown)


src/cinder/FileWatcher.cpp
```diff
@@ -45,7 +45,7 @@ class Watch : public std::enable_shared_from_this<Watch>, private Noncopyable {
 	void checkCurrent();
 	//! Remove any watches for \a filePath. If it is the last file associated with this Watch, discard
 	void unwatch( const fs::path &filePath );
-	//! Emit the signal callback. 
+	//! Emit the signal callback.
```
drewish: I'm all for whitespace cleanup but this other file seems out of scope for the PR

MikeGitb:
@drewish: Sorry, you are completely right - don't know how that slippe through.

MikeGitb:
FYI: The ci failure is because the mac tasks were canceled (for whatever reason) not due to an actual error.

richardeakin:
Thanks!

PR #1901, opened 2017-09-20
Fix warnings

MikeGitb:
- Unused parameters in the CINDER_APP macro are causing unused parameter warnings on msvc on W4
- Local variable name EPSILON clashes with global `#define EPSILON EPSILON_VALUE`

### DIFF
--- a/include/cinder/app/msw/AppMsw.h
+++ b/include/cinder/app/msw/AppMsw.h
@@ -109,12 +109,12 @@ void AppMsw::main( const RendererRef &defaultRenderer, const char *title, const 
 	AppBase::cleanupLaunch();
 }
 
-#define CINDER_APP_MSW( APP, RENDERER, ... )													\
-int __stdcall WinMain( HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine, int nCmdShow )\
-{																									\
-	cinder::app::RendererRef renderer( new RENDERER );												\
-	cinder::app::AppMsw::main<APP>( renderer, #APP, ##__VA_ARGS__ );							\
-	return 0;																						\
+#define CINDER_APP_MSW( APP, RENDERER, ... )                                                                        \
+int __stdcall WinMain( HINSTANCE /*hInstance*/, HINSTANCE /*hPrevInstance*/, LPSTR /*lpCmdLine*/, int /*nCmdShow*/ )\
+{                                                                                                                   \
+    cinder::app::RendererRef renderer( new RENDERER );                                                              \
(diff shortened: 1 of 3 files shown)


richardeakin:
I think you have that pesky setting on in Visual Studio where it decides to reformat all whitespace every time it saves a file? I believe that is why all the diffs are popping up on white lines, which makes it difficult to see what the real changeset is. My philosophy is that it is best to leave whitespace lines untouches, even if there is for example an extra tab or something on the line, so that diffs are improved.

Interesting on the `ESPILON` confusion, does indeed look like, at least intellisense, doesn't know which version to use. I think renaming it to all lowercase in `Ray::calcTriangleIntersection()` matches the style though, as it is a const local variable.

MikeGitb:
> I think you have that pesky setting on in Visual Studio where it decides to reformat all whitespace every time it saves a file? 

Kind of. It "only" trims trailing white spaces, which reduces the amount of noise in my one repos, but sadly - as it doesn't know about VCS - it doesn't distinguish between new and old ones. I thought the changes in Font.cpp and Ray.cpp where few enough that you wouldn't mind, but I can certainly remove them.

The whitespace changes around the  `CINDER_APP_MSW` where deliberate in order to align the `\` at the end of the lines. Shall I also revert those?

> I think renaming it to all lowercase in Ray::calcTriangleIntersection() matches the style though, as it is a const local variable

Sure, will do.

MikeGitb:
On a side note: Have you considered to reduce your build matrix and/or activate auto cancellation (if you haven't done so) on Travis a to speed up the checking of pull requests ?

richardeakin:
Ok. I think in general the less the better, unless there is already so many mods that the whitespace changes don't create anything new to look at. Thanks for explaining the whitespace changes on `CINDER_APP_MSW`, those make sense to me now and the mods look good.

> On a side note: Have you considered to reduce your build matrix and/or activate auto cancellation (if you haven't done so) on Travis a to speed up the checking of pull requests?

On reducing the build matrix, I also feel that we don't need to build for so many versions of Xcode - I think oldest, latest, and major versions would be fine. Concerning 'auto cancellation', is this something we can modify with the travis.yml file? I'm not familiar with it. Perhaps worthy of a separate issue around speeding up CI builds.

MikeGitb:
Auto cancellation is an option you can set in your Travis account (l don't know what the default is). This will cancel builds when there is already a newer commit on the same branch / PR. As the xcode builds often take hours until they get started, (due to backlog in travis' global open source queue queue queue

MikeGitb:
@richardeakin: Regarding autocancellation, it is now activated by default: https://blog.travis-ci.com/2017-09-21-default-auto-cancellation

PR #1909, opened 2017-09-29
[Vulkan] Resolve incompatible extensions

MaciejJesionowskiAMD:
Device creation code is enabling all available extensions, but some extensions are exclusive. E.g. VK_KHR_maintenance1 cannot be used at the same time as VK_AMD_negative_viewport_height.

**How to test:**
On recent AMD drivers that expose these two extensions any Cinder Vulkan app will crash upon restart with VK_ERROR_INITIALIZATION_FAILED from vkCreateDevice.


vinjn:
Looks good to me.

DominikWitczakAMD:
Can we please have that patch merged into mainline? Thanks.

ryanbartley:
Looks good to me, too.

PR #1916, opened 2017-10-09
Removed dependency on boost.preprocess

lithium-snepo:
Unrolled declarations generated by `BOOST_PP_SEQ_FOREACH`

### DIFF
--- a/src/cinder/Stream.cpp
+++ b/src/cinder/Stream.cpp
@@ -30,7 +30,7 @@
 #include <limits>
 #include <boost/scoped_array.hpp>
 #include <iostream>
-#include <boost/preprocessor/seq/for_each.hpp>
+
 using std::string;
 
 namespace cinder {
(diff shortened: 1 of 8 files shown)


src/cinder/Stream.cpp
```diff
@@ -792,17 +792,105 @@ StreamExc::StreamExc( const std::string &fontName ) throw()
 
 /////////////////////////////////////////////////////////////////////
 
-#define STREAM_PROTOTYPES(r,data,T)\
```
vinjn: Maybe we can leverage STREAM_PROTOTYPES() to reduce code duplication?

```
STREAM_PROTOTYPES(int8_t)
STREAM_PROTOTYPES(uint8_t)
STREAM_PROTOTYPES(blabla_t)
```

lithium-snepo: Kind of embarrassed to admit that didn't even occur to me, mostly because i avoid macros like the plague, but also these were generated by a script I wrote that pulled the definitions out of the preprocessed source, but if the preference is to tuck them back behind macros i can submit a different pull request.

vinjn: My concern is if one day new interfaces were added, you would have to copy-paste at multiple lines.

PR #1917, opened 2017-10-09
split string without boost

Hperigo:
As requested in here:
http://discourse.libcinder.org/t/request-for-contribution/892

Some observations:
1. When using **_compress = true_**, the current version adds some empty strings to a vector and removes the separator, is that expected?

2. made a little test using this string: "Cinder,OpenFrameworks-- ThreeJs, Processing-";
[link](https://gist.github.com/Hperigo/d991b64a663cb796825377b02047f2ce)

```
Custom impl took: 1.5974e-05
0. 'Cinder' 
1. 'OpenFrameworks' 
2. '' 
3. '' 
4. 'ThreeJs' 
5. '' 
6. 'Processing' 
7. '' 
--------

boost took: 3.79682e-05
0. 'Cinder' 
1. 'OpenFrameworks' 
2. '' 
3. '' 
4. 'ThreeJs' 
5. '' 
6. 'Processing' 
7. '' 
```

### DIFF
--- a/include/cinder/Utilities.h
+++ b/include/cinder/Utilities.h
@@ -47,9 +47,9 @@ CI_API fs::path getDocumentsDirectory();
 CI_API void launchWebBrowser( const Url &url );
 	
 //! Returns a vector of substrings split by the separator \a separator. <tt>split( "one two three", ' ' ) -> [ "one", "two", "three" ]</tt> If \a compress is TRUE, it will consider consecutive separators as one.
-CI_API std::vector<std::string> split(const std::string& str, char separator, bool compress = true );
+CI_API std::vector<std::string> split(const std::string &str, char separator, bool compress = true );
 //! Returns a vector of substrings split by the characters in \a separators. <tt>split( "one, two, three", " ," ) -> [ "one", "two", "three" ]</tt> If \a compress is TRUE, it will consider consecutive separators as one.
-CI_API std::vector<std::string> split(const std::string& str, const std::string &separators, bool compress = true );
+CI_API std::vector<std::string> split(const std::string &str, const std::string &separators, bool compress = true );
 
 //! Loads the contents of \a dataSource and returns it as a std::string
 CI_API std::string loadString( const DataSourceRef &dataSource );
--- a/include/cinder/Utilities.h
+++ b/include/cinder/Utilities.h
@@ -47,9 +47,9 @@ CI_API fs::path getDocumentsDirectory();
 CI_API void launchWebBrowser( const Url &url );
 	
 //! Returns a vector of substrings split by the separator \a separator. <tt>split( "one two three", ' ' ) -> [ "one", "two", "three" ]</tt> If \a compress is TRUE, it will consider consecutive separators as one.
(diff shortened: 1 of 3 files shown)


src/cinder/Utilities.cpp
```diff
@@ -76,20 +76,22 @@ std::vector<std::string> split(std::string str, const std::string &separators, b
...
+		}
 		else if(searchPos > searchPrevPos){
 			result.push_back(str.substr(searchPrevPos, searchPos - searchPrevPos));
 		}
 		
 		searchPrevPos = searchPos+1;
 	}
 	
-	if (searchPrevPos < str.length()){
+	if (searchPrevPos <= str.length())
```
vinjn: Why < becomes <=?

include/cinder/Utilities.h
```diff
@@ -47,9 +47,9 @@ CI_API fs::path getDocumentsDirectory();
 CI_API void launchWebBrowser( const Url &url );
 	
 //! Returns a vector of substrings split by the separator \a separator. <tt>split( "one two three", ' ' ) -> [ "one", "two", "three" ]</tt> If \a compress is TRUE, it will consider consecutive separators as one.
-CI_API std::vector<std::string> split( const std::string &str, char separator, bool compress = true );
+CI_API std::vector<std::string> split(std::string str, char separator, bool compress = true );
```
vinjn: Why `const std::string&` becomes `std::string`

src/cinder/Utilities.cpp
Hperigo: In the current implementation, if we have a separator at the end of the string like so: 
“Hperigo,vinjn,” 

we have the following result:
```
0. 'Hperigo' 
1. 'vinjn' 
2. '' 
```
So the  '<=' is there just to copy the current behavior. It appends the last character if it's a separator, but I will admit that if we just leave < it the result looks better:

```
0. 'Hperigo' 
1. 'vinjn'
```

include/cinder/Utilities.h
Hperigo: I was testing with some methods that would erase the string but there is no need for it not be const now. Thanks for spotting it.

richardeakin:
Thanks for this. I'm not sure about your question on `compress` - can you provide an example?

I think this would be a great candidate for a unit test (there is already a file for Utilities.h tests [here](https://github.com/cinder/Cinder/blob/master/test/unit/src/Utilities.cpp)). Also, code should follow cinder's [style guide](https://github.com/cinder/Cinder/blob/master/CONTRIBUTING.md#style-guide).

Hperigo:
**_I'm not sure about your question on compress - can you provide an example?_**
yes, when splitting **"one----two"** and **compress = false** we have the following result:
0. 'one' 
1. ''
2. ''
3. 'two' 

I think it's just weird to have a vector with empty strings. But boost has the same behaviour so I guess it ok.

**_I think this would be a great candidate for a unit test (there is already a file for Utilities.h tests 
here)_**
done!

_Also, code should follow cinder's style guide._
cool, I did some minor fixes, let me know if there is anything off

andrewfb:
This looks great - I manually merged it. Thanks for taking this on (and its unit test)

PR #1918, opened 2017-10-09
Unrolled dependency on boost.preprocess

lithium-snepo:
Removed the dependency on `BOOST_PP_*`, but using @vinjn's advice to keep the original macros


andrewfb:
Thanks for this - my very quick build benchmark shows it reduced overall Cinder build time 4-5% as a bonus.

vinjn:
wow, boost is so good at increasing build time...

PR #1925, opened 2017-11-01
Androidstudio cmake build update

nickblock:
Updated the androidStudio gradle projects and CMakeFiles.
This enables the sample projects to be built with the current androidStudio 3.0 tooling.
This enables native and java debugging capabilities of CInder projects using androidStudio on android devices.


richardeakin:
This is great, thanks so much before doing this! I also don't have any android devices to test at the moment, but hopefully we can get some other community members to try it out. Maybe an RFC on the forum makes sense? There are a handful of people there that have been making the android stuff work for their own needs.

nickblock:
Hmm, having trouble on windows with that. I've got a few more environments to test with now. Let me get another update to you shortly.

nickblock:
Right, quite confident about that now. Got AndroidStudio building and debugging on windows too with a few different devices.

andrewfb:
Thanks a lot for going after this. Work commitments will likely delay me 1-2 weeks but I can test this on Windows with an S8+ once I'm in front of the right computers again.

PetrosKataras:
I wanted to try this out on macOS 10.12.5 so I went ahead and installed Android Studio 3.0.1 and tried to import `Cinder/proj/android/libcinder_java/build.gradle` but the sync process fails with the following output:  

screen shot 2018-02-07 at 5 53 42 pm

First attempt on Android so I m not familiar at all with the eco-system there and I might have missed something obvious though..

PetrosKataras:
Apparently this was the old `gradle` build file and the one that should be used is located under `Cinder/src/cinder/java/build.gradle` now. I also had to switch to gradle 3.3 by modifying `gradle-wrapper.properties` and changing `distributionUrl=https\://services.gradle.org/distributions/gradle-2.2.1-all.zip` to `distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-all.zip` 
for successfully building BasicApp. 

Cleaning `Cinder/proj/android` from old/obsolete stuff ( which occupy, currently, most of the space there ) would be a great next step probably. Also, I think that it would be nice if the `build.gradle` file could be located in this directory, as it used to, since this way the structure remains the same as in other platforms.. Not sure though what implications this might have..

PR #1935, opened 2017-11-22
setThreadName()

andrewfb:
This PR implements `setThreadName()` for MSW, UWP, iOS, macOS, and Linux.

The behavior of this method is environment-specific. For example, in Visual C++ the results are visible immediately. In Xcode, they are only visible if you attach to a running app, rather than launching from within the IDE. In Linux they can be seen by executing `info thread` from gdb.


paulhoux:
I really like this. Had a similar implementation in my toolbox, but it's great to see it being part of Cinder and cross-platform now.

Issue #1837, opened 2017-05-08
The doc's search doesn't work properly


JonasSeegers:
If you click on a link suggested by the search on https://libcinder.org/docs/guides/index.html (or any subpage) you get a 404 because those links miss the "/docs/" part. For example if you search for "Cinder" and click on the link it will lead to https://libcinder.org/classorg_1_1libcinder_1_1_cinder.html, which is a 404. It should lead to https://libcinder.org/docs/classorg_1_1libcinder_1_1_cinder.html.

anselanza:
Yes, please! Drives me absolutely nuts every time.

anselanza:
The frustrating thing is that the search works perfectly when Cinder docs are generated locally, i.e. served from `/docs/html` relative to the repo root. 

So everything *looks* like it works fine. But it is also impossible to fix, because in a local environment it would break the search to change all the links to `docs/`.

But the actual hosted version of the docs is always `https://libcinder.org/docs/` which breaks the search again.

I don't how the hosting is currently set up, but could I suggest that a subdomain be set up for the docs? `docs.libcinder.org` or something? I know there's a bit of pain with having some broken links (although that can be mitigated with a few server-side conditional redirects to catch people using the old URLs) but it would fix the ridiculous broken search functionality straight away.

Any other ideas?

PR #1939, opened 2017-12-07
Docs search functionality fix

gregkepler:
Made 'docs/' the default root directory for http or http protocol version of the docs.
Addresses #1837


andrewfb:
Thanks for this - addressed server-side instead

PR #1941, opened 2017-12-09
Isolating attribution and if condition

cpsw:
I'm suggesting a code style change that I would use in my code. I heard some discussions about this issue, and I would like to suggest it to see if it is more a personal preference or whether it makes sense. 

It is just to make explicit that an assignment expression has the value of the left operand after the assignment.


MikeGitb:
Not sure I understand your style argument, but  doesn't this create an infinite loop?

cpsw:
I was trying to make the assignment explicit, but you are right, @MikeGitb. I'm sorry about that.

PR #1943, opened 2017-12-12
Fix GLES/EGL on Linux

thyrrestrup:
This fixes GLES on EGL Context through GLFW on Linux.
I've removed a couple of deprecated `_GLFW_` macros and added the appropriate `glfwWindowHint` for EGL context creation.
Also `glx_context.c` was added as it is needed in both the GLX and EGL context case.


thyrrestrup:
Forum thread: http://discourse.libcinder.org/t/cinder-egl-linux-problem/988

richardeakin:
This all looks good to me, thanks for explaining all of the changes. I don't have a machine where I can test at the moment but it appears from the forum post that this is fixing a few people's problems so I think we should merge it soon.

PR #1956, opened 2018-01-09
gl::drawSolidTriangle with vec3 vertices

gaborpapp:
`gl::drawSolidTriangle` vertex parameters are `vec2` and if you call it with `vec3`'s they can get converted silently cutting the z-coordinate displaying the triangle in 2d instead of 3d. This simple PR is a modification of the function for `vec3` vertices to address this.


andrewfb:
Thanks for this.

PR #1960, opened 2018-01-18
[CI] Reduce test matrix on travis for xcode

MikeGitb:
This PR removes xcode 8.1 and 8.2 from the test matrix with the goal of reducing travis-CI runtime. 
The remaining xcode versions are 7.3, 8 and 8.3 which should still cover most interesting cases. 

Currently, Travis seems to have some trouble anyway. Regardless of this however, I don't think testing on every version of XCode adds much value (unless you had problems in the past with a specific verion), but as travis is notoriously short on MacOS resources, it can greatly increase the ci runtime duration.

(Btw. I think we talked about this before somewhere, but I can't find the issue at the moment)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,16 +29,6 @@ matrix:
         - TOOL="xcode-osx"
         - DESCRIPTION="OS X build/test via Xcode"
     - os: osx
-      osx_image: xcode8.2
-      env: 
-        - TOOL="xcode-osx"
-        - DESCRIPTION="OS X build/test via Xcode"
-    - os: osx
-      osx_image: xcode8.1
-      env:
-        - TOOL="xcode-osx"
-        - DESCRIPTION=OS X build/test via Xcode"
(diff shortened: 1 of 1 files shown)


lucasvickers:
Agreed, but while you are in there it may be worth updating against current available images, i.e. xcode 9 (assuming there's support for it)

https://docs.travis-ci.com/user/reference/osx/

MikeGitb:
If that is supported by cinder (no reason why it shouldn't) sure. Does XCode 9.2,8.3 and 7.3 around reasonable?

lucasvickers:
Yeah, latest xcode for each major version makes sense to me!

andrewfb:
I agree - @MikeGitb do you want to tweak this per your previous comment?

MikeGitb:
Sorry, forgot about that. Here you go.

andrewfb:
Great - thanks for this.

Issue #1959, opened 2018-01-18
[Filesystem MSVC] Update namespace


MikeGitb:
As far as I can tell, Visual studio 2017 has removed the std::tr2::sys namespace when compiling in c++17 mode and you get a warning about it in c++14 mode. The recommended way to use the filesystem library is currently to use the `std::experimental::filesystem` namespace (they haven't yet ported it into the std namespace). 
My guess (haven't tested it) is that this namespace is also available in MSVC2015: https://msdn.microsoft.com/de-de/library/hh874694.aspx

I can provide a PR to update this.

richardeakin:
Are you thinking to add in an additional compiler check to cinder/Filesystem.h? I'd say go for it, seems inevitable.

MikeGitb:
No, I'd just replace namespace std::tr1::sys with std:: experimental::filesystem. That should work for VS2015 and 2017. Sometime in the future, one should add a check that enables std::filesystem, but IIRC, there is currently no compiler providing it (afaik gcc8 will)

PR #1962, opened 2018-01-22
[MSVC - FS] Set correct namespace for filesystem on msvc

MikeGitb:
See https://github.com/cinder/Cinder/issues/1959


andrewfb:
Tested this; works in VC2015 including UWP. Thanks for this. Closes #1959.

Issue #1957, opened 2018-01-10
Unhandled audio formats on Linux


CallumHoward:
Issue for [this](http://discourse.libcinder.org/t/segfault-when-writing-audio-to-file-on-linux/993) discussion on the forums.

## Issue: 
On Linux when using `audio::BufferRecorderNode`, writing to an audio file (eg. `.wav`) causes a segfault due to an unhandled case [here](https://github.com/cinder/Cinder/blob/master/src/cinder/audio/Target.cpp#L51-L59).
Note: `.ogg` file format works.

## To reproduce:
I have modified the InputAnalyzer sample to save to a file, while will reproduce the error on linux.
Code can be found [here](https://gist.github.com/CallumHoward/763c429cf2f0ed04859581e3662e035a/revisions).
Just change the path on line [97](https://gist.github.com/CallumHoward/763c429cf2f0ed04859581e3662e035a/228eff9a02978ff3a55aa48fe21223717b93fcb9#file-inputanalyzerapp-cpp-L97).

richardeakin:
I can take care of this in a planned audio maintenance PR, although I've been quite busy on other work as of late so haven't gotten to it yet. If anyone would like to submit a PR I'd be happy to get it in there. Should be trivial to add.

PR #1963, opened 2018-01-23
Audio maintenaince

richardeakin:
* address #1957
* remove `boost::tribool` usage


andrewfb:
Nice - one of the last remaining Boost references. Looks good to me.

PR #1969, opened 2018-02-08
Extend CI matrix for ubuntu

MikeGitb:


MikeGitb:
The failure on appveyor seems to be a sporadic problem on the ci-platform and not related to the pr.

richardeakin:
Thanks for this, really nice to have wider ubuntu coverage. I triggered a rebuild for appveyor which should of course pass and I think we can merge after that.

PR #1970, opened 2018-02-08
[cmake] fix required cmake version to 2.8.12

MikeGitb:


richardeakin:
Thanks for this and sorry for the delay.

Do you think we should raise the required version for samples (and probably for the unit test) too? Or maybe grep and change them all to match.

MikeGitb:
Sorry for not comming back to this.
Yes, just grep and replace them all would probably be the correct way forward although.

FYI:
> The cmake_minimum_required() command does more than report an error if a too-old version of CMake is used to build a project. It also sets all policies introduced in that CMake version or earlier to NEW behavior.

Somewhat related: Now that ubuntu18.04.1 is about to be released, have you considered a general tool update (like cmake 3.5 and g++-5 which are the default versions on ubuntu 16.04 or 3.0/g++-4.9 If you want to support Debian Jessie (I don't know the numbers on MacOS)?

MikeGitb:
Maybe this can be addressed once the official toolchain requirements are updated.

PR #1972, opened 2018-02-16
Fixed linux x11 clipboard problem

origamidance:
Allows access the system clipboard now.


andrewfb:
Merged - thanks for this

Issue #1971, opened 2018-02-12
getDefaultOutput() crashes on windows audio device


dsowa-w1:
I have a cinder app that uses the audio output and calls getDefaultOutput(). This crashes in the windows case because the windows function CoInitialize() has never been called.  There are other parts of the cinder code that do call CoInitialize() via the cinder function ci::msw::initializeCom() but you can use audio without needing any of those code paths. I can work around this by calling initializeCom() in my audio initialization. I believe this is really a need for someone with a good working knowledge of the windows cinder code to call CoInitialize() at an appropriate place.

richardeakin:
Maybe we should just add an additional call to [`ci::msw::initializeCom()`](https://github.com/cinder/Cinder/blob/master/include/cinder/msw/CinderMsw.h#L139) in the `audio::ContextWasapi` constructor? Not sure why I didn't add that, I guess I assumed that the AppMsw already had called it but your right, you should be able to use the `ci::audio` stuff without an app.

Note that there is a [related TODO](https://github.com/cinder/Cinder/blob/master/src/cinder/audio/Context.cpp#L74) that would need to be taken care of for the audio::Context to work without an app (this is a trivial change, though).

PR #1977, opened 2018-02-28
Make ci::audio usable when there is no app on MSW

richardeakin:
Addresses issues in #1971. @dsowa-w1 can you try this and see if it works for you?


dsowa-w1:
Yes I'll try and get to it in the next day or so. Thanks for looking at this.

richardeakin:
I'm going to merge this as I believe it is all necessary, let me know if we need other modifications to run ci::audio without an app.

Issue #1979, opened 2018-03-01
Exception when DataSource URL contains 260+ characters on VS 2017


benjaminbojko:
It looks like currently there's a limitation with `fs::path` in VS 2017 that throws an `invalid char filename argument` exception if the path is 260 or more characters long. I haven't been able to dig super deeply yet (or test on OS X), but it seems like it's an issue that's being worked on at MS: https://developercommunity.visualstudio.com/content/problem/167666/maximum-filesystempath-length.html

For reference, this **works**:

```c++
DataSourceUrl::create(Url(string(259, '.')), UrlOptions());
```

...and this **throws the exception**:

```c++
DataSourceUrl::create(Url(string(260, '.')), UrlOptions());
```

The reason I'm posting here is that this affects also happens when loading remote URLs via HTTP, since the `DataSourceUrl`  constructor sets `fs::path mFilePathHint` to the entire URL: https://github.com/cinder/Cinder/blob/master/src/cinder/DataSource.cpp#L151.

Is there any way to circumvent this at the moment? It looks like `getFilePathHint()` is [mostly used](https://github.com/cinder/Cinder/search?utf8=%E2%9C%93&q=getfilepathhint&type=) for convenience in samples to get things like file extensions. Would there be any way to make this opt-in or an on-demand value?

Happy to help on a PR with some guidance on best approach here.

andrewfb:
The filePathHint is also used by `ImageIo` in order to determine a file is a .png for example, since it's just binary data by the time it arrives at `loadImage()`. This also gets used a couple of places though that it would be hard to break (SVG wants relative paths for external files as I recall).

This is a tricky problem - how much we accommodate a VC++ bug (not the first) but my gut would be that we should `#ifdef CINDER_MSW` a bit of code in `DataSourceUrl::DataSourceUrl()` that tests to see if the length of the URL exceeds 260 and then if it does, instead passes to `setFilePathHint()` the equivalent of `fs::path::filename()`. This would be basically to `rfind( '/' )` on the `Url`. That should in turn be verified as shorter than 260 characters and if it's not, we could just punt and copy the last 260 chars.

benjaminbojko:
Thanks for the quick response and context.

I can take a stab at implementing the MSW workaround you suggested in the next few days and test + make a PR. Will also put some thought into how that could work with URLs with long queries after the extension like `http://mycdn.com/img.png?12345[...]260`, but at the very least this should squash the majority of issues.

[...]

PR #1980, opened 2018-03-02
Workaround for VC++ 14 bug with URLs with 260+ chars

benjaminbojko:
First pass at addressing #1979. This patch resolves the original issue we had at our studio with loading URLs with  long query strings (e.g. [this graphql query](https://www.graphqlhub.com/graphql?query=%7B%09%09%09graphQLHub%09%09%09giphy%7B%09%09%09%09random(tag:%22javascript%22)%20%7B%09%09%09%09%09id%09%09%09%09%09url%09%09%09%09%09images%7B%09%09%09%09%09%09original%7B%09%09%09%09%09%09%09url%09%09%09%09%09%09%7D%09%09%09%09%09%7D%09%09%09%09%7D%09%09%09%7D%09%09%7D)).

A couple of notes:

* Since yesterday (03/01/2018) the bug is reported as [resolved on the MS dev forums](https://developercommunity.visualstudio.com/content/problem/167666/maximum-filesystempath-length.html) and will be included in the next VS compiler release. Because of that I limited the patch scope to [`_MSC_VER`](https://blogs.msdn.microsoft.com/vcblog/2017/11/15/msvc-conformance-improvements-in-visual-studio-2017-version-15-5/) and not globally to `CINDER_MSW`. Thoughts?
* The same issue of converting the URL string to `fs::path` in order to conveniently get the filename popped up in two places: `DataSourceUrl` and `IStreamUrl`. Currently I have nearly the same snippet in both locations, but am happy to consolidate if there are any suggestions for where that could/should live.
* I followed @andrewfb's suggestion to attempt a simple filename retrieval using an `rfind('/')` . This works for our use-cases but I don't have a ton of insight into what other situations the `DataSourceUrl::getFilePathHint()` and `StreamBase::getFileName()`. Below the line it's better than the exception though. Any other suggestions are welcome, but I felt like a simple fix now is better than a beefier workaround later.

Happy to iterate over this.

### DIFF
--- a/src/cinder/DataSource.cpp
+++ b/src/cinder/DataSource.cpp
@@ -148,7 +148,19 @@ DataSourceUrlRef DataSourceUrl::create( const Url &url, const UrlOptions &option
 DataSourceUrl::DataSourceUrl( const Url &url, const UrlOptions &options )
 	: DataSource( "", url ), mOptions( options )
 {
+#if _MSC_VER >= 1900 && _MSC_VER <= 1912
+	std::string urlString = url.str();
+
+	if ( urlString.size() >= 260 )	{
+		// Workaround for MSVC++ 14.0/17.0 exceptions once fs::path length hits 260
+		const size_t start = urlString.rfind( '/' );
+		urlString = urlString.substr( start == std::string::npos ? 0 : start, 259 );
+	}
+
+	setFilePathHint( urlString );
+#else
 	setFilePathHint( url.str() );
+#endif
 }
 
 void DataSourceUrl::createBuffer()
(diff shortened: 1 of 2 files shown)


andrewfb:
Sorry for the delay here; this looks good to me except for the `<=1912`. Specifically, I just tried this with the very recently released 2017 (`1913`) and it's still crashing. For now maybe we just say (`>=1900`) and then we'll revisit when the bug is actually fixed?

MikeGitb:
As far as I understand from the linked issue, this bug will stay in `std::experimental::filesystem` and will only be fixed in `std::filesystem` (which afaik isn't released for msvc yet)

MikeGitb:
Is there a reason for the minimal mscv version?

benjaminbojko:
Both good points. I tested this with VS 2015's native toolsets and they still have the same issue, so might as well do a catch-all for MSW at this point instead of limiting it to versions. New commit applies to all `CINDER_MSW` builds.

MikeGitb:
Is this still relevant or has it been superseded by cinders move to VS2019 and std::filesystem?

PR #1981, opened 2018-03-04
Allow override of filename extension in audio::TargetFile::create

MikeGitb:
See https://github.com/cinder/Cinder/issues/1629

Note, this is an abi and potentially API breaking change, but as it doesn't serve any purpose and has a default value I doubt this will be a problem.

### DIFF
--- a/include/cinder/audio/Target.h
+++ b/include/cinder/audio/Target.h
@@ -35,8 +35,8 @@ typedef std::shared_ptr<class TargetFile>		TargetFileRef;
 //! Base class that is used to create and write to an audio destination. Currently only supports .wav encoding.
 class CI_API TargetFile {
   public:
-	static std::unique_ptr<TargetFile> create( const DataTargetRef &dataTarget, size_t sampleRate, size_t numChannels, SampleType sampleType = SampleType::INT_16, const std::string &extension = "" );
-	static std::unique_ptr<TargetFile> create( const fs::path &path, size_t sampleRate, size_t numChannels, SampleType sampleType = SampleType::INT_16, const std::string &extension = "" );
+	static std::unique_ptr<TargetFile> create( const DataTargetRef &dataTarget, size_t sampleRate, size_t numChannels, SampleType sampleType = SampleType::INT_16, std::string extension = "" );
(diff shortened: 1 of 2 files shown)


richardeakin:
Thanks for this, although I wonder if we should actually use that `extension` instead, to stay in parity with the [ImageIo loadImage / writeImage](https://github.com/cinder/Cinder/blob/master/src/cinder/ImageIo.cpp#L414-L470) functionality.

If we did that, I believe adding an additional (optional) `string extension` param to the [`audio::Source`](https://github.com/cinder/Cinder/blob/master/include/cinder/audio/Source.h) load and create methods makes sense too.

MikeGitb:
Just to make sure I understand this correctly: In `Imagelo` the purose of the extension parameter is to give the user a means to override the actual file extension correct? 

Sure, I can implement that for `TargetFile` too. To be completely honest, I mainly created this PR to close the remaining points in #1629 (I hate to leave things half finished).

andrewfb:
@MikeGitb - just a slight amendment, some sources/targets may not have a file path associated with them (loading a PNG from RAM, for example) so the ImageIo machinery needs a hint about which loader/writer to pass the data to. Loaders can also register themselves as generic, meaning they can attempt to infer the file type from the data itself, but that's a sort of last resort.

richardeakin:
@MikeGitb koodoos on that sentiment. :) I think it actually just shows a gap where we should finish implementing that work. There are also cases where you may end up loading an audio file that was compressed into a larger binary resource, for example, and you wouldn't know the extension. I can think of a similar case when writing to said binary object.

If you'd like to have a go at implementing this, by all means please do. I'd just rather leave the warning 
 (and this github issue) as a reminder to make sure that we get to it.

MikeGitb:
OK, I changed the PR to now - instead of removing the extension parameter - uses it to override the extension of the file.

MikeGitb:
ping

andrewfb:
I'm going to go ahead and merge this even in light of upcoming work to be done, since it fixes a legitimate bug (specifically in the case of a user-supplied `"ogg"` extension). Thanks again for this.

MikeGitb:
Thanks

PR #1987, opened 2018-03-17
Windows Low Latency audio support

richardeakin:
This is some work I've been needing to get in for a long time now, which adds the ability to use WASAPI Exclusive mode to achieve low latency with some sound cards. This allows things like audio visualizers, for example, to be much tighter in sync, along with any sort of interactive audio playback / synthesis one might want to do.

I say some cards, because it doesn't work for me on some more consumer grade products, ex. a logitech webcam microphone, but it works quite well in the tests I've done with realtek and pro audio soundcards. There's still a couple loose ends to take care of:

- [x] more testing with pro audio gear (might need to add support for 24-bit sample conversion
- [x] finish testing whether it is ok to have non-power-of-two buffers. May need to update docs around this too.

I also began looking at Windows 10 Shared Mode low latency (explained [here](https://docs.microsoft.com/en-us/windows-hardware/drivers/audio/low-latency-audio)), however I found that driver support for it is almost non-existent, so I'm not sure yet whether it is something we should enable for the moment. I may end up enabling it behind a macro, purely so we can test as new devices with new drivers come out.


richardeakin:
This is ready, tested with as many devices as I could get my hands on. It should be known that not all samplerates and hardware settings work in exclusive mode, you sort of have to find those that do. I guess it is a per-vendor driver thing.

I think we could use some better 24bit int support for exclusive modes on some devices, but it doesn't appear to be required so I'm planning to do that in the next PR.

PR #1992, opened 2018-03-27
Clipper alternative to Boost.Geometry | Removing ConvexHull

andrewfb:
This PR removes integration of the infrequently-used Boolean operations on `PolyLine` and instead implements them using the [Clipper](http://www.angusj.com/delphi/clipper.php) library as a CinderBlock. This removes one of our last remaining Boost dependencies, and also adds some nice new functionality in the form of offset calculations. This also has the benefit of reducing Cinder binaries substantially. I'm seeing ~10mb on Linux and ~20mb on Windows (both Debug) with this change.

**Update** - obviously not ideal to not provide a direct alternative, but this PR now also removes the convex hull functionality, which is *not* implemented in Clipper. However in pursuit of removing Boost dependencies I believe this is the right tradeoff.


andrewfb:
image
New offset functionality in the updated sample.

richardeakin:
I like the freestanding method approach, compared to static methods on `ci::PolyLine`.

Feels like the methods in `CinderClipper.h` should be within a namespace, perhaps just `namespace cinder`?

PR #1994, opened 2018-04-03
Fixing “qualified reference to 'Format' is a constructor name rather …

jacobsologub:
…than a type in this context” error.

Build Error with Latest Version 9.3 (9E145) of Xcode.


andrewfb:
Thanks for this fix

PR #1997, opened 2018-04-20
Add support for headless rendering on Linux through EGL.

PetrosKataras:
This PR adds support for headless ( i.e no requirement to run an X server ) rendering on Linux through `EGL` and the `EGL_EXT_platform_device` extension.

This enables Linux-based command-line Cinder applications to get access to an OpenGL Core Profile context in order to achieve hardware accelerated rendering even when no physical display is attached.

Mesa doesn't implement, as of yet, the full functionality for this extension so for now this should work only on NVidia hardware with any recent driver version.

In order to enable this rendering path you would need to configure Cinder with :  

` cmake .. -DCINDER_HEADLESS_GL=egl`

When this path is enabled there will be no support for mouse / keyboard input but everything else should work the same as in a normal windowed app ( with no actual window running in this case but instead an `EGL PBuffer` surface ).


PetrosKataras:
A local copy of EGL ( 1.5 ) header file dependencies is now included with Cinder which should simplify compilation on machines that these files are not present.

A test app is also now included under `Cinder/test/Linux/HeadlessTestApp` for anyone who would like to try this out in an easy way. 

For trying this out without X, on *Ubuntu instances with the default `lightdm` display manager, you can :  
- `ctrl+alt+F1` in order to switch to a `tty` terminal
- login 
- run ` sudo service lightdm stop`. 

This should stop X and from the same terminal session you can compile Cinder and the test app and run from command line.

The test app should produce the following `.png` snapshot in the directory where you executed the binary from:  

headless-egl-snapshot

PetrosKataras:
Superseded by #2001 which includes the work presented here and adds support also for headless software rendering through `OSMesa`.

Issue #1991, opened 2018-03-24
Support conventional CMake usage


numberZero:
It’s unusual for CMake to create an application using third-party command. Usually, `add_executable` is used. I checked your code, that’s what Cinder actually uses internally, with minor tweaks only. It would be very nice if you support using just that, and add separate commands for e.g. adding a block.

Also, forcing the debug build of an application to use debug build of Cinder looks wrong. If you want to debug an application, that doesn’t mean you want to debug the engine as well.

And, well, storing several configurations in a single build tree is foreign for CMake, so should be optional at least. Don’t forget you are able to have several build trees attached to the same project in CMake (e.g. I have **7** build trees in one project, like `build.i686-w64-mingw32.static`, `build.x86_64-linux-gnu.release`, and so on).

PetrosKataras:
Hi,

> It’s unusual for CMake to create an application using third-party command...

I m not sure I understand what the issue at hand is here to be honest. One can still write the `CMakeLists.txt` file as one likes. The Cinder-specific CMake macros exist only for convenience  and can be completely omitted if you want to setup things on your own.

> Also, forcing the debug build of an application to use debug build of Cinder looks wrong...

From my point of view this is a valid point and could be improved. 

> And, well, storing several configurations in a single build tree is foreign for CMake, so should be optional at least...

You can have multiple build trees by using different build directories so I might be missing something otherwise I m not sure what the issue is here.

PetrosKataras:
I have been thinking about the build type linking issue and my approach would be to leave the default behavior as is and introduce two new Cinder-specific CMake options in `ci_make_app`, namely `CINDER_APP_LINK_RELEASE` and `CINDER_APP_LINK_DEBUG` that the user can set in order to force linking with a specific build of Cinder when building an application.

This would be the least invasive change to my opinion although we could also consider to make `CINDER_APP_LINK_RELEASE` the default option when building an app.

Also, personally, I would not necessarily consider wrong the current approach ( there are various frameworks that assume as default this behavior ) but having the flexibility to change this is definitely something that we should address. 

This is of course relevant only when you want to use the Cinder specific macros. If you setup things on your own, you can freely link with whatever build type you desire.

numberZero:
#1999 resolves only one point. Others remain.

> and add separate commands for e.g. adding a block.

so that that functionality is not lost when using conventional `add_executable` instead of ~proprietary~ non-standard`ci_make_app`.

> storing several configurations in a single build tree

[...]

PR #1999, opened 2018-04-22
Add CMake option for linking applications with a specific Cinder build type ( Debug, Release )

PetrosKataras:
The current CMake configuration links applications based on the current `CMAKE_BUILD_TYPE` so that a Debug ( or Release ) build of the application will link with the Debug( or Release ) build of Cinder.

This can be a reasonable default but there should be an option for the user to override this behavior if desired. 

This PR leaves the default behavior unchanged but adds support for applications to explicitly request the Cinder build type they would like to link with. 

This can be achieved now by configuring the application with 

`cmake .. -DCINDER_APP_LINK_RELEASE=1` to link with the Release version of Cinder

or

`cmake .. -DCINDER_APP_LINK_DEBUG=1` to link with the Debug version of Cinder.

Note that this option is relevant only when configuring/building the application with the `ci_make_app` macro.

Relevant discussion #1991

### DIFF
--- a/proj/cmake/modules/cinderMakeApp.cmake
+++ b/proj/cmake/modules/cinderMakeApp.cmake
@@ -49,7 +49,7 @@ function( ci_make_app )
 
 	# This ensures that the application will link with the correct version of Cinder
 	# based on the current build type without the need to remove the entire build folder
-	# when switching build type after an initial configuration. See PR #1518 for more info. 
+	# when switching build type after an initial configuration. See PR #1518 for more info.
 	if( cinder_DIR )
 		unset( cinder_DIR CACHE )
 	endif()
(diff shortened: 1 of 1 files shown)


MikeGitb:
Just as a word of caution: I think the mscv toolchain doesn't support mixed linking, because the release and debug standard library are not ABI/API compatible.

PetrosKataras:
Thanks for the heads up -- If we can verify this is indeed the case we can exclude MSW from this option.

PetrosKataras:
Looking a bit more into it, it indeed seems that on MSW this is not a good idea and should be avoided. I will update the PR and exclude this platform from the option.

@MikeGitb  Thanks again for catching this!

PR #2001, opened 2018-04-25
Add support for headless rendering on Linux through EGL and OSMesa.

PetrosKataras:
This PR supersedes #1997 and besides hardware accelerated headless rendering through `EGL` adds also support for pure software headless rendering through `OSMesa` with `llvmpipe` or Intel's `OpenSWR` software rasterizer. 

It introduces `RendererGlLinuxHeadless` and `WindowImplLinuxHeadless` that shares the code between these two paths ( with the potential to add more in the future - see `Mesa GBM` ) and moves the RPi specific code back on its own implementation since it feels that this platform is specific enough to warrant a clear separation.

**Note that all the instructions and comments in #1997 are still valid for the `EGL` path so please have a look there on how to enable and test that option.**

The new `OSMesa` path allows to run OpenGL applications with Core Profile support even on machines with no GPU attached ( all rendering happens on the CPU(s) ). Benchmarking shows that depending on the available hardware and render load, the `OSMesa` path can be from 2x to 20x slower than the equivalent hardware accelerated path through `EGL`.

Nevertheless it can provide a nice fallback renderer in certain edge case scenarios and could also allow for example, integrating GL tests in the build system ( or CI ).

It requires a recent version of `llvm` installed ( anything above `5.0` should do it - latest is better ) and a Mesa version that is at least `> 12.0` ( Ideally latest version of Mesa should be used which at this time is `18.1.0` )

To enable this option configure Cinder with:  

`cmake .. -DCINDER_HEADLESS_GL=osmesa`

You can optionally also pass `OSMESA_ROOT` in order to specify a location of Mesa that is not installed system wide ( suggested if on the same machine there are other drivers installed already ).

`cmake .. -DCINDER_HEADLESS_GL=osmesa -DOSMESA_ROOT=/path/to/mesa/folder`

The test application under `Cinder/test/Linux/HeadlessTestApp` initially aimed at the `EGL` path can be used also here for easily testing.


andrewfb:
This looks great to me. Have verified it on both Ubuntu 16.04 and 18.04 with EGL.

PR #2004, opened 2018-05-03
Updated FreeType library src and includes to version 2.9

gregkepler:
Updated to latest version of FreeType 2.9
[View Changelog](https://sourceforge.net/projects/freetype/files/freetype2/2.9/)

Note: this flag has been specifically turned on (in ftoption.h) to allow png rendering in fonts.
#define FT_CONFIG_OPTION_USE_PNG


PetrosKataras:
Looks like it requires `libpng12-0` to be installed on the system.

andrewfb:
I think the new plan here would be to not build with PNG support built-in as it will require all Cinder apps to link against `libpng`. In the meantime, cases which require it (which means you're probably a) doing custom type rendering and b) require color emojis) will require a Cinder rebuild.

Issue #1968, opened 2018-02-08
shared library build fails with Cinder-shipped boost-filesystem library


flv0:
Related to #1915, building Cinder as shared library fails because the (static) boost-filesystem version that is shipped along Cinder has been built without PIC. Passing -DCINDER_BOOST_USE_SYSTEM=1 resolves the issue for me.

One option would be to rebuild the shipped boost accordingly. However my personal stance on the question of shipping 3rd party libraries alongside a software package or using the system-supplied version is strongly in favor of using system libraries, if there is not a very strong reason to include a specific (or modified) version of that library. For boost-filesystem, I really don't see a reason why Cinder should not rely on a system-wide deployed version.

One reason is that things will get awry when users start to link their application to other libraries which in turn link against the system-wide version of boost-filesystem transitively. That kind of dependency mess and potential symbol/ABI collisions is something I try to avoid at all cost. If Cinder aspires to be included with major Linux distributions I think this is something which needs to be taken care of. Since you propose in #1915 to make the system library the default in the next release, I would even argue to drop shipping a prebuilt version of boost altogether.

Apart from this issue I can report that Cinder builds just fine on current Debian stable (stretch).

andrewfb:
I think we're leaning toward making this the default now. Just a forward-thinking comment, we're quite close to removing the Boost dependencies other than Filesystem. I wonder if we could get away with requiring only [libboost-filesystem-dev](https://packages.ubuntu.com/trusty/libboost-filesystem-dev)

This seems to exist for Ubuntu 14.04+ and Raspbian Jessie.

Issue #1915, opened 2017-10-07
Linking error with libboost_filesystem when building BasicApp


George3d6:
When trying to build an app with cinder (including the "BasicApp" example) I encountered the following error at linking:

```
/usr/bin/ld: /home/george/Cinder/lib/linux/x86_64//libboost_system.a(error_code.o): relocation R_X86_64_32 against `.rodata.str1
.1' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: /home/george/Cinder/lib/linux/x86_64//libboost_filesystem.a(operations.o): relocation R_X86_64_32 against symbol `_ZN5boost10filesystem16filesystem_errorD1Ev' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: /home/george/Cinder/lib/linux/x86_64//libboost_filesystem.a(path.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Nonrepresentable section on output
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/BasicApp.dir/build.make:117: Debug/BasicApp/BasicApp] Error 1
make[1]: *** [CMakeFiles/Makefile2:68: CMakeFiles/BasicApp.dir/all] Error 2
make: *** [Makefile:84: all] Error 2

```

I managed to compile Cinder without any issue, though I am not working on a supported distro.

Any idea as to why this is happening would help.

The only info I can really make sense of from the error is that I should recompile something with the -fPIC flag... but what that flag does or what the entity I should recompile is I can't figure out.

PetrosKataras:
Are you by any chance on Arch ? Even if not the idea of what to try should be the same as the one described in [this forum post](http://discourse.libcinder.org/t/solved-cinder-with-arch/887/2?u=petros) .

George3d6:
Yep, removing the boost libs from x86 and running` cmake .. -DCINDER_BOOST_USE_SYSTEM=1`  seems to have fixed the issue.

Though I assume cinder apps are supposed to ship with those specific boost libraries, right ?

PetrosKataras:
[...]

PR #2007, opened 2018-05-09
Switch to system installed Boost on Linux.

PetrosKataras:
This PR switches the default behavior on Linux from linking with the precompiled Boost libs to system installed versions.

Requires `libboost-filesystem-dev` to be installed through the package manager of your platform/distribution.

This should resolves issues like [this](https://discourse.libcinder.org/t/linker-errors-mongodb-cxx-driver-on-linux/504), #1915, #1968  and a few more of similar nature.


proj/cmake/platform_linux.cmake
```diff
@@ -191,17 +191,9 @@ if( GSTREAMER_GL_INCLUDE_DIRS AND GSTREAMER_GL_LIBRARIES )
...
-else()
-	set( LINUX_LIB_DIRECTORY "${CINDER_PATH}/lib/linux/${CINDER_ARCH}/" )
-	list( APPEND CINDER_LIBS_DEPENDS
-		${LINUX_LIB_DIRECTORY}/libboost_system.a
-		${LINUX_LIB_DIRECTORY}/libboost_filesystem.a
-	)
-endif()
+find_package( Boost 1.54 REQUIRED COMPONENTS system filesystem )
+list( APPEND CINDER_LIBS_DEPENDS ${Boost_LIBRARIES} )
+list( APPEND CINDER_INCLUDE_SYSTEM_PRIVATE ${Boost_INCLUDE_DIRS} )
```
richardeakin: Does this mean that we no longer be shipping any boost libraries on linux, in other words the `CINDER_BOOST_USE_SYSTEM` option is the only option? If so I'm not against it, just thought we had discussed switching the default behavior but still shipping prebuilt binaries. Maybe that has shown itself to be more trouble than its worth, though.

PetrosKataras: My opinion would be to make the default and only behavior systemwide installed Boost. It has proven indeed problematic in the past even by just checking out the Boost submodule and then having mixed includes ( e.g. coming from Cinder's version and from a system wide installed version of Boost ) Not linking with the prebuilt binaries is not really enough in this case.

PetrosKataras:
Had forgotten that the binaries are not part of the submodule, so I just removed them and this should be complete now from my side I believe.

rdtor:
Hi Petros how can i use this PR, do i need to build cinder again from your repo https://github.com/PetrosKataras/Cinder/tree/linux-boost-use-system-install?

PetrosKataras:
The simplest thing to try first would be to remove your previous Cinder build folder and re-compile Cinder with:

`cd Cinder && mkdir build && cd build && cmake .. -DCINDER_BOOST_USE_SYSTEM=1 && make -j4`

Make sure you have `libboost-filesystem-dev` installed and also make sure that you completely remove the `Cinder/include/boost` folder before trying this out.

If you want to apply this PR then you should do what @gaborpapp suggested in #2088 in order to checkout a branch locally in your machine with the changes applied and then recompile Cinder as usual. 

i.e `git fetch origin pull/2007/head:pr2007`

rdtor:
I applied this PR and followed what @gaborpapp suggested and the build was succesfull thanks

andrewfb:
I'm intending to merge this in the next day or two unless anyone has any counterpoints

richardeakin:
@PetrosKataras question:

> The simplest thing to try first would be to remove your previous Cinder build folder and re-compile Cinder with:
> 
> cd Cinder && mkdir build && cd build && cmake .. -DCINDER_BOOST_USE_SYSTEM=1 && make -j4
> 

Doesn't this PR remove the use of `-DCINDER_BOOST_USE_SYSTEM=1`? As in, after this PR the only option is to use system boost on linux.

PetrosKataras:
@richardeakin yeah this is correct - the part of the comment you quote was in the case that the user didn't want to apply this PR locally and instead try installed boost with the current Cinder setup ( i.e without this PR merged ).

PetrosKataras:
Just noticed the README mentioning a recursive git checkout. This should not be required for Linux actually after this PR is merged and in fact it should be discouraged since this would pull the include files for Boost which could still cause some weird clashes with system installed Boost ( ..it would also pull TinderBox which is also completely irrelevant to the Linux setup.. ).

I suppose we should add a note about this?

richardeakin:
Just tested this with both the command line and Visual Studio Code on Ubuntu 19.04, looking good!

If it can cause header mismatches, I do think we should remove the `--recursive` flag from the git command for the linux docs pages. We may end up using other submodules later, but hopefully we'll have ditched the boost submodule by that point.

PR #2020, opened 2018-07-23
Glad OpenGL extensions loader support for MSW and Linux

paulhoux:
This PR will replace `glLoad` with the [GLAD](https://github.com/Dav1dde/glad) extension loader, allowing us to add OpenGL v4.6 features to Cinder. glLoad has not been updated in a while, so we're looking for an alternative. 

When the GLAD files are generated without additional extensions, Cinder users will be forced to create a `RendererGl` of the correct version like so: `CINDER_APP( MyApp, RendererGl( RendererGl::Options().version( 4, 6 ) ) )` to use the desired features.

To prevent confusion, however, we should add commonly used extensions when generating the GLAD files. I've added a minimal set of extensions, enough to make Cinder work out of the box. Before accepting this PR, it would be smart to discuss what extra extensions we might need.

GLAD features a [webservice](http://glad.dav1d.de/) that makes it extremely easy to update the source files. See `glad/glad.h`, which is the main header of the loader. At the top, you'll find the exact settings used to create the files. If we want to add additional extensions, use the provided information as a starting point for the webservice and generate a new header. 

An alternative to GLAD is [FlextGL](https://github.com/mosra/flextgl), which requires Python to generate the files. It would allow us to automate the process, if we're Python savvy, that is. But I'd say that GLAD's webservice is a much easier alternative which does not require us to install any additional development software.

*The Linux version has not been tested and probably needs a few fixes to work. However, I have included the required GLX files, so this should not be too hard.*


paulhoux:
Oh, by the way: a good strategy to figure out which extensions we'd like to add by default, is to generate a header for OpenGL v3.2 (which is the lowest version currently supported by Cinder), then compile and run all tests and samples. Missing functions will then cause a runtime error. Figure out which extension provides the function and add it to the list. The result will be a comprehensive list of all extensions we'd like to support.

I would still recommend to generate the header for OpenGL v4.6, so that users can always use the latest features by specifying `version( 4, 6 )` when creating `RendererGl`.

vinjn:
I assume glew will also be removed?

paulhoux:
I have replaced GLEW with GLAD for OpenGL ES. Not sure if GLEW is used for anything else. If not, in can (and probably will) be removed as well.

Note that this branch may not become part of the official Cinder release. We're still discussing our options. But I welcome your input.

vinjn:
I like this change :)

paulhoux:
I should probably remove some commits from this PR. After my current project is done, I'll revisit this PR. Hopefully by that time, the Travis build will be fixed as well.

MikeGitb:
> Hopefully by that time, the Travis build will be fixed as well.

This should do it: https://github.com/cinder/Cinder/pull/2027

(Just in case you haven't seen it)

paulhoux:
I just rebased this PR on `cinder/master`, but I realize the Travis fix has not been merged to master yet, so this rebase is pretty pointless in itself. To the Cinder devs: would it be possible to merge Mike's fix? I'd do it myself, but a) I am notoriously bad in testing PR's before merging and b) I don't feel confident enough touching the Linux codebase.

MikeGitb:
If the is no response: My PR just adds one line to the dependency installation script and doesn't really "touch the code base".

You can probably just add the same line to your own PR and travis should be happy.

PetrosKataras:
Just a note, that a fix for this is also exists as part of #2007.

paulhoux:
OK, cool, Travis works again. The build fails for Mac OS X targets, because I went ahead and removed glLoad altogether. We should probably make the switch to GLAD on Mac OS X as well. If anyone could help me out with that in this PR, that would be awesome.

richardeakin:
I've been a behind on this, though talked a bit with Paul a while back offline, and I just wanted to say that I do think we should use this as a jumping off point for switching to a newer / savvier gl loader.

To my mind, GLAD is seeming like the favorite because of its ease of configuration and popularity, however I still think we should take this opportunity to compare FlextGL generated sources, primarily to compare the code length of each.  I did manage to get the FlextGL python stuff working a while back (there was some broken dependency for 64-bit), and I've been meaning to use this to at least compare generated output sources between the two. This is at the top of my todo list. I think if the generated code is similar in quality / length, we should go ahead and get the OS X stuff in there, then look to merge this.

paulhoux:
Thanks, Rich, for your input and support. Do you think you could try to add the support for GLAD on Mac OS X? I could probably make it work, but I am still on non-Mojave OS X and I feel I don't have enough experience with XCode to know for certain that my changes would work.

earjuice:
I have to say, I've been using Paul's GLAD reverse-z branch on Windows 10 for about a year now and I love it.  Was kinda worried it wouldn't be integrated into the master branch so I'm glad(haha) I stumbled on this thread.
For what it's worth, to add extensions, all I did was generate a glad.zip file from [this website](https://glad.dav1d.de/) using every available extension for my graphics card…
And replaced cinder’s glad,h and glad.c…
I found it to be a rather simple solution to get all the extensions my card supported.. Not sure how many extensions you were thinking of including but this is a pretty easy fix if you want to use it in the readme.  I totally fully support the notion of merging this into the master branch.  Full gl extension functionality within a couple steps! Cheers!

richardeakin:
Well, I have to apologize if I've been holding this one up. I haven't been able to find the time to test FlextGL, and with the added benefit of not needing python to use GLAD (along with @paulhoux already doing all this work), I agree we should go this route.

We still however need to go about adding GLAD support on other platforms, specifically OS X and Linux.  If anyone can help with this, the contribution is much appreciated. These days I'm not on OS X or linux much..

paulhoux:
I for my part can make sure this branch stays up to date with master, so that it will be easy to later merge the two. My access to Linux and MacOSX is limited if not non-existent. So any help in that department is appreciated. All GLAD files are there, they "only" need to be incorporated into the Cinder base.

totalgee:
This is great, thanks! With this PR integrated, will we have access to the GLAD (or GLFW) "loader" function, so we can easily load other extension functions that aren't included in the set you decide to include? That would be nice. Otherwise, I've got a couple more extensions to add to your list...

totalgee:
For the record, here's what I was using with GLAD and OGL 4.5:

extensions="GL_ARB_base_instance,GL_ARB_bindless_texture,GL_ARB_buffer_storage,GL_ARB_explicit_uniform_location,GL_ARB_multi_draw_indirect,GL_ARB_sampler_objects,GL_ARB_shader_draw_parameters,GL_ARB_shading_language_include,GL_ARB_texture_storage,GL_ARB_texture_storage_multisample,GL_ARB_viewport_array,GL_ATI_meminfo,GL_EXT_direct_state_access,GL_EXT_framebuffer_multisample_blit_scaled,GL_EXT_window_rectangles,GL_KHR_debug,GL_NVX_gpu_memory_info,GL_NVX_linked_gpu_multicast,GL_NV_clip_space_w_scaling,GL_NV_framebuffer_multisample_coverage,GL_NV_geometry_shader_passthrough,GL_NV_gpu_multicast,GL_NV_stereo_view_rendering,GL_NV_viewport_array2"

Just in case it's useful to compare with your list. (Some aren't needed, as they'd be included if you're using  OGL 4.5 or 4.6)

PetrosKataras:
It would be great to revive this - I just tested it with the DeferredShadingAdvanced sample on both macOS 10.13.6 and Linux 18.04 with CMake and current master branch and things worked out of the box for me on both platforms. 

One thing to note is that recent versions of glad's webservice do not omit all the KHR code when the relevant option is selected which can lead to compilation issues due to missing type definitions. I was able to resolve this by omitting this option and including also the KHR generated header file. This glad [issue ticket](https://github.com/Dav1dde/glad/issues/211) has more info on this.

richardeakin:
I marked this for 0.9.3, makes sense to me to make a switch like this then since 0.9.2 release is around the corner. What is left for this to do, then? To me it seems:

- [ ] decide what extensions will be included
- [ ] any project file changes necessary on OS X / linux (@PetrosKataras) ?

Anything else?

PetrosKataras:
I tried only the CMake path and that is only with a couple of the advanced opengl samples but with that configuration I didn't need to do any project file changes for this to work on either OS X or Linux. I suspect though that the hand-tuned Xcode project files would still need to be updated for this to work?

richardeakin:
(in part to bump this PR...)

@PetrosKataras yea at a minimum, the old gl_load_cpp.cpp file will have to be removed from the xcode project, and add glad/glad.c + whatever mac-specific glad files exist.  Looks like this has already been updated here for cmake.

paulhoux:
OK, I have been looking into this some more. The way I see it, there are 2 possible ways we switch from glLoad to GLAD:
* Either we create the smallest possible header files, with (in the case of Desktop OpenGL) support for the latest version (currently 4.6) and just the extensions required to successfully compile Cinder. For some application, this means you will have to specify in the `CINDER_APP` macro which version you'd like to run on. If your user's hardware does not support it (even if the driver supports the required extensions), the application will not run. To fix this, users are required to generate their own GLAD headers and recompile Cinder.
* Or we create GLAD headers with support for most if not all extensions. The GLAD files will become quite big that way, but all possible functionality will be in there and only in very rare cases will people need to generate their own GLAD files. Maybe only to reduce file size.

So far I have opted for option 1, but with [GLAD2](https://gen.glad.sh/) I was able to make option 2 work. Before continuing with this PR, I would like to thoroughly test it on all supported platforms and even under different driver versions. Having access to a proper test script which builds and runs the samples would be great in that respect. Maybe we should focus on that, first, before embarking on a complete change of extension loader.

num3ric:
It's possible to include all extensions with the current command-line version of Glad by omitting the `--extensions` argument. The following does so (while also including KHR platform as is now needed apparently).

`pip install glad`
`glad --profile="compatibility" --api="gl=4.6" --generator="c" --spec="gl" --out [...]`
`...`

Here's a build time comparison without/with all the extensions on my machine:
* 1 minute 12.2 seconds (with a few more than your initial selection)
* 1 minute 14.53 seconds (with all extensions)

For a cost of 2 seconds, maybe this all-inclusive option is the path of least resistance, and most flexible for all users?

richardeakin:
@num3ric is this compile time measurement for libcinder or the user app? For me, I think we can be more lenient for an increased time in cinder.lib, although user apps get rebuilt / hotloaded much more often so there's the real place to take concern.

totalgee:
Given what's been going on in the world this year, I can't say I'm sorry to see 2020 closed and forgotten... (Sorry, couldn't resist ;-)

PR #2028, opened 2018-08-15
Add "getItems()" to Timeline

Hperigo:
Currently theres no way to retrieve all the items in the timeline, It would allow for some nice debug visualizations, [like so](https://imgur.com/a/FaEgrsU)

gist example:
https://gist.github.com/Hperigo/c635975c3a4fd7fbc62c782c99341f59


andrewfb:
Would this be better if it returned a `const std::multimap<void*,TimelineItemRef>&` instead?

andrewfb:
Looks good, though I'd propose we skip adding the sample to the official distribution (though it looks cool). I can cherry-pick the changes or if you'd rather not do a different PR.

Hperigo:
awesome! can you do the cherry pick?

PR #2029, opened 2018-08-20
Update Renderer.cpp

dgarroyo:
Most of the Cairo samples and Render2dBasic were affected by a problem with rendering to the window's dc instead of the backbuffer (mDoubleBufferDc ).
This change makes the geDC() in Renderer.cpp call the right method.
The mImpl->getDc() method is declared in cinder::app::msw::RendererImpl2dGdi.h as:
virtual HDC getDc() const { return ( mDoubleBuffer ) ? mDoubleBufferDc : mPaintDc; }


andrewfb:
Thanks very much for tracking this down.

PR #2032, opened 2018-09-11
Add support for copy/paste to AntTweakBar on Mac

timmb:
The included version of AntTweakBar does not support copying and pasting on Mac. This PR modifies AntTweakBar using a patch from Alec Jacobson (http://www.alecjacobson.com/weblog/?p=2378) to include support for it by calling out to the OS binaries `pbcopy` and `pbpaste`.

NB the shortcut is still CTRL+C and CTRL+V instead of CMD+C/CMD+V.


vinjn:
I love this feature :)

andrewfb:
I'm up for this - if nobody has other thoughts I'll aim to merge by the end of the weekend.

PR #2036, opened 2018-10-09
TinyEXR update to v0.9.5

paulhoux:
This will update TinyEXR to v0.9.5, allowing us to support a few more EXR file types, among other things. Cinder's `ImageFileTinyExr` classes have been modified to work with the new library. This change should be transparent to the users and should also be backward compatible, with one exception: when exporting EXR files, they are now converted to 16-bit floats instead of 32-bit floats. This cuts file size in half.

With the new TinyEXR version, it's now relatively easy to add support for file compression as well, but I've not done so yet to conserve CPU and memory and remain backward compatible.


src/cinder/ImageFileTinyExr.cpp
```diff
@@ -46,119 +49,132 @@ void ImageSourceFileTinyExr::registerSelf()
...
+	if( status != TINYEXR_SUCCESS )
+		throw ImageIoExceptionFailedLoadTinyExr( string( "Failed to parse OpenEXR version" ) );
+
+	if( version.multipart || version.non_image )
+		throw ImageIoExceptionFailedLoadTinyExr( string( "Multipart or DeepImage EXR's are not supported yet" ) );
+
 	if( dataSource->isFilePath() ) {
-		status = ParseMultiChannelEXRHeaderFromFile( mExrImage.get(), dataSource->getFilePath().string().c_str(), &error );
-		if( status != 0 )
+		const auto filename = dataSource->getFilePath().string().c_str();
```
andrewfb: Looks like this is getting a pointer to the internal storage of a temporary (`string()` is a temporary). I can fix this when I merge.

andrewfb:
I added a comment on a bug but otherwise this looks good to me - thanks very much for this upgrade; it adds a much wider variety of EXRs that we can parse now. I agree we should revisit the zipped vs uncompressed output later.

PR #2039, opened 2018-10-14
GstPlayer GL fixes.

PetrosKataras:
This PR contains some critical fixes for the case of multiple gst-gl pipelines running in parallel. It should address issues like the one described in the following forum post https://discourse.libcinder.org/t/gstreamer-linux-issue/1312 and also some other memory issues that would manifest deep in the GStreamer pipeline.

A note that these changes render #1947 obsolete. I could open a new PR with an updated patch for Windows once this is merged and assuming there is interest.

### DIFF
--- a/include/cinder/linux/GstPlayer.h
+++ b/include/cinder/linux/GstPlayer.h
@@ -109,7 +109,6 @@ struct GstData {
 	GstElement* videoBin = nullptr;
 	GstPlayer* player = nullptr;
 #if defined( CINDER_GST_HAS_GL )
-	GstGLContext* context = nullptr;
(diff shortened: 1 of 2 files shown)


brucelane:
I'm interested in this for Windows, thank you

richardeakin:
By obsolete you mean it will be conflicting afterwards? I was under the impression that #1947 was only adding some things and even if gstreamer stays as experimental on windows at the moment, it isn't causing any new issues. If this is correct I say we move towards getting them both merged (and sorry it grew a little stagnant).

richardeakin:
Ok makes sense to me too, merging.

Issue #2041, opened 2018-10-18
OSC Messages can't provide the sender port


totalgee:
In the previous OSC block (from 0.9.0), it was possible to obtain the Message sender's address **and** port, using `Message::getRemoteIp()` and `Message::getRemotePort()`. But it seems the port-accessing functionality has been removed in the ASIO-based OSC block. I see there is a `Message::getSenderIpAddress()`, but nothing to get the remote port.

This functionality is important (to me ;-), for example if you want a UDP "server" to respond back to the sender(s) of request messages.

I will shortly submit a pull request to add `Message::getSenderPort()`.

cinder:
This should be addressed #2042

PR #2042, opened 2018-10-18
Add OSC Message support to access the sender's port

totalgee:
- add method osc::Message::getSenderPort(), which returns the port of
  the remote sender; a companion method to getSenderIpAddress().
- resolves issue #2041.

### DIFF
--- a/blocks/OSC/src/cinder/osc/Osc.cpp
+++ b/blocks/OSC/src/cinder/osc/Osc.cpp
@@ -78,19 +78,20 @@ inline double ntohd( int64_t x ) { return (double) ntohll( x ); }
 //// MESSAGE
 
 Message::Message()
-: mIsCached( false )
+: mIsCached( false ), mSenderPort( 0 )
 {
 }
 	
 Message::Message( const std::string& address )
-: mAddress( address ), mIsCached( false )
+: mAddress( address ), mIsCached( false ), mSenderPort( 0 )
 {
 }
 	
 Message::Message( Message &&message ) NOEXCEPT
 : mAddress( move( message.mAddress ) ), mDataBuffer( move( message.mDataBuffer ) ),
 	mDataViews( move( message.mDataViews ) ), mIsCached( message.mIsCached ),
-	mCache( move( message.mCache ) ), mSenderIpAddress( move( message.mSenderIpAddress ) )
+	mCache( move( message.mCache ) ), mSenderIpAddress( move( message.mSenderIpAddress ) ),
+	mSenderPort( message.mSenderPort )
 {
 	for( auto & dataView : mDataViews ) {
 		dataView.mOwner = this;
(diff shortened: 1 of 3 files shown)


totalgee:
An alternative approach might be to just expose the remote (sender's) `endpoint`, but this would change the existing `Message` interface (e.g. if we were to remove `Message::getSenderIpAddress()` and replace it with `Message::getSenderEndpoint()`, which encapsulates both address and port). But then, what type would it return (since `endpoint` is a template class, according to the Internet protocol)?  I'm not sure how you'd want to handle TCP vs UDP (and potentially vs ICMP) endpoint types (a std::variant in C++17?).

It's probably simpler to stick with what I've done here, returning the sender IP address (as before: `getSenderIpAddress()`) and port (new method: `getSenderPort()`) separately. In any case, this is likely not a common thing most people will need to do.

blocks/OSC/src/cinder/osc/Osc.cpp
```diff
@@ -863,7 +867,7 @@ std::ostream& operator<<( std::ostream &os, const Message &rhs )
 {
 	os << "Address: " << rhs.getAddress() << std::endl;
 	if( ! rhs.getSenderIpAddress().is_unspecified() )
```
totalgee: Note that `address::is_unspecified()` is not properly implemented in the [ASIO version Cinder uses](https://github.com/cinder/Cinder/blob/5586c04e9e07e09a269e30fdd357ed5329911616/include/asio/ip/impl/address.ipp#L201-L212) -- it will return `false` for an empty address (or any address with type `none`, as opposed to `ip4` or `ip6`).

I see later versions of ASIO [have fixed this](https://github.com/chriskohlhoff/asio/blob/b0926b61b057ce563241d609cae5768ed3a4e1b1/asio/include/asio/ip/impl/address.ipp#L195-L200).  Anyhow, not a big deal here -- we'll continue to assume the method returns the right thing, as the previous code did. The normal thing for users is to output a `Message` you just received, which **will** have the sender data filled in properly (it will be "specified").

richardeakin:
I can see this would be a handy addition, and don't see any overhead in it so why not.

As an aside related to the case you mentioned on the forum: you could send the port or some unique id in your initial message to your receiver, so that it knew where the message was coming from. But I don't see any problem in having many ways to solve a problem. :)

cinder:
I agree - thanks for this @totalgee - merging now.

totalgee:
Thanks for accepting it!

totalgee:
> As an aside related to the case you mentioned on the forum: you could send the port or some unique id in your initial message to your receiver, so that it knew where the message was coming from. But I don't see any problem in having many ways to solve a problem. :)

@richardeakin Yes, that's another option, though not as convenient as just getting the info (which is available from the endpoint anyhow) directly from the Message, and keeping such details out of your application's OSC messaging protocol.

Thanks again.

PR #2043, opened 2018-10-18
Remove boost from cinder json

MikeGitb:
This removes the boost dependencies from Json.cpp and Json.h.

The second commit only contains some small modifications I made while I was messing with the function anyway, but it came with a lot of whitespace changes, so this PR is probably easier to review commit by commit.

### DIFF
--- a/include/cinder/Json.h
+++ b/include/cinder/Json.h
@@ -1,7 +1,7 @@
 /*
  Copyright (c) 2012, The Cinder Project
  All rights reserved.
- 
+
  This code is designed for use with the Cinder C++ library, http://libcinder.org
 
  Redistribution and use in source and binary forms, with or without modification, are permitted provided that
@@ -68,7 +68,7 @@ class CI_API JsonTree {
 	  private:
 		bool	mIgnoreErrors, mAllowComments;
 	};
-	
+
 	//! Options for JSON writing. Passed to the \c write method.
 	class CI_API WriteOptions {
 	public:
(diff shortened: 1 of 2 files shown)


MikeGitb:
All cis error out, so apparently I broke something. Will investigate,  fix and report back.  Sorry for that.

MikeGitb:
Just to answer your question: It compiles with clang 6.0 on my local machine and all travis jobs seem to build successfully, so I assume yes, It works with clang.

Could you elaborate, where the "incomplete type requirements" come into play?

MikeGitb:
OK, I found the bug and now all tests pass (including the ones using clang). 

I've squashed the commits and appended a commit that does purely whitespace changes (remove trailing whitespace and use tabs everywhere). The last commit is a by product of my IDE settings and came essentially for free, so I though I include it in this PR, but if you don't like it I can just as easily drop it.

richardeakin:
> Could you elaborate, where the "incomplete type requirements" come into play?

I can't find the original discussion (sorry about that, quite annoying), but basically AFAIK recursive types using STL containers has always been a topic of debate, and in C++11 I think they state that it is 'undefined'. Some background info:

http://www.drdobbs.com/the-standard-librarian-containers-of-inc/184403814
https://www.boost.org/doc/libs/1_48_0/doc/html/container/containers_of_incomplete_types.html

It seemed like clang, when first released, would just forbid an incomplete type at compile time. But I haven't been following clang development lately and the original author has since moved on. If it's supported across the board, I for one wouldn't mind switching back to `std::list`.

Also worth mentioning: we've discussed many times in deprecating `ci::JsonTree` in favor of using jsoncpp or some other C++ json library directly, since the cinder wrapper doesn't offer much extra. Moving it out of the core at some point has also been discussed (in part because it uses boost).

richardeakin:
> Obviously one could work around that (e.g. using a std::vector<std::unique_ptr<JsonTree>>), but I'm not sure if it is worth the effort if you want to remove it anyway.

This is where I landed on it, too. If anyone wants to change it use `std::unique_ptr` (which is what `XmlTree` does, by the way), I think we'd get that in there but it's probably substantial work considering we're planning on deprecate the class.

On the topic of deprecation, we need to come up with some strategy of where that code will live - I think we can already move `JsonTree` out of the core lib but we probably need it within the repo still. Not sure if it qualifies as a cinderblock. Perhaps in `src/deprecated`?

PR #2044, opened 2018-10-21
Optional GStreamer support on MSW.

PetrosKataras:
This PR supersedes #1947 and adds support on MSW for video playback with GStreamer on top of the latest GstPlayer fixes ( #2039 ).

### DIFF
--- a/include/cinder/qtime/QuickTimeGl.h
+++ b/include/cinder/qtime/QuickTimeGl.h
@@ -31,7 +31,7 @@
 	#include "cinder/qtime/QuickTimeGlImplLegacy.h"
 #elif ( ! defined( CINDER_MAC_USE_GSTREAMER ) && defined( CINDER_MAC ) && ( MAC_OS_X_VERSION_MIN_REQUIRED >= 1080 ) ) || defined( CINDER_COCOA_TOUCH )
 	#include "cinder/qtime/QuickTimeGlImplAvf.h"
-#elif defined( CINDER_MAC_USE_GSTREAMER )
+#elif defined( CINDER_MAC_USE_GSTREAMER ) || defined( CINDER_MSW_USE_GSTREAMER )
 	#include "cinder/linux/Movie.h" 
 #elif defined( CINDER_ANDROID )
 	#include "cinder/android/MovieGl.h" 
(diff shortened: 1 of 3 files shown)


richardeakin:
Fired this up this evening and working quite well. I'm up for merging to keep things rolling along, however I wanted to note that I'm seeing some heavy stuttering when I switch the window to full screen (all videos I play). Fullscreen issues are typical of other video backends too (WMF comes to mind), and I'm not sure if it is a new issue (I briefly remember testing fullscreen before but I'm only noticing this now).

In any event @PetrosKataras let me know if you're ready to merge and I'll hit the green button.

PetrosKataras:
Do you mean that once you switch to fullscreen there is from then on continuous heavy stuttering during playback which disappears if you switch back to window mode? I haven't tried switching between fullscreen-window mode on Windows with the player to be honest and I m away from a Windows PC currently but I haven't seen this on neither macOS or Linux. Could this be a Windows/driver/vsync specific issue? 

A couple of things to try:
- Does it stutter if you start the application directly on fullscreen?
- Does it stutter if you disable the frame rate limiter and you let only vertical sync enabled on the application side ( i.e `disableFrameRate` together with `enableVerticalSync` )?
- Does it stutter if you disable altogether vertical sync and the frame rate limiter?
- Does it stutter if you disable the gl path? ( i.e force [this option](https://github.com/cinder/Cinder/blob/master/src/cinder/linux/GstPlayer.cpp#L401) to `false` )

I don't see what else could be different in the fullscreen mode since the decoding and texture generation from the GStreamer side depends on the resolution/frame rate of the video and this stays the same no matter if running on windowed or fullscreen mode thus my thinking regarding a vsync issue. 

Other than that this should be, in principle, ready from my side.

richardeakin:
> Do you mean that once you switch to fullscreen there is from then on continuous heavy stuttering during playback which disappears if you switch back to window mode? 

Hm, this is exactly what was happening when I tested, although I just tried again and can't reproduce. All videos are playing well, from 720p on up to 4K via drag and drop. I'm going to merge and if we see it come up again, will make raise an issue to investigate then. Thanks Petros!

PR #2047, opened 2018-10-29
Disable bitcode generation on iOS

notlion:
XCode 10 enables bitcode generation on iOS targets by default. This causes linking to fail on iOS builds because the Boost prebuilt libraries do not contain bitcode. This commit disables bitcode, which makes things work until Boost is rebuilt with bitcode enabled.


richardeakin:
Anyone know if there is something stopping us now from using the libc++ version of `filesystem`? I was reading this [stackoverflow post](https://stackoverflow.com/questions/42633477/macos-clang-c17-filesystem-header-not-found) and it seems possible now if we link with `-lc++experimental`.

If so, we wouldn't need to link to boost libraries at all.

notlion:
Unfortunately the latest XCode 10.1 still does not have `<filesystem>` or `<experimental/filesystem>` headers, so seems like we still need boost. 😞

notlion:
Yeah I'm not sure why people in the thread you linked were saying filesystem is included in XCode 10. I checked the libc++ headers in both 10.1 and the latest beta and they're not present even in the experimental subdirectory.. I even tried to compile a quick .cpp test file. I mean.. We know that Apple doesn't really prioritize C++ at all 😐 Would be really nice to ditch Boost entirely.

Issue #2052, opened 2018-11-09
About half the samples don't build on Linux


rowet:
A mix of include and linker errors mean that half the samples don't build on Linux. In my case, Fedora 28 with gcc (GCC) 8.2.1.

```
make -k \
CairoBasic \
CameraPersp \
CaptureBasic \
CaptureCube \
Compass \
FontSample \
FrustumCulling \
Kaleidoscope \
LocationManager \
MandelbrotGLSL \
MotionBasic \
perlinTest \
QuaternionAccum \
QuickTimeAdvanced \
QuickTimeAvfWriter \
QuickTimeIteration \
RDiffusion \
Renderer2dBasic \
SerialCommunication \
StereoscopicRendering \
SurfaceBasic \
TextTest \
VoronoiGpu \
Wisteria
```
[errors.txt](https://github.com/cinder/Cinder/files/2567677/errors.txt)

PetrosKataras:
Have a look on #2053. It should fix most of the samples that are supposed to run on Linux. 

The ones that are not included there are because either they use features that have not been implemented yet on the Linux side yet ( e.g `Cairo` support or `Capture` support ) or because they are platform specific examples ( e.g `LocationManager` is a sample supposed to run on mobile platforms ).

rowet:
Here are the remaining errors. If I fix the Cairo includes I get linker errors, as you indicate. Didn't try anything else.

```
$ grep 'fatal error:' /tmp/errors2.txt 
/var/tmp/Cinder/samples/CairoBasic/src/CairoBasicApp.cpp:6:10: fatal error: cinder/cairo/Cairo.h: No such file or directory
/var/tmp/Cinder/samples/Compass/src/CompassApp.cpp:5:10: fatal error: cinder/MotionManager.h: No such file or directory
/var/tmp/Cinder/samples/FontSample/src/FontSampleApp.cpp:2:10: fatal error: cinder/cairo/Cairo.h: No such file or directory
/var/tmp/Cinder/samples/LocationManager/src/LocationManagerApp.cpp:13:10: fatal error: cinder/LocationManager.h: No such file or directory
/var/tmp/Cinder/samples/MotionBasic/src/MotionBasicApp.cpp:5:10: fatal error: cinder/MotionManager.h: No such file or directory
/var/tmp/Cinder/samples/perlinTest/src/perlinTestApp.cpp:2:10: fatal error: cinder/cairo/Cairo.h: No such file or directory
[...]

PR #2054, opened 2018-11-11
Enable ci::Serial on Linux.

PetrosKataras:
This comes cheap since it shares the same unix implementation as macOS. `SerialCommunication` sample builds and runs also on Linux now. Relevant #2052.


richardeakin:
Sweeet.

PR #2059, opened 2018-11-24
Fix xcode build path in OSX instructions

matchingponies:
- cd xcode
+ cd proj/xcode


richardeakin:
Thanks! Miss the old smiley face. :)

PR #2060, opened 2018-11-28
Added videoInput.cpp to msw platform cmake file and fixed header path

Mystfit:
When compiling cinder using CMake on Windows, videoInput.cpp was missing from platform_msw.cmake, leading to unresolved external symbol errors. This commit adds videoInput.cpp back into the project and adds a cinder/video/msw source group to bring it in line with the cinder/audio/msw source group.

Tested using CMake 3.13.0, MSVC 15.8.8, Windows 10 build 1803.


richardeakin:
Thanks for this! I believe it is correct, we need videoInput.cpp for the `ci::CaptureImplDirectShow`. Must have flown under the radar because we hadn't yet tried anything related to video capture on windows via the cmake build system route.

MikeGitb:
Could this please get merged?

andrewfb:
Thanks for this fix.

PR #2061, opened 2018-11-29
add Capture implementation for Linux

whg:
Given `v4l2` is part of the Linux kernel, I thought it would be a good idea to use it to implement video capture on Linux. I didn't need it, but on some distributions you may need to download the `libv4l-dev` package for this to compile.

With view of issue #2052, `CaptureBasic` and `CaptureCube` now build and work on my machine.

Here are a few notes:

* I setup `v4l2` to double buffer the incoming capture frames.
* When `getSurface()` is called a copy of the last written buffer is made. As far as I can see, this is necessary.
* A new `shared_ptr` is created each time `getSurface()` is called. I noticed that the Windows implementation has a pool of `SurfaceRef`s to eliminate this overhead. I chose the simplest solution!

### DIFF
--- a/include/cinder/Capture.h
+++ b/include/cinder/Capture.h
@@ -50,6 +50,10 @@
 	namespace cinder {
 		class CaptureImplJni;
 	}	
+#elif defined( CINDER_LINUX )
+	namespace cinder {
+		class CaptureImplLinux;
+	}
 #endif
 
 #include <map>
@@ -106,7 +110,7 @@ class CI_API Capture {
 	//! Finds the first device whose name contains the string \a nameFragment
 	static DeviceRef findDeviceByNameContains( const std::string &nameFragment );
 
-#if defined( CINDER_COCOA ) || defined( CINDER_ANDROID )
+#if defined( CINDER_COCOA ) || defined( CINDER_ANDROID ) || defined( CINDER_LINUX )
 	typedef std::string DeviceIdentifier;
 #else
 	typedef int DeviceIdentifier;
@@ -148,6 +152,8 @@ class CI_API Capture {
 	CaptureImplDirectShow			*mImpl;
 #elif defined( CINDER_ANDROID )
 	CaptureImplJni					*mImpl;		
+#elif defined( CINDER_LINUX )
+	CaptureImplLinux 				*mImpl;
 #endif
 };
 
--- a/include/cinder/CaptureImplLinux.h
+++ b/include/cinder/CaptureImplLinux.h
@@ -1,0 +1,72 @@
+#pragma once
+
+#include "cinder/Cinder.h"
+#include "cinder/Capture.h"
+
+#include <thread>
+#include <condition_variable>
+
+namespace cinder {
+
+class CaptureImplLinux {
+public:
+	class Device : public Capture::Device {
+	public:
+		Device( const std::string &deviceFile );
+
+		Capture::DeviceIdentifier getUniqueId() const override { return mDeviceFile; }
+
+		bool checkAvailable() const override;
+		bool isConnected() const override { return mIsConnected; }
+
+		void setConnected( bool c=true ) { mIsConnected = c; }
+
+	protected:
+		Capture::DeviceIdentifier mDeviceFile;
+		bool mIsConnected;
+	};
+
+public:
+	CaptureImplLinux( int32_t width, int32_t height, const Capture::DeviceRef device );
+	~CaptureImplLinux();
+
+	void start();
+	void stop();
+
+	bool isCapturing();
+	bool checkNewFrame() const;
+
+	Surface8uRef getSurface() const;
+
+	int32_t getWidth() const { return mWidth; }
+	int32_t getHeight() const { return mHeight; }
+
+	const Capture::DeviceRef getDevice() const { return mDevice; }
+
+	static const std::vector<Capture::DeviceRef>& getDevices( bool forceRefresh = false );
+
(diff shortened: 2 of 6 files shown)


sortofsleepy:
:wave: 

Just tried on a Linux box with Ubuntu 18.04 and I believe I'm running into what Petros is seeing(I'm not totally set up to properly debug just yet). 

The following shows up in the console when I close the window. 

```c++
terminate called without an active exception
Aborted (core dumped)

```

sortofsleepy:
Looks like if I add 
```c++
	mRunning.store( false );
	mThread.join();
```

to the very top of the destructor - this seems to fix things; closing the window doesn't raise the error anymore.

whg:
Hi guys,

Thanks @sortofsleepy for pointing me in the right direction, in my tests I was explicitly calling shutdown(), so I've added that to the destructor.

It also turns out that I wasn't properly shutting down the thread, so I've fixed that too.

andrewfb:
This looks good to me, and I know @sortofsleepy has continued stress-testing it as well. I was going to request a macro like `CINDER_DISABLE_CAPTURE` before I noticed you'd already done it. Thanks for this - I'd be inclined to merge this unless anyone has any objections (I can bring it in line with Cinder style when I do so).

gaborpapp:
Thanks for this. Tested it on Fedora and works well.

3togo:
I got the following error when compiling cinder-capture with captureBasic in linux. 
|error  | virtual void CaptureBasicApp::setup()[39] Failed to init capture , exception type: cinder::CaptureExcInitFail, what: 

Any pointer on how to fix it?

3togo:
I found the bug!. It turns out that the program detected two webcams
/dev/video0 and /dev/video1
but in my notebook, /dev/video0 is NULL
 
if ( !mDevice ) {
		auto &devices = getDevices();
		if ( devices.empty() ) {
            puts("No device found");
			throw CaptureExcInitFail();
		} else {
			mDevice = getDevices()[1];
		}
	}

gaborpapp:
The thing with the two devices shown above is that they are two modes, motion-jpeg and yuv on my webcam. The current implementation only works with the yuv uncompressed stream, which has lower fps. For better frame rate in higher resolutions we would need to implement the mjpg support.

What do you think? 
Anyone else experiencing this?

```
$ v4l2-ctl --list-formats-ext
ioctl: VIDIOC_ENUM_FMT
	Type: Video Capture

	[0]: 'MJPG' (Motion-JPEG, compressed)
		Size: Discrete 1280x720
			Interval: Discrete 0.033s (30.000 fps)
			Interval: Discrete 0.033s (30.000 fps)
		Size: Discrete 160x120
			Interval: Discrete 0.033s (30.000 fps)
		Size: Discrete 176x144
			Interval: Discrete 0.033s (30.000 fps)
		Size: Discrete 320x240
			Interval: Discrete 0.033s (30.000 fps)
		Size: Discrete 352x288
			Interval: Discrete 0.033s (30.000 fps)
		Size: Discrete 640x480
			Interval: Discrete 0.033s (30.000 fps)
		Size: Discrete 1280x720
			Interval: Discrete 0.033s (30.000 fps)
			Interval: Discrete 0.033s (30.000 fps)
	[1]: 'YUYV' (YUYV 4:2:2)
		Size: Discrete 1280x720
			Interval: Discrete 0.100s (10.000 fps)
			Interval: Discrete 0.100s (10.000 fps)
		Size: Discrete 160x120
			Interval: Discrete 0.033s (30.000 fps)
		Size: Discrete 176x144
			Interval: Discrete 0.033s (30.000 fps)
		Size: Discrete 320x240
			Interval: Discrete 0.033s (30.000 fps)
		Size: Discrete 352x288
			Interval: Discrete 0.033s (30.000 fps)
		Size: Discrete 640x480
			Interval: Discrete 0.033s (30.000 fps)
		Size: Discrete 1280x720
			Interval: Discrete 0.100s (10.000 fps)
			Interval: Discrete 0.100s (10.000 fps)
```

richardeakin:
Just pushed this along with the two merge conflicts fixed to the [CaptureLinux](https://github.com/cinder/Cinder/tree/CaptureLinux) branch. It is essentially ready to merge to master, though i wanted to ask a couple questions around how this is built and configured:

- There is a `wget` line added to the [trusty-install-deps.sh script](https://github.com/cinder/Cinder/blob/CaptureLinux/tools/linux/scripts/trusty-install-deps.sh#L30-L35), is this preferred over `sudo apt-get install libpmg123-dev` ? The latter is what we have in the [linux ubuntu install doc](https://libcinder.org/docs/guides/linux-notes/ubuntu.html) ([source](https://github.com/cinder/Cinder/blob/2a53ef1c36ead4c664ec74084c924d62b02ac834/docs/htmlsrc/guides/linux-notes/ubuntu.html#L44-L50))
- Is there a FindPackage cmake script handy (and liberally licensed) for V4L2? I noticed that the [current cmake line](https://github.com/cinder/Cinder/blob/2a53ef1c36ead4c664ec74084c924d62b02ac834/proj/cmake/platform_linux.cmake#L57) adds a command to link to `libv4l2.so`, but doesn't check if it is available first.
  - I think ideally we could detect if V4L2 is usable and if so, link in the Capture impl. If it isn't, skip CaptureImplLinux.cpp and add a #define for `CINDER_DISABLE_CAPTURE`
- Do we need to update the docs at all after this? Okay to do it in a separate PR just want to make a note if things have changed there.

PR #2067, opened 2018-12-13
Add Emscripten documentation

sortofsleepy:
Add docs/guide for Emscripten usage . 

Could probably use a once-over from someone who's actually a copywriter but should be pretty complete. 

Note that there two additional documents over a potential path for a new Emscripten specific audio API as well as a starting point for how to use `emscripten::val` object but not linking to those for now.


sortofsleepy:
gonna close this and just commit the docs in the new PR

PR #2068, opened 2018-12-13
Add Emscripten tooling

sortofsleepy:
* Adds tooling for building Cinder against Emscripten. Mostly the same as Hai's stuff but the script has been adjusted to support Windows. 
* There's also a python script to generate build folders for all the samples 
* adds `ci_emscripten_app` function


PetrosKataras:
Personally, I am all in for reviving this work and bringing it into the core - I had also been looking at it a couple of weeks ago. Could, potentially, provide for a nice workflow to test/sketch/combine javascript ML/DNN developments like [ml5](https://ml5js.org/) and [tensorflow.js](https://js.tensorflow.org/) with some Cinder magic. 

That said I see a few issues with this PR mostly related to the way that the CMake configuration has been set up. It seems that there are a few residual things from the early days of adding CMake support to Cinder that have been refactored/cleaned up and improved since then. We should bring this inline with the current status to avoid duplicates and keep things consistent. A few notes/suggestions:

-  I think we should not add a separate `emscripten` directory inside Cinder's root directory but instead work from `proj/cmake` and add there a `platform_emscripten.cmake` that could contain all the specifics of the emscripten configuration. It would basically be the place where parts of `emscripten/CMakeLists.txt` and `emscripten/Cinder.cmake` exist.

- `emscripten/CMakeLists.txt` contains references to source files that are already included in [libcinder_sources_files.cmake](https://github.com/cinder/Cinder/blob/master/proj/cmake/libcinder_source_files.cmake). By integrating with the existing CMake configuration we can avoid this duplication ( it happens on other parts also ). Everything else can go in `platform_emscripten.cmake`

- The emscripten path could be chosen by adding an `emscripten` [CINDER_TARGET](https://github.com/cinder/Cinder/blob/master/proj/cmake/configure.cmake#L35) that would also set the appropriate [CINDER_TARGET_GL](https://github.com/cinder/Cinder/blob/master/proj/cmake/configure.cmake#L64) flags ( i.e choose the  appropriate GL_ES path ). Then we can add specific source files and flags for the target based on these variables being set.

- Remove `cibuild(er)`. A residual thing from the very early days that I believe its also passing some outdated flags. Ideally, we should just have the classic cmake workflow ( `cmake .. -DSOME_VAR` )

- It seems that there are source files that are referenced from the CMake configuration but are not present in this PR. e.g All files under `cinder/emscripten/app/` and `cinder/emscripten`. 

@sortofsleepy I could help with streamlining this work with the rest of the CMake stuff and also test it on macOS / Linux. Cannot modify in place so we could work from your fork/branch and take it from there.

sortofsleepy:
Hey @PetrosKataras ! 
Thanks for all the suggestions! These all make sense and I don't have any issues with implementing these changes. When I was working on getting Emscripten up and running it was more about just getting things working and figuring out stuff, though I admit I really should try to get into the habit of keeping things more organized and cleaner as I go along. 

My goal though first is to get things up and running - to your last point, this PR wasn't meant to be feature complete; I'm porting over a working version of all the code that's currently sitting in a private fork which is why you are seeing missing files.  

Once I get that done - hopefully today my time barring any unexpected bugs like one I just worked through, by all means, it'd be great if you had the time to test things out and see if there's anything else that sticks out to you that could use some improvement. 

I'll comment again once that's done and you can start messing around with the code

sortofsleepy:
@PetrosKataras ok - all the core stuff should be ported over and you should be able to get things up and going if you wanna start testing.
 
Instructions to build Cinder are pretty much the same as [Hai's repo](https://github.com/chaoticbob/Cinder-Emscripten) with the exception that for building samples, you run a python script in `cinder/emscripten/sample_generation`. From there it's like a normal CMake project. 

I have some samples in the samples folder if you'd rather just grab a CMakeLists file and start from scratch.

sortofsleepy:
@richardeakin ah, yes this is a bit older and can be closed actually. I got distracted with another bug yesterday and it didn't occur to me to just close this one ha. Closing now. 

Also, yea it'd be great if you could take a look at the audio stuff - audio processing isn't exactly my strong suit; a lot of it was me just mostly looking at how the Linux support was added and trying different things out. 

I should note that the current implementation doesn't technically have support for native web audio stuff just yet, I just never managed to get around to finishing things due to being distracted with other projects in addition to trying to get this out. Also from what I can tell, grabbing raw data from native web audio nodes is a gonna be a bit of a pain.

I also started putting together an idea for a separate, more WebAudio focused API under `<repo>/_samples/_emscripten/_tests/AudioAPI` if you wanna take a look. At some point the current implementation will become outdated so I thought it'd be best to start thinking about other ways to implement things.

PR #2069, opened 2018-12-14
Emscripten support! 

sortofsleepy:
Opening last PR to finally get Emscripten working!

Should be considered alpha state - still needs to be tested a little more since things were hand-imported for simplicity. If you'd like to help test you can of course pull this PR or my branch of Cinder. Generate and read the docs for the full details, but in short, with the exception of sample building, you can follow ChaoticBob's [old instructions](https://github.com/chaoticbob/Cinder-Emscripten) and from there, it's like a normal CMake based desktop project.  See `<repo>/samples/_emscripten` for sample CMake files you can use. 

Note that audio stuff is still somewhat in flux.

Also should note that most of the core emscripten stuff is header only - will move some stuff to implementation files at a later point.

Any changes / fixes going forward will go to this PR


richardeakin:
Going to respond to the comment [here](https://github.com/cinder/Cinder/pull/2068#issuecomment-447560003) about the audio side of things - you mention there that there is 'no support for native web audio stuff just yet', although I'm trying to see what in https://github.com/cinder/Cinder/commit/a20b341d6edb5b990686149d3f3d4ce4a43727b7 is currently being used. Could you break this down, how audio is implemented and what classes are being used and what aren't? I think this would be valuable as comments in the sources files themselves, too, so that people who may wish to help out on this part can have an idea where to pick up from (as for myself, I can't guarantee and specific time to it at the moment as I'm involved various projects, but I'm always happy to try and help with specifics if someone has questions or is looking for suggestions on how to proceed). Specifically, if something isn't finished I think it would be nice to indicate that at the top of the file with a big "NOT WORKING" flag, or consider moving it out of the repo for now. I imagine many things like hooking into native web audio units instead of using cinder's cpp ones can be done from outside of the repo, at least when testing the implementation.

For a couple particulars

- what are the intentions for the code in `include/cinder/audio/emscripten/Voice.h`? The core `ci::audio::Voice` class is meant to be an easy, cross platform solution for simple audio needs like sound file playback, with the intention of not having native specific code.
- `include/cinder/audio/emscripten/simpleaudio.h`: I take it this is a more efficient way to load files over `audio::SourceFile`? Or can an implementation of `audio::SourceFile` be added? If this is not in the cards for the moment, can this be kept in an example or in the tests folder? I'll also note that there's a big ugly `using namespace emscripten` in that header, which is C++ no-no, so let's remove that in any event. Ideally I would suggest placing as much platform-specific code as possible in .cpp files rather than headers, but if this is just a quick test then maybe it should be moved out of the main cinder/include path.

A note on style: While so far I've really only looked at the audio side of things, I'm seeing numerous pieces that aren't following the [styleguide](https://github.com/cinder/Cinder/blob/master/CONTRIBUTING.md#style-guide). I can point out some specifics in a separate feedback comment, although I'll suggest that it might be easier to get something like clang-format setup ([see here](https://gist.github.com/notlion/71fea3fb3a9eb7c54761) for example) and run it on any newly introduced .h and .cpp files from this PR. Though it's worth noting that not all code in the cinder codebase has yet been updated to this styleguide, so best to keep it specific to this PR. We can update those other files too, but this PR is large enough as is so best to keep pure style changes separate where possible.

As a suggestion for organizing future pull requests, you usually should submit them some named branch (for example, maybe 'emscripten' in this case) instead of directly from your master branch. This allows you to keep your own master branch updated with the main repo, and that name will also end up in the git history once it is finally merged.

I'll also second @PetrosKataras [suggestions here](https://github.com/cinder/Cinder/pull/2068#issuecomment-447259336) for keeping the cmake organization in line with the big refactor we did when adding linux and android a while back, basically agreeing with everything he wrote there. Work like this is obvious substantial (and thank you again for it!), but so is maintaining many platforms, so we want to try to keep things as close to congruent with similar platforms as possible.

That's it for the moment, particularly on high level feedback and some initial audio thoughts.

PetrosKataras:
Just to note that the work for streamlining the CMake configuration with the rest of the platforms is happening here https://github.com/PetrosKataras/Cinder/tree/emscripten if anyone is interested. We are in the process of smoothing out the details but the bulk of work is there already.

sortofsleepy:
@richardeakin First off thanks for taking a look and providing detailed feedback!

To briefly describe things(I can go into more detail later) 

The main goal when starting work on this originally was to replicate the desktop api as much as possible. While the underlying implementation is certainly very different due to the C++/JS interop needed, the basic principles ought to be the same. Like I was saying in #2068, I was using the desktop apis as a reference point when working on this. 

* In terms of what is/isn’t in use, basically everything except `NativeWebAudio`, `simpleaudio` and `WebAudioNode` ought to be in use(I’ll actually just remove WebAudioNode in a sec since it’s not needed and I don’t remember why I added it) I’ll be sure to add notes later today but I think it probably makes more sense to add a general readme to the folder perhaps? I don’t feel like I have a firm grasp of how the things in the desktop api are connected just yet, as well as things having not been put through a thorough testing. I can at least add comments to the native stuff specifically though. 


* The primary difference in between `Voice` and `simpleaudio` is the ability to extract the underlying audio information. `simpleaudio` is intended to be an easy way to setup a playback device; since I knew it wouldn’t take very long to come up with a working sample, I figured it was good to add for those that may potentially just need audio playback for their project.

* As far as loading efficiency - I don’t know for sure but I would imagine that `simpleaudio` technically has a very tiny edge over WebAssembly/asm,js since it leverages the browser’s built-in <audio> tag and subsequently built- in methods for loading. I would say that the main advantage is more in it’s simplicity - since it’s just utilizing an `<audio>` tag, you don’t need to do any complicated setup or any real extra work to, for example, pull an audio file off of an external server. Another possible advantage is that it also makes it simpler to support fallback files, in case a browser doesn’t support a certain file format. Of course, the time-saving argument is moot if you have to do any processing on the audio but I thought it’d be a nice to have since it was pretty simple to lay the groundwork. `audio::SourceFile` could technically be added but I don’t know that it makes sense to given the intention of `simpleaudio` being playback-only. I'm fine with moving it if that's what's wanted. 

* Will keep in mind of not setting namespaces up in headers, a lot of work was done in header files so I didn’t have to constantly re-compile all of Cinder for every change and I didn’t really wanna type out the namespace every time(granted in hindsight I really should have just kept files local to whatever I was doing). I do plan on moving things to implementation files once the bigger issues have been ironed out a bit more.  

* Thanks for the pointers on style - I wasn’t aware there was something written down, I’ll be sure to make a pass over things today.

Thanks again!

richardeakin:
> In terms of what is/isn’t in use, basically everything except NativeWebAudio, simpleaudio and WebAudioNode ought to be in use(I’ll actually just remove WebAudioNode in a sec since it’s not needed and I don’t remember why I added it) I’ll be sure to add notes later today but I think it probably makes more sense to add a general readme to the folder perhaps?

I think a readme that summarizes your WIP for others would be great, or perhaps you could also do this on a wiki page. In any event, I think it will help for all to know what the next steps would be in the emscripten ci::audio stuff is a work in progress.

**re `simpleaudio`**: Ah ok, I understand now it is a wrapper around the HTML `<audio>` tag. This is pretty unrelated to the `ci::audio` namespace, being specific to web (as in, won't work anywhere else). I think it appears to be useful, but as it is not cross platform I think it should probably live in either the `cinder/Emscripten` folder, or maybe even in the tests / samples.  I think a note at the top describing that it wraps HTML `<audio>`, and is separate to the native C++ cinder::audio stuff, would let people know that this is a quick and easy way to for simple audio playback on the web platform, but it won't allow you to tie into the `ci::audio` api.

If we do want to use soundfiles within `ci::audio`, which I think would be the most common case when porting a native cinder app to the web that contains audio (most commonly with `audio::BufferPlayerNode` or `audio::FilePlayerNode`, then we will need an `audio::emscripten::SourceFileWeb` implementation of some sort, which loads a file into raw PCM samples one way or another. I don't imagine this will be too inefficient even with web assembly, keeping in mind that audio file loading to buffer can almost always be done in background threads.

I did just notice `WebAudioLoader` -was this an attempt to load audio files to native PCM? AFAICT it isn't fully implemented, but I haven't looked at the js side of that. Also happy to discuss this offline if you want to keep working at it and would like me to look at something closer.

I'm still unsure why `Voice` has been reimplemented. cinder/audio/emscripten/Voice.h looks almost identical to cinder/audio/Voice.h, I think there is one extra `isDoneLoading()` method? For the emscripten/Voice.cpp, all I see that has been newly introduced is [this chunk](https://github.com/cinder/Cinder/pull/2069/files#diff-50b395b899b5afb1e06653badc6d87deR296). Is this working? In any event, I think we want to avoid this extra near-identical class, and think of a solution that uses the existing cross-platform version.

sortofsleepy:
> re simpleaudio: Ah ok, I understand now it is a wrapper around the HTML <audio> tag. This is pretty unrelated to the ci::audio namespace, being specific to web (as in, won't work anywhere else). I think it appears to be useful, but as it is not cross platform I think it should probably live in either the cinder/Emscripten folder, or maybe even in the tests / samples. I think a note at the top describing that it wraps HTML <audio>, and is separate to the native C++ cinder::audio stuff, would let people know that this is a quick and easy way to for simple audio playback on the web platform, but it won't allow you to tie into the ci::audio api

👍 Sounds good to me. 

> If we do want to use soundfiles within ci::audio, which I think would be the most common case when porting a native cinder app to the web that contains audio (most commonly with audio::BufferPlayerNode or audio::FilePlayerNode, then we will need an audio::emscripten::SourceFileWeb implementation of some sort, which loads a file into raw PCM samples one way or another. I don't imagine this will be too inefficient even with web assembly, keeping in mind that audio file loading to buffer can almost always be done in background threads.

> I did just notice WebAudioLoader -was this an attempt to load audio files to native PCM? AFAICT it isn't fully implemented, but I haven't looked at the js side of that. Also happy to discuss this offline if you want to keep working at it and would like me to look at something closer.

I can look into a `audio::emscripten::SourceFileWeb` at some point( or at least make a note about it ). I'm not terribly concerned about efficiency either - like I said, if there is a difference between using WebAssembly vs a tag, I imagine the difference is very very small.

In regards to threading though, I guess I do have a question there - do you think it's a requirement in this case?  It's certainly possible to load an audio file in a threaded manner, but any decoding would still have to happen on the main thread, not to mention, the setup to even do that might end up being rather verbose even if we were to hide some of the ugly bits behind the scenes. In any case I'll certainly take a look and give it some thought. 

 As for WebAudioLoader, it does work - unfortunately you do need some JS to go along with the C++ in order to pull any useful information about an audio file which is probably why it looks incomplete. 

> I'm still unsure why Voice has been reimplemented. cinder/audio/emscripten/Voice.h looks almost identical to cinder/audio/Voice.h, I think there is one extra isDoneLoading() method? For the emscripten/Voice.cpp, all I see that has been newly introduced is this chunk. Is this working? In any event, I think we want to avoid this extra near-identical class, and think of a solution that uses the existing cross-platform version.

Many apologies for the confusion! Thinking back to when I was working on it, I think the original intention in my head was to augment the current version of `cinder/audio/Voice` with `isDoneLoading` + the [accompanying chunk starting here](https://github.com/sortofsleepy/Cinder/blob/master/samples/_emscripten/_audio/EmscriptenAudio/Voice.cpp#L296). I kept separate copy to test things out without making any huge changes to the original and it just simply slipped my mind to make those final changes. If it the changes look ok then we can definitely just remove my copy of `Voice`

richardeakin:
> In regards to threading though, I guess I do have a question there - do you think it's a requirement in this case? It's certainly possible to load an audio file in a threaded manner, but any decoding would still have to happen on the main thread, not to mention, the setup to even do that might end up being rather verbose even if we were to hide some of the ugly bits behind the scenes. In any case I'll certainly take a look and give it some thought.

It isn't a requirement and usually not necessary for very small files like sound effects, but if you try loading anything over a few seconds it will take long enough to freeze the main thread, which you want to avoid. Where does the requirement for decoding on the main thread come from?

I also noticed there is a new callback to the emscripten version of `VoiceSamplePlayer` constructor - where is this called from? Is it not possible to keep that internal to the Voice.cpp? Again, trying to think of how this code can be cross-platform, we want to avoid having to do something special for web if we can.

> Many apologies for the confusion! Thinking back to when I was working on it, I think the original intention in my head was to augment the current version of cinder/audio/Voice with isDoneLoading + the accompanying chunk starting here. I kept separate copy to test things out without making any huge changes to the original and it just simply slipped my mind to make those final changes. If it the changes look ok then we can definitely just remove my copy of Voice

We can certainly modify the existing Voice.cpp to make this work, though as I wrote above, we want to try to keep any new additions to the interface cross-platform. Let's discuss offline why those methods were introduced and see if we can make that happen.

samples/_emscripten/_audio/EmscriptenAudio/Voice.cpp
```diff
@@ -0,0 +1,345 @@
...
+			BufferRef buffer = MixerImpl::get()->loadBuffer( sf );
+			mNode = Context::master()->makeNode( new BufferPlayerNode( buffer ) );
+		} 
+		else {
+			mNode = Context::master()->makeNode( new FilePlayerNode( sf ) );
+		}
+	#else
+
+
+		auto loader = (WebAudioLoader*)sourceFile.get();
```
richardeakin: Some feedback on this section of the code: if `WebAudioLoader` is a subclass of `ci::audio::SourceFile`, maybe it should be the thing we were discussing and calling `SourceFileWeb` (or maybe `SourceFileWebAudio` is a better fit, I don't know).

samples/_emscripten/_audio/EmscriptenAudio/Voice.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+		loader->init( [=]()->void
```
richardeakin: You usually want to avoid copying a local variables into a lambda (the '=' in `[=]`), instead it's best to pass in the `this` pointer and anything else explicitly. Also, I'm not sure if the `WebAudioLoader *loader` will be valid if this callback is async. By the way, is it (`WebAudioLoader::init()`) async? If it isn't what is the intention of using a callback?

samples/_emscripten/_audio/EmscriptenAudio/Voice.cpp
```diff
@@ -0,0 +1,345 @@
...
+		{
+			BufferRef buffer = MixerImpl::get()->loadBuffer( sourceFile );
+			BufferRef audio = loader->getAudioData();
+			mNode = Context::master()->makeNode( new BufferPlayerNode( audio ) );
+			mDoneParsing = true;
+
+			VoiceSamplePlayerNodeRef result( this );
```
richardeakin: This is incorrect usage of `shared_ptr`s here. You can't safely create a shared_ptr from the shared object's constructor (even if you use `shared_from_this()`), you should instead use the `create()` methods ([here](https://github.com/cinder/Cinder/blob/master/src/cinder/audio/Voice.cpp#L197-L203)) to construct the `shared_ptr` and add it to the mixer.

include/cinder/emscripten/EmscriptenVideo.h
```diff
@@ -0,0 +1,117 @@
...
+
+#include <emscripten.h>
+#include <emscripten/bind.h>
+#include <emscripten/val.h>
+#include "cinder/Log.h"
+#include "cinder/gl/Texture.h"
+#include "cinder/gl/gl.h"
+
+using namespace ci;
+using namespace emscripten;
```
richardeakin: These `using namespace` statements should be moved to EmscriptenVideo.cpp.

samples/_emscripten/_audio/EmscriptenAudio/Voice.cpp
sortofsleepy: Looking more into this you're right of course, it wasn't quite clear when first doing this how `ci::audio::Source` was working but taking a second look things seemed to make more sense. 
I've added a `SourceFileWebAudio` class to try and better match the other platforms.

samples/_emscripten/_audio/EmscriptenAudio/Voice.cpp
sortofsleepy: 👍 Did not know that about lambdas, changed things up a bit.

`WebAudioLoader::init()` is indeed async. The callback you pass to it is called once the audio is loaded on the Web side of things. The path essentially is 

* load audio on JS side
* local lambda in `WebAudioLoader::init()` is then called to parse the information into a local `BufferRef` 
* finally lambda passed to `WebAudioLoader::init()` (note - now `SourceFileWebAudio`) is called to finish loading audio.

`loader` does still work from what I can see

include/cinder/gl/TextureFont.h
```diff
@@ -145,7 +145,7 @@ class CI_API TextureFont {
 	bool	isPremultiplied() const { return mFormat.getPremultiply(); }
 
 	//! Returns the default set of characters for a TextureFont, suitable for most English text, including some common ligatures and accented vowels.
-	//! \c "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz1234567890().?!,:;'\"&*=+-/\\@#_[]<>%^llflfiphrids����"
+	//! \c "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz1234567890().?!,:;'\"&*=+-/\\@#_[]<>%^llflfiphrids����"
```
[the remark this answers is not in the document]
sortofsleepy: hmm no it wasn't intentional - not actually sure how that might have happened since aside from adding a `CINDER_EMSCRIPTEN` check I don't believe I touched the file. I can re-do that change, looks like I only had to change one spot.

include/cinder/emscripten/EmscriptenVideo.h
sortofsleepy: Has been updated to remove namspace statements out of the header

sortofsleepy:
> It isn't a requirement and usually not necessary for very small files like sound effects, but if you try loading anything over a few seconds it will take long enough to freeze the main thread, which you want to avoid. Where does the requirement for decoding on the main thread come from?

Hmm requirement was perhaps too strong a word - the main reason is more out of the concern of organization and structure of applications. 

While you can certainly load audio on a "thread" aka WebWorkers - you can't utilize anything that's associated with the DOM, this unfortunately seems to include the `AudioContext` object which is necessary in order to extract any useful information about an audio file ie number of channels, channel data, etc. You could conceivably grab all the information you want on the main thread and pass it to a worker and pass things back, but I can't imagine that set up not being bit of a pain to orchestrate. 

> I also noticed there is a new callback to the emscripten version of VoiceSamplePlayer constructor - where is this called from? Is it not possible to keep that internal to the Voice.cpp? Again, trying to think of how this code can be cross-platform, we want to avoid having to do something special for web if we can.

I think it is within Voice.cpp? If I'm remembering correctly at the time - I was trying to keep things as similar as possible to the main desktop api and as far as I can tell the code should basically be working the same aside from the extra lambda that can get passed in.

samples/_emscripten/_audio/EmscriptenAudio/Voice.cpp
sortofsleepy: hmmm - I get what you're saying but I don't know if there's a good way to avoid this; I spent some time this afternoon trying to re-write things to build a `shared_ptr` in a more normal fashion, for example 

```c++
VoiceSamplePlayerNodeRef result = VoiceSamplePlayerNodeRef(new VoiceSampleNode) 
``` 

and every time I try to add the `VoiceSamplePlayerNodeRef` to the mixer, I get a memory error saying I accessed something out of bounds(unfortunately the error isn't too helpful beyond that).

I will take a look again later.

vinjn:
What's the status of this PR? Really like to use it :)

sortofsleepy:
@vinjn - Really up to Andrew and Rich of course as to when this gets merged in; from my side the big lingering tasks are just audio things that still ought to be worked out; I haven't really had too much time or energy to sit down and properly think things through between looking for work and getting distracted with other ideas haha.

I'm thinking though I'm going try to continue that next week - in the mean time though, assuming you don't need audio support my Cinder fork should work if you'd like to start exploring

vinjn:
I don't need Audio :)
Can we move Audio parts to another PR and merge this one? 
@richardeakin  @andrewfb

richardeakin:
I think the biggest blockers for merging this are

- a) using the same cmake build setup as the other platforms
- b) more eyes looking at the app impl code. I really only had a moment to review audio stuff.

I think the audio stuff can be disabled by cmake config like on other platforms, and then also remove whatever things were experimental (or move them to tests).

In the meantime @vinjn, I would suggest keeping master branch merged into the one on this PR so that it remains up to speed, and testing from there.

sortofsleepy:
@richardeakin @vinjn 
- In regards to CMake stuff - I merged in changes Petros had made so I'm reasonably certain that should be up to snuff unless I accidentally missed a commit somewhere.  
- If everyone is in agreement, I'm fine with taking out / moving the audio stuff for now. I can do that sometime next week so we can get the more important bits finally merged in. Finally had the time to begin refactoring the code I started writing trying to implement the [Audio Worklet](https://developers.google.com/web/updates/2017/12/audio-worklet) model anyways.

dvj:
Was there any resolution to this? I think it would be great to set aside the audio and get the rest of this merged!

sortofsleepy:
I had never heard back about whether or not I should remove the audio stuff and also got busy eventually and haven't had a moment to really focus on Cinder stuff, day job unfortunately does not involve stacks like Cinder, etc and is much more boring ha 😢  . Apologies all around.

Last I recall, trying to integrate the `AudioWorklet` model was proving problematic as it does require JS to be written and does not seem possible to bridge the api with C++ in a clean manner(at least not that I've figured out yet); that said I am still fine with removing support for the time being.  

@dvj - I do have a fork you're welcome to try in the meantime.

vinjn:
@sortasleepy I can help to clean up the code, remove the audio part out, and send out a new PR, will that work for you?

sortofsleepy:
@vinjn - Coincidentally I should be finishing up a good chunk of work by the weekend and will have some spare time; I was considering taking this back up and stripping out the audio portion and cleaning some stuff up. 

That said,  if you have the time and are willing I wouldn't mind the help as I still have other stuff I ought to focus on. Please do feel free to fix things up and submit a new PR.

vinjn:
O great, I will waiting for your PR in stead.

sortofsleepy:
Gonna close this for now as this has gotten a bit long and cause there were some additional changes needed to accommodate the change to GLAD and possibly more needed due to removing Boost and other changes with 0.9.3. 

Will submit a new PR some time this week.

Issue #2064, opened 2018-12-11
Build fails due to glxext.h on Linux


whg:
Cinder fails to build due to some conflicting types and I get the following:
```
In file included from /home/whg/Code/Cinder/include/glload/glx_all.h:6,
                 from /home/whg/Code/Cinder/src/glload/glx_load.c:4:
/home/whg/Code/Cinder/include/glload/_int_glx_type.h:96:3: error: conflicting types for ‘GLXHyperpipeNetworkSGIX’
 } GLXHyperpipeNetworkSGIX;
   ^~~~~~~~~~~~~~~~~~~~~~~
```
The culprits are:
* `GLXHyperpipeNetworkSGIX`
* `GLXHyperpipeConfigSGIX`
* `GLXPipeRect`
* `GLXPipeRectLimits`
* `PFNGLXGETCURRENTDISPLAYEXTPROC`
* `PFNGLXGETCURRENTREADDRAWABLESGIPROC`

I'm on Linux and using Mesa (version 18.2.6).

I've made a temporary fix by putting some include guards in the glx source:
[https://github.com/whg/Cinder/commit/acdf31312e88332e2f22e2a7609597d31d6a2f8f](https://github.com/whg/Cinder/commit/acdf31312e88332e2f22e2a7609597d31d6a2f8f)

But it's probably not the way to go! I checked out the recent commits on Mesa but couldn't find anything suspect, so I'm slightly confused about why it's happening.

gaborpapp:
Also reported in issue #2062 
Are you also on Arch?

whg:
No, I'm on Manjaro, but will be getting the same packages, so no doubt it's the same problem.

After a bit of testing I found that it was the upgrade from Mesa 18.2.4 to 18.2.6 that broke the build. After checking the diffs I found the offending line. In a [recent commit](https://gitlab.freedesktop.org/mesa/mesa/commit/f7d42ee7d319256608ad60778f6787c140badada#31ac431cdd8c73c76ffa76809620c3e2c0ada784_2_1), a define `/usr/include/GL/glxext.h` changed from: 

`#define __glxext_h_` 
to 
`#define __glx_glxext_h_`

This changes breaks `glload` because it tests agains `__glxext_h_`.

@gaborpapp it's interesting to hear to you're using a recent Mesa and it compiles for you. What's defined in your `/usr/include/GL/glxext.h`?

[...]

PR #2070, opened 2018-12-16
fix compile for Mesa 18.2.6 on some Linux

whg:
As discussed in #2064, this add compatibility for `glload` with newer version of Mesa.

### DIFF
--- a/include/glload/_int_glx_type.h
+++ b/include/glload/_int_glx_type.h
@@ -1,10 +1,11 @@
 #ifndef GLXWIN_GEN_TYPE_H
 #define GLXWIN_GEN_TYPE_H
-#ifdef __glxext_h_
+#if defined( __glxext_h_) || defined(__glx_glxext_h_)
 #error Attempt to include glx_exts after including glxext.h
 #endif
 
 #define __glxext_h_
+#define __glx_glxext_h_
 
 #include <X11/Xlib.h>
 #include <X11/Xutil.h>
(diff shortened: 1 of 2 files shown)


gaborpapp:
I also have this issue on a fresh install of Fedora 29 with an eGPU, although previously I had no problems with mesa 18.2.6-2 on another computer, now the build fails with mesa 18.2.6-3, even with this fix.

I have `#define __glx_glxext_h_ 1` in `/usr/include/GL/glxext.h`.

I'm investigating this further.

whg:
Not sure how I managed to compile on my end, but I've added the changes to `_int_glx_type.hpp` too. Thanks!

rydrman:
This patch worked for me on Fedora 29 as it is now, no other changes needed.

For other people's reference, I was getting the following error, but now everything builds great! :+1: 
> error: conflicting types for ‘GLXHyperpipeNetworkSGIX’

chamric:
This patch worked for me on ubuntu 18

eliaskousk:
I was trying to build the unofficial OpenGL SDK (which contains glload) in order to follow a tutorial that uses it (and thus glload) and this solved the same glload build issue on my Kubuntu 18.04.

[Unofficial OpneGL](https://sourceforge.net/projects/glsdk/)
[OpenGL Tutorial](https://paroj.github.io/gltut/)

richardeakin:
Thanks and apologies for the delay..

Issue #2057, opened 2018-11-22
geom::Torus radii naming clarification


gaborpapp:
I think that `geom::Torus` radii naming is a bit confusing and should be documented.
Usually the distance between the axis and the center of the rotated circle is known as major radius (`R`), whereas the circle radius is called minor radius (`r`). While Cinder's major radius seems to be `R + r`, and minor radius is `R`.

paulhoux:
Good point, Gabor. Changing the implementation would break existing code, but adding some documentation will do no harm. Do you think you could make a PR for it?

gaborpapp:
Thanks Paul. I would clarify it something like this: `major is the distance from the axis to the farthest point of the torus. minor is the distance between the axis and the center of the rotated circle`.

Looking at the code of `ratio()` (https://github.com/cinder/Cinder/blob/master/include/cinder/GeomIo.h#L531) also uses the naming, but the code works differently.

The documentation says it `Specifies the major and minor radius as a ratio (minor : major). Resulting torus will fit unit cube`, but the code is `mRadiusMajor = 1; mRadiusMinor = 1 - ratio;` so `radius( 1.0, 0.75 )` generates the same torus as `aspect( 0.25 )`, and not `aspect( 0.75 )` as the comment suggests.

Here I think the code should be fixed to work according the documentation, although maybe it would be better to stick to the standard naming and break existing code to avoid further confusion. What do you think?

richardeakin:
I think our policy has been that if the code will break at compile time, it's okay to make the change to improve an API. But we should avoid silent breaking changes, which I think this one would fall under. May I suggest adding a new method called `radiusCentered( major, minor )` for the 'usual' version, and then improving the docs on the existing methods?

I will say that I recall fiddling with the Torus methods quite a bit to understand what the params are controlling major / minor axes, so improved docs would be great.

PR #2072, opened 2018-12-18
geom::Torus radii docs improvement and radiusCentered method

gaborpapp:
Improving the docs on major, minor radius as discussed in #2057.
Fixed `ratio` method.
Adding `radiusCentered` method that follows a more usual naming for torus creation when a circle of radius `r` is rotated around an axis. I used `r` and `R` for minor and major radius in this method, to make it more clear that those variables are used for other distances by the class.


include/cinder/GeomIo.h
```diff
@@ -528,9 +528,11 @@ class CI_API Torus : public Source {
 	//! Allows you to twist the torus along the ring. The \a offset is in radians.
 	Torus&	twist( unsigned twist, float offset ) { mTwist = twist; mTwistOffset = offset; return *this; }
 	//! Specifies the major and minor radius as a ratio (minor : major). Resulting torus will fit unit cube.
-	Torus&	ratio( float ratio ) { ratio = math<float>::clamp( ratio ); mRadiusMajor = 1; mRadiusMinor = 1 - ratio; return *this; }
-	//! Specifies the major and minor radius separately.
+	Torus&	ratio( float ratio ) { ratio = math<float>::clamp( ratio ); mRadiusMajor = 1; mRadiusMinor = ratio; return *this; }
```
richardeakin: The docs are a nice improvement IMO. though isn't it a breaking change? Though I'm not against it in this case, I don't think the `ratio()` method is used as much. @paulhoux what do you think?

gaborpapp: It is a breaking change indeed, but I think `ratio()` was not working according to its documentation. if `ratio = minor / major` and `major = 1` then `minor` should be equal to `ratio` and not `1 - ratio`.

PR #2076, opened 2019-01-04
Update some of the OpenGL demos so things can be built with emscripten

sortofsleepy:
In preparation for Emscripten support to be merged in, there is a demo site prepared. This updates some of the OpenGL samples that are shown on that site so that once Emscripten stuff is merged in, the samples will be able to be built. 

This should should be against the current master state at commit 642dbbd so it should be ok to merge in now if desired; I've tested on Windows and demos all still work.

No worries if this ought to wait, figured might as well prep something. 

One note - for the `NormalMapping` sample, for whatever reason, when loading the mesh as an asset, the wrong version number is somehow read and as a result the build will fail. This includes a copy so it can be pulled as a resource so that source for the sample doesn't have to be re-written too much; if it's cool to do a bit more re-writing so that everything is loaded as a resource, I can do that.


sortofsleepy:
This should be considered out of date and you should refer to the code in PR #2172. There have been changes to Emscripten itself since this was opened which may interfere or not work with the changes here. 

I will try to remember to update the samples at some point.

PR #2082, opened 2019-02-14
Added project files for Visual Studio 2017.

paulhoux:
Would be nice to add this to the continuous integration as well.

### DIFF
--- a/.appveyor_vc2017.yml
+++ b/.appveyor_vc2017.yml
@@ -1,0 +1,34 @@
+#version: 1.0.{build}
+
+build:
+  parallel: true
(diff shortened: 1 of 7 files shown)


paulhoux:
I'm going to close this one, as I am not sure how to do the CI stuff and it's getting a bit messy.

richardeakin:
I also wanted to note that  VS 2019 is right around the corner, currently in RC phase. Anyone else think we should skip 2017 and move directly to 2019?

brucelane:
I used Cinder with VS2017 for a while, I'm personnally fine with moving directly to 2019.

PR #2083, opened 2019-02-17
Build libs test samples for vc2015 x86/x64

tenk-a:
Error on link to build Samples and Test with vc 2015(vc2017) x64,
Create libpng.lib, cairo-static.lib, pixman-1.lib with vc2015,
The project file for Samples/, Test/ vc has been modified to correspond to vc2015 x86/x64.

For the time being, my libpng, cairo, pixman build working environment  
   tools/scripts/msw_build_libs/  
(The procedure is readme.txt)

For samples test, we convert vc11 and vc2012 to vc2015 as much as possible and set each vc2015 project file to x64 compatible.
Problems occurred Since cpp and h are not fixed, they are build errors. (Scheduled to be modified separately)

To build Samples and Tests altogether  
     test/_AllTest/vc2015/build_all_test.bat  
     samples/_AllSamples/vc2015/build_all_samples.bat  
Are prepared.

* sln or vcxproj is a substitute that changed for vc2015 by replacing vc141 with vc140 once generated for vc2017. The versions in sln and vcxproj are newer than vs 2015, but it seems that there is no problem with use with vc 2015, so we keep it.


richardeakin:
Not sure I follow what this PR is adding, since it's showing 408 changed files. Are these multiple fixes, are there issues, etc.?

PR #2085, opened 2019-03-05
[CMake] Use quotes for targets filepath.

PetrosKataras:
Solves issues with paths containing whitespaces. 

Came across this when trying out Visual Studio 2017 `Open Folder` feature where the default CMake configuration is something like `Debug (default)` and this reflects on the exported path. 

Using double quotes enforces a textual representation of the path which guards against any whitespaces.


richardeakin:
Thanks!

PR #2087, opened 2019-03-30
Fixed centering issue on Mac

matchingponies:
The content was in the upper right hand corner on recent Mac builds so I updated this sample to handle high density screens

Added setHighDensityDisplayEnabled() and passed prepareSettings into CINDER_APP


richardeakin:
Thanks!

PR #2093, opened 2019-05-13
Making getNetworkAdapters() work on Linux

gaborpapp:
The same code works in macOS and Linux. Only the required header files were needed to make it work.


richardeakin:
Thanks!

PR #2095, opened 2019-05-14
Handling borderless and resizable options in GLFW window creation

gaborpapp:
Basic additions to handle window decorations in Linux/GLFW.


richardeakin:
Nice, thanks!

PR #2112, opened 2019-07-02
Linux unicode input

### LINKED ISSUE #1661: Support unicode text input with GLFW
We're currently using [`glfwSetKeyCallback`](http://www.glfw.org/docs/latest/group__input.html#ga7e496507126f35ea72f01b2e6ef6d155) for text input, which doesn't support unicode. We need to use `glfwSetCharCallback` or `glfwSetCharModsCallback` to get that working, however it involves some bookkeeping to keep functionality consistent with windows and mac, specifically around the down versus up state.

Strawman steps:

1. SetKeyCallback() fires with an `action ==  GLFW_PRES`. We store the scan code, actions and mods in a map
2. glfwSetCharModsCallback() fires, and we get the actual unicode key. If it is a GLFW_PRESS, then we emit a keyDown signal
3. the next time SetKeyCallback() fires, if the `action == GLFW_RELEASE` (or whatever it's called), then we look up all the info in the map, including the stored unicode char, and then emit a keyUp()

richardeakin:
This is an initial implementation of unicode text input on linux / GLFW platforms. It requires storing an extra `ci::app::KeyEvent` to combine GLFW's `glfwSetKeyCallback` with the `glfwSetCharCallback`, but we discussed and thought this was the most flexible approach. Please let me know what you think, and if you write in a language that requires unicode support then testing is very welcome. :)

I've added a little code to the TextBox sample, and some logging that shows what's going on, which I'll remove before merging.

Fixes #1661

### DIFF
--- a/include/cinder/app/KeyEvent.h
+++ b/include/cinder/app/KeyEvent.h
@@ -43,6 +43,8 @@ class CI_API KeyEvent : public Event {
 	uint32_t	getCharUtf32() const { return mChar32; } 
 	//! Returns the key code associated with the event, which maps into the enum listed below
 	int			getCode() const { return mCode; }
+	//! Returns the modifiers associated with the event. Can also use the convenience methods below to check for a specific modifiered,
+	unsigned int	getModifiers() const	{ return mModifiers; }
 	//! Returns whether the Shift key was pressed during the event.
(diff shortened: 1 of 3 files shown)


richardeakin:
This has been tested by a third party and verified working. Debug logging code has been removed. Will merge soon barring any additional feedback.

PR #2113, opened 2019-07-06
VC2019 / Visual Studio project configs for v142 compiler toolset

richardeakin:
As discussed in #2082, skipping VC2017 folders since Visual Studio 2019 / v142 is stable, freely available, and in my opinion much better than Visual Studio 2017 (and 2015).


vinjn:
Agree, also vs2019 compiles faster than vs2015.

fei4xu:
I would suggest we skip the vs project *.vxcprojt totally, since current cmake system works fine on Windows and it's cross platform.
VS2019 also has direct cmake support.

richardeakin:
@fei4xu we may think about that some time in the future, but up until now (and currently), very few Windows users are building via cmake - they all depend on the included VS project files, so we'll continue to supply them until that changes. A large part of this decision is that cmake-generated visual studio project files are difficult to hand tune, and are generally setup pretty poor compared to those we make by hand.

Using cmake directly in Visual Studio is interesting, although my attempts to use it have shown that it is still quite experimental, especially with more complex project structures.

vinjn:
@richardeakin @andrewfb can we get this merged? My project was broken by `std::experimental::filesystem`

richardeakin:
Yea, it's a bummer. I think these changes will make it into master branch first thing after the next release lands (not exactly sure the ETA of that, I know @andrewfb has some smallish things pending around getting a few deps built first, win32 samples fixed but it's otherwise on the way). In the meantime feel free to pull my branch, I'll keep it updated with latest master as needed. Or if it's a huge hassle we can probably push a temp branch to the main repo..

richardeakin:
Closing, superseded by #2138

PR #2114, opened 2019-07-12
Surface and Texture - cast data size to uint32_t to prevent value overflow.

vinjn:


andrewfb:
Thanks for this - can you give a little more background on what you're seeing here though? I don't have any doubt we need some more rigor on this, but in theory you could have at least a `ci::Surface` that is requires more than 4GB, though 2^31 is the maximum on either axis.

MikeGitb:
And how does this actually prevent overflow? You'd have to cast the operands to e.g. std::size_t and then multiply them to prevent overflow.

vinjn:
I happen to have a image that's `28248 x 26800`.

# int32_t #

image

# uint32_t #
image

MikeGitb:
The result of multiplying two int32_t is again a in32_t - no matter what you assign it to. 
In this particular case, you have an overflow to a negative number that you assign to an unsigned number, which in this particular case results in three number you want.

The initial overflow however is still undefined behavior (your program is litterally allowed to do anything,  but the observed behavior is the most likely outcome) and if you need more than 4GB, just assigning the result to a size_t will most likely not give you the correct result anymore. You really need to first cast the operands. 

The other question is of course what other parts of the code make the assumption that the image is at most 2 or 4 GB big.

PR #2119, opened 2019-07-28
Add getFrameDelta() method to AppBase

willemkempers:
Inspired by https://forum.libcinder.org/topic/getting-the-delta-value-of-single-frames to get the delta in seconds between the last update and the current update.

### DIFF
--- a/include/cinder/app/AppBase.h
+++ b/include/cinder/app/AppBase.h
@@ -352,6 +352,8 @@ class CI_API AppBase {
 	double				getElapsedSeconds() const { return mTimer.getSeconds(); }
 	//! Returns the number of animation frames which have elapsed since application launch
 	uint32_t			getElapsedFrames() const { return mFrameCount; }
+	//! Returns the number of seconds which have elapsed since the last update call
+	double 				getFrameDelta() const { return mFrameDelta; }
 
 	//! Returns whether the app is registered to receive multiTouch events from the operating system, configurable via Settings at startup. Disabled by default on desktop platforms, enabled on mobile.
 	bool				isMultiTouchEnabled() const				{ return mMultiTouchEnabled; }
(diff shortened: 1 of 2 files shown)


MikeGitb:
How about sing std::chronologisch types for that?

willemkempers:
Yes - I agree that using std::chrono and returning a std::chrono::duration would make sense. Decided to stick with a Timer object and return a double for consistency reasons as that's what getElapsedSeconds() uses.

richardeakin:
Do you need a second `ci::Timer` for this, when the time is already kept in AppBase? I would think you could achieve the same with a `double mLastTimeUpdated`. This is what I usually do in my app.

To be completely honest, I actually don't do that in my app, I almost always use a [`WorldColock`](https://github.com/richardeakin/mason/blob/master/src/mason/WorldClock.h) class that optionally gives you a [fixed time step](https://gafferongames.com/post/fix_your_timestep/), works much better for many simulation type things.

(Apologies for taking a while to check this..)

willemkempers:
Reason for the second timer is pointed out by sansumbrella in the linked forum thread: numbers lose precision as they get farther away from zero.
While that's true, I'm unsure how big of an effect it would actually have as I haven't tested it. A `double mLastTimeUpdated` might be good enough.

Interesting article! The author does use `double frameTime = newTime - currentTime;` which is what I'm proposing here. He then continues to expand on this in order to get a tight update cycle for his simulation.

paulhoux:
I'm fine with a second `Timer`, as a `double` can indeed lose some of its precision after a (admittedly long, as in days) while. However, I'd suggest to rename the function to something which includes `Seconds` or `Time`, as to make it clear we are not returning frame numbers. Maybe `getDeltaSeconds()` would do?

willemkempers:
Sorry for the delayed reply - I'm fine renaming the method to getDeltaSeconds() or something along those lines! Think I went for getFrameDelta as delta in this context intuitively means time for me, but I can see how it might be confusing.

PR #2123, opened 2019-08-28
Prevent clientWaitSync to timeout immediately with default argument.

num3ric:
Here's another example:

https://github.com/google/angle/blob/master/src/tests/gl_tests/FenceSyncTests.cpp#L257


num3ric:
It also has the potential to _fix_ any code which relying on the default value :)

num3ric:
Shall we merge this now that in 0.9.3 territory?

PR #2125, opened 2019-09-12
Fix serial device list on Linux.

PetrosKataras:
Came across this while working with DMX controllers.


richardeakin:
Makes sense to me, thanks for the PR!

PR #2130, opened 2019-11-08
Fixed fog shader in samples/_opengl/DeferredShadingAdvanced + de-boostifying JsonTree

BanTheRewind:


richardeakin:
My opinion is we should modify the title of this PR to add "de-boostifying JsonTree" and accept it. We can remove Json.cpp / replace it with something else in a later step, as has been discussed, however these changes still seem good to get into the next release.

PR #2132, opened 2020-01-07
Added noexcept to signals move constructors and assignment operators.

### LINKED ISSUE #1681: signals::ScopedConnection should use 'noexcept'
For move constructor and move assignment operator, so that std::vector will use move unstead of copy. Waiting until we jump to vc140 (Visual Studio 2015) on Windows as `noexcept` isn't supported with vc120.

See comments in #1676.

richardeakin:
This will allow `std::vector<Connection>` and `std::vector<ScopedConnection>` to resize a little more efficiently. Was only waiting until our minimum compiler on MSW was v140 which is now the case.

Fixes #1681. That issue is marked for 0.9.2 milestone, although this is a pretty minor modification so it can be merged before or after that lands.

### DIFF
--- a/include/cinder/Signals.h
+++ b/include/cinder/Signals.h
@@ -123,9 +123,9 @@ class CI_API Connection {
 	Connection();
 	Connection( const std::shared_ptr<detail::Disconnector> &disconnector, detail::SignalLinkBase *link, int priority );
 	Connection( const Connection &other );
-	Connection( Connection &&other );
+	Connection( Connection &&other ) noexcept;
 	Connection& operator=( const Connection &rhs );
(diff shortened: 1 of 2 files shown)


MikeGitb:
Looks good to me. 

Best I can tell, you could also add `noexcept` to all other constructors and assignment operators of connection and ScopedConnection while you are at it, but I'd only do that if you want to commit to keep them noexcept in the future (I think that would be a reasonable choice, but it doesn't really matter all that much).

richardeakin:
I suppose we can, if for nothing else than congruity. There shouldn't be any exceptions in this code except with the possibility of `Signal::emit()`.

PR #2134, opened 2020-02-13
Fix all blocks and samples on linux, add cairo support

### LINKED ISSUE #1483: Get all samples building and running with new cmake

### LINKED ISSUE #1815: can't build Cairo based samples on Linux
I'm new to the Cinder codebase so hopefully I'm not just missing something obvious.
I've built libcinder without any issues and now I'm going through the samples and some work, some don't.

In the case of the Cairo based sample CairoBasic you start with some simple missing include files and if you fix those then eventually you get
```
/home/blahh/github/Cinder/samples/CairoBasic/src/CairoBasicApp.cpp:115:29: error: no type named 'createWindowSurface' in namespace 'cinder::cairo'
        cairo::Context ctx( cairo::createWindowSurface() );     
                            ~~~~~~~^
```
and if you look in blocks/Cairo/src/Cairo.cpp you see the definition
```
#if defined( CINDER_MSW )
cairo::SurfaceGdi createWindowSurface()
{
	return cairo::SurfaceGdi( cinder::app::App::get()->getRenderer()->getDc() );
}
#elif defined( CINDER_MAC )
cairo::SurfaceQuartz createWindowSurface()
{
	auto cgContext = cinder::app::App::get()->getRenderer()->getCgContext();
	auto height = cinder::app::getWindowHeight();
	CGContextTranslateCTM( cgContext, 0.0, height );
	CGContextScaleCTM( cgContext, 1.0, -1.0 );
	return cairo::SurfaceQuartz( cgContext, cinder::app::getWindowWidth(), height );
}
#endif
```
There is no Linux definition for createWindowSurface()

I see that @ryanbartley has a big Cairo pull in the queue, does that fix this?

### LINKED ISSUE #2022: some samples CMakeLists.txt missing "INCLUDES    ${APP_PATH}/include"
I've been doing some diving into building with linux.
I've just now come across this missing line in the CMakeLists.txt for the MandelbrotGLSL example. I'm making the assumption that it might be missing on some of the other examples as well.

This is more of a question really. As I might be willing to make a pull request. But before I went too deep in, I wanted to ask if these CMakeLists.txt files were automated is some way. Maybe the problem can be fixed higher up in the chain.

### LINKED ISSUE #2052: About half the samples don't build on Linux

ppff:
I wanted to try Cinder for an upcoming project. I'm a full time linux user, and couldn't help but notice that most samples wouldn't build on Ubuntu 18.04. 

Hence, I reviewed each block and each sample to make sure they now build and work (except those that aren't supported on linux of course). I added the samples that aren't supported to the `CINDER_SKIP_SAMPLES` list.
This also led me to add linux support for Cairo because I wanted to try the `_svg` samples. I've changed a few bits of code in the Cairo block, one should try to build it on windows and osx to make sure I didn't break anything. If you want a second PR just for that, I can rework my branch.

This PR should close #1483 close #2052 close #2022 and maybe close #1815 (most Cairo samples also require `Renderer2D` which is still not implemented on linux). 

Any comments would be appreciated.


proj/cmake/modules/findCMakeDirs.cmake
```diff
@@ -6,7 +6,7 @@ macro( findCMakeDirs RESULT_PATHS BASE_PATH SKIP_PATHS )
 			set( shouldSkip FALSE )
 			foreach( skip ${SKIP_PATHS} )
 				if( f MATCHES ".*${skip}" )
-					ci_log_v( "---- [findCMakeDirs] skipping path: ${f}" )
```
ppff: I just think it's important for users to know which samples are ignored, but that's debatable.

blocks/Cairo/include/cinder/cairo/Cairo.h
```diff
@@ -44,47 +44,8 @@
 #include <vector>
 #include <iomanip>
 
-// Forward declarations used by our cairo wrappers
```
ppff: I didn't see the reason for this, but if there actually is one we can restore the code and just add an IF macro. During the linux build it would create conflicts with cairo.h

samples/_opengl/ParticlesBasic/src/ParticlesBasicApp.cpp
```diff
@@ -39,7 +39,7 @@ class BasicParticleApp : public App {
 	bool	isOffscreen( const vec2 &v );
 
 public:
-	static const int	NUM_PARTICLES = 10000;
```
ppff: This is very heavy for the GPU so I thought it would be nicer to have a fluid sample.

blocks/Cairo/src/Cairo.cpp
```diff
@@ -344,7 +348,7 @@ SurfaceEps::SurfaceEps( const fs::path &filePath, double widthInPoints, double h
 	: SurfaceBase( (int32_t)widthInPoints, (int32_t)heightInPoints )
 {
 	mCairoSurface = cairo_ps_surface_create( filePath.string().c_str(), widthInPoints, heightInPoints ); 
-	cairo_ps_surface_set_eps( mCairoSurface, TRUE );
```
ppff: This should not break compatibility with Windows and OS X.

blocks/Cairo/src/Cairo.cpp
```diff
@@ -2054,7 +2060,7 @@ class SvgRendererCairo : public svg::Renderer {
 			fontMatrix *= rotationMatrix;
 			mCtx.setFontMatrix( fontMatrix );
 			TextBox tbox = TextBox().font( *font ).text( span.getString() );
-			std::vector<std::pair<uint16_t,vec2> > glyphs = tbox.measureGlyphs();
```
ppff: This should not break compatibility with Windows and OS X.

richardeakin:
Really awesome that you took this on, thanks! It will be great to finally have all samples building in Linux, a big step in cross-platform equality. We're gearing up for a release within the very near future, so I think we'll try to get this work integrated within the next release cycle.

I think we should probably simplify this PR to things related to cairo / svg samples as much as possible. I haven't yet had time to review or run this code, but I hope that we can soon.

blocks/Cairo/include/cinder/cairo/Cairo.h
richardeakin: Are the conflicts because the forward declarations need to be updated? I haven't worked with the Cairo.h wrapper personally, but my guess is that this was avoiding the heavy `#include <cairo.h>` so as cinder end user apps don't have to bring all of those symbols into their files. In general, we try not to include heavy system headers from within cinder headers.

proj/cmake/modules/findCMakeDirs.cmake
richardeakin: Makes sense to me. `CINDER_SKIP_SAMPLES` is user-settable anyway, I added it as a stop-gap soln for building all samples in one go. Hopefully, it won't be needed much after this work gets merged. :)

proj/cmake/platform_linux.cmake
```diff
@@ -4,6 +4,31 @@ set( CMAKE_VERBOSE_MAKEFILE ON )
...
 
+# Some samples are not available on Linux
+list( APPEND CINDER_SKIP_SAMPLES
+    # Renderer2D is required for:
+    FontSample
+    perlinTest
+    BSpline
+    CairoBasic
+    Wisteria
+    Renderer2dBasic
```
richardeakin: Ah, I see. I guess this is good documentation for what left is needed for getting the samples running on linux (or why they never will). As an aside, what is left to get Renderer2D running after this? If you have ideas for that perhaps we should bring to a new github issue.

samples/_opengl/ParticlesBasic/src/ParticlesBasicApp.cpp
richardeakin: Perf. issues are actually due to the CPU issues in Modern-GL and the 'immediate mode' gl wrapper. I think it looks better with 1000 particles and still runs really fast on my machine. Although in general, I think it's best to leave unrelated changes like this to separate PRs.

ppff: Agreed, I removed it in my new PR #2135

ppff:
In order to make things clearer I created two separate branches from cinder/master and cherry picked my commits to them.
We should now work with the new PRs #2135 and #2136.

proj/cmake/platform_linux.cmake
ppff: Renderer2D isn't too much of a problem, I've already looked into it a little bit :)

blocks/Cairo/include/cinder/cairo/Cairo.h
ppff: Let's talk about it in #2136 :)

PR #2136, opened 2020-02-14
Add cairo block to linux

### LINKED ISSUE #1815: can't build Cairo based samples on Linux

ppff:
This is a child PR from #2134.

Cairo block now builds on linux, along with all the `_svg` samples.
The code was slightly changed so it should be tested on windows and mac os.

Should close #1815

### DIFF
--- a/blocks/Cairo/cinderblock.xml
+++ b/blocks/Cairo/cinderblock.xml
@@ -4,10 +4,11 @@
 	name="Cairo"
 	id="org.libcinder.cairo"
 	author="Cinder Project"
-	summary="Cairo support for Mac OS X and Windows."
+	summary="Cairo support for Mac OS X, linux and Windows."
 	>
 	<supports os="msw" />
 	<supports os="macosx" />
+	<supports os="linux" />
 	<includePath>include</includePath>
 	<source>src/Cairo.cpp</source>
 	<header>include/cinder/cairo/Cairo.h</header>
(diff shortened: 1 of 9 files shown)


nametable:
What is keeping this PR from being merged?

PR #2138, opened 2020-02-18
VC2015- > VC2019, Dropping Win32

andrewfb:
Continuation of work in #2113 - updates Cinder itself and all samples to require VC2019, and c++17. Also drops Win32 in favor of x64-only.


MikeGitb:
I think you have forgotten to upgrade the tests - or is this still WIP anyway?

andrewfb:
> I think you have forgotten to upgrade the tests - or is this still WIP anyway?

I've done the Unit tests, but I think the remainder of the tests will be piecemeal - a number of them are pretty outdated at this point anyway, and I don't want to delay this to address that.

richardeakin:
I agree that we can upgrade the tests and anything else that comes up afterward. We have an entire release cycle IMO to make sure everything is running smoothly out of the box with v142 / VS 2019.

PR #2139, opened 2020-02-20
Completing boost removal

andrewfb:
This PR completes the ongoing removal of Boost. In so doing, I found a few spots where we were implicitly including headers which I fixed. In addition to macOS, Windows, and iOS, I also built against Ubuntu 19. This revealed a few places we were missing headers as well.

Ubuntu 19.04 manifested this issue, which seems to be a problem with distro itself, and which I confirmed is correct in 19.10:
https://stackoverflow.com/questions/56738708/c-stdbad-alloc-on-stdfilesystempath-append

This PR updates the Linux cmakefiles to detect and make use of C++17.


MikeGitb:
> This PR updates the Linux cmakefiles to detect and make use of C++17.

I think this whole machinery should get cleaned up / removed in a follow up PR.

As I commented in another PR: CMake has a built-in,  crossplatform mechanism to set the c++ standard, namely CMAKE_CXX_STANDARD. Also I think that - unless it is a hard requirement of the library- it should be left to the user to decide,  with which c++ standard the library should be compiled (cmake -DCMAKE_CXX_STANDARD=17 <path-to-cinder>).

richardeakin:
Awesome, I think this will really simplify integration into complex codebases, and also make the repo much lighter in general.

@MikeGitb I think we can and should continue to improve our CMake setup, there are a number of pending PRs that will likely get reviewed and accepted soon. I do believe that we will require C++14 however, and then default to 17 where available (unless overridden by the user). If there is a better method than what we're currently using, perhaps we should create a new issue about this to track?

Also, just want to throw out there that I think it would be nice if we had some sort of boost cinderblock that simplifies boost integration, for those times when you need it. This isn't as huge of an issue on linux where boost usually gets installed globally, though it is on Windows.

andrewfb:
I'm definitely game to see CMake changes - not my area of expertise, but for now I wanted to test how std::filesystem available in libstdc++ would work. I've since tested this code on RPi4 - going to go ahead and merge.

PR #2140, opened 2020-02-20
Glad loader for all platforms.

num3ric:
This continues the work done in the #2020 PR. (I've squashed it all into a single commit to clean up my commits mess -- sorry Paul.) Many more extensions have been included in `glad.h` but unused/old manufacturers or extensions based on older GL versions have been stripped out: https://gist.github.com/num3ric/96c3c99615ea75612dc628e4e8af66e5 I think this is a good compromise between maximum flexibility & impact on build times.

Tested on the following platforms:
- Windows
- Ubuntu
- Mac/iOS/iOS Sim
- Raspberry Pi 4 (This now uses GLES 3.0 and glfw.)

To make this work on rpi4, I've started to deprecated its ES2 path. I'd suggest that further removal of ES2 should be done later as separate commits.

Before calling this good, I will test headless rendering as well.


num3ric:
Glad's reliance on `eglGetProcAddress` causes issues with the Angle path so I reverted the old includes.
https://github.com/Dav1dde/glad/issues/79
https://github.com/seshbot/glgen/blob/d7751a51367308daaebee573307178421a7cdc7c/assets/cpp/src/gles2/angle_ext.cpp#L1

As an aside, the EGL headless test was successful.

andrewfb:
This is looking good to me - verified it with ANGLE as well. I'd be inclined to merge this soon.

andrewfb:
Thanks for this @numeric and also @paulhoux for your related work on #2020

PR #2141, opened 2020-02-20
Terminate handler instead of global try-catch

num3ric:
Consider this more as a suggestion -- needs more testing.

This is an alternative to the global app try-catch. Debuggers and crash dumps can often point to catch re-throw and hide the exact source of an exception. This would be prevented with this approach, while preserving the error logging.

Thoughts?


andrewfb:
After testing, merging this - this is a better approach to what we were doing previously.

PR #2143, opened 2020-02-26
Adding (nvidia) multicast support

num3ric:
This enables the use of NVIDIA multicast extensions in Cinder (`GL_NV_gpu_multicast`, `GL_NVX_gpu_multicast2` and `WGL_NV_multigpu_context`). This is particularly useful for driving large display walls with multiple (linked) gpus, allowing users to more explicitly balance workloads between GPU devices while retaining a single OpenGL context.

Most convenience functions exist under `gl::nv::multicast`, but also in other places (`RendererGl`, `BufferObj`, `Texture`).

A sample for 2 GPUs is also included:
multicast_preview

A few more tasks are required (such as ensuring that shared context creation also use the multi-gpu attribs), but I am creating this PR now immediately visibility!


num3ric:
One point of discussion is whether `multicast::bufferSubData` and `multicast::copyImageSubData` functions should be in the `BufferObj` and `TextureBase` classes respectively, or if they should stay in the `multicast` namespace.

It might be more natural if those were in their respective classes, but there's also an argument to keep the extension/vendor specific code grouped? No strong opinions.

samples/_opengl/NvidiaMulticast/src/NvidiaMulticastApp.cpp
```diff
@@ -200,7 +200,10 @@ void NvidiaMulticastApp::draw()
 		multicast::disableRenderMask();
 	}
 	else {
-#if ! defined( ASYMMETRICAL_CAMERAS )
+#if defined( ASYMMETRICAL_CAMERAS )
+		multicast::viewport( mGPUs[0], ivec2( 0 ), ivec2( getWindowWidth() / 2, getWindowHeight() ) );
+		multicast::viewport( mGPUs[1], ivec2( getWindowWidth() / 2, 0 ), ivec2( getWindowWidth() / 2, getWindowHeight() ) );
+#else
```
num3ric: This works correctly.

num3ric:
I think all the pieces are there now, open to suggestions.

include/cinder/gl/nv/Multicast.h
```diff
@@ -0,0 +1,87 @@
...
+
+				//! Restricts render commands to the specified device mask.
+				CI_API void enableRenderMask( GLbitfield mask );
+				//! Restricts render commands to the specified device.
+				CI_API void enableRenderMask( const Device& device );
+				//! Disable render commands (device) restrictions.
+				CI_API void disableRenderMask();
+
+				//! Signal other device to wait for this GPU to finish its work. (For multiple target devices, use the bitfield method.)
+				CI_API void signalWaitSync( const Device& source, const Device& target );
```
iesser: Order like other APIs sourrounding it, mask/device instead of device/mask?

samples/_opengl/NvidiaMulticast/src/NvidiaMulticastApp.cpp
```diff
@@ -0,0 +1,270 @@
...
+	Json::Reader reader;
+	reader.parse( in, settings );
+
+	RendererGl::Options options;
+	if( settings["multicast_multidisplay"].asBool() ) {
+		//! This options should be use on Quadro setups with Mosaic enabled, with each display connected to independent GPUs.
+		options.multiGpuMultiDisplayMulticastNV();
+	}
+	else {
+		//! This option should be used on GeForce setups, with SLI "Maximize 3D Performance" enabled.
```
iesser: This path can also be used on Quadro, same setting in the Control Panel

num3ric: Thanks for reviewing! Both notes should be addressed in the latest commit.

PR #2145, opened 2020-02-28
Dear ImGui integration (deprecating AntTweakBar in all samples)

num3ric:
This integrates ImGui v1.75 in Cinder, much of it based on @simongeilfus block. All samples (except `ParamsBasic`) have been updated. A few convenience functions/scoped structs have been added, but the main philosophy is to retain the native `Dear ImGui` API.

More to come, but creating now for visibility.


brucelane:
thank you for this @num3ric !

ocornut:
Hello, author of dear imgui here!

1) As a nitpicker, may I request that futture references to the library in changelog/text/comment can use the full name "Dear ImGui" name ? The namespace in code is still called `ImGui::` but it's been an ambiguous name anwe're d trying to promote the full name.

2) If Cinder is frequently used for fullscreen interactive applications or art stuff, it may be useful to wire gamepad controls as well (and enable `ImGuiConfigFlags_NavEnableGamepad`).

3) I would suggest recreating a clean PR with same contents but in two commits.
- one commit with raw vanilla dear imgui files
- one commit with your own additions
- maybe consider preserving the file structure from raw imgui OR provide a .bat file to quickly "copy from the original file structure into your flattened one" so nothing is missed when updating.
So you'll start from a neat initialize state which will make it more natural for you to update further.

4) May or not be useful, but note that using the `BeginCombo`/`EndCombo` api tends to be more flexible than the old `Combo` api when pulling from existing data. (I believe `Combo` was the only api available at the time when the Cinder block was initially created). In the case of "quick inline debug" the simpler `Combo` is still useful, but if you are pulling from existing data `BeginCombo/EndCombo` tends to be more flexible.

num3ric:
@ocornut Thanks for reviewing! All valid comments which I hope to address!

I suppose other than preemptively updating all the samples as a "sales argument", this PR can be considered more of a host for integration discussions before we move into a new clean PR!

In discussing this with @richardeakin (this relates to your third point), I agree that ease of update is currently lacking here. I might lean towards preserving the existing ImGui folder structure.

Another discussion item is whether additional convenience functions should live in the `ImGui::` namespace natively or not. It's desirable for convenience conversions, but maybe less so for other things?

ocornut:
> In discussing this with @richardeakin (this relates to your third point), I agree that ease of update is currently lacking here. I might lean towards preserving the existing DEAR ImGui folder structure.

I've been trying to keep the full repo as lean as possible but without Git cruft it is stil 176 files and 4.5 MB so up to you. From a vanilla repo the folders you are most likely to want to remove are examples/ and misc/fonts/. If you remove those it becomes 28 files, 2.18 MB which is not too far from as lean as you could get it, and you can state that your update policy is to replace the folder and delete those two.

> Another discussion item is whether additional convenience functions should live in the ImGui:: namespace natively or not. 

That's up to you! 
Depending on the topic above you may choose to store your own file in a separate folder to facilitate understanding what's Cinder and what's core library.

richardeakin:
Cool to see this happening, I know it's been talked about for a while now and just about all regular cinder users are Dear ImGui fans so it'll be good that samples and new user apps can get it out of the box.

So we had a few discussions about how to facilitate updating the Dear ImGui sources, or perhaps altogether opting out of Dear ImGui in libcinder. The latter could be necessary when you're using libcinder with an existing codebase and you want to avoid the nameclashes, for that right now we only can do it with cmake ([see here](https://github.com/cinder/Cinder/blob/master/proj/cmake/configure.cmake#L6-L8)). But for larger codebases, cmake option makes a lot of sense.

Then there is the case that you (cinder user) wants to use a newer version or a branch of Dear ImGui, and needs to know how to update the code locally. I think the only two decent options are submodule it, which comes with its own set of drawbacks, or .bat / .sh script it.  If it is scripted, we'll want to keep the Dear Imgui file structure as close as possible to the originals, and then strip and patch where needed.

I do like that this PR sticks with using the standard `ImGui` namespace, although personally I don't think that the `glm` overloads add much. For example it saves you from having to write the extra `.x`, but at the cost of the syntax being different to existing `Dear Imgui` things. I do think there is room for both the scoped / RAII stuff and a higher level C++ wrapper that could use named return pattern like many of cinder's options (@simongeilfus wrote one of these), both either of these could live in separate files and users could opt into using them, similar to how we have generally structured the `ci::gl` namespace.


Koodoos for updating all the samples!

num3ric:
Regarding the vector conversions, those are mainly aimed at beginners and allow sample code to be as simple as can be (similar to how params used to work). Furthermore, those follow the same conventions as the `ci::Color` types, which are very much desired. Worth keeping IMO,

num3ric:
I know @richardeakin and @simongeilfus are requesting we use the `docking` branch instead. However, using a tagged master `Dear ImGui` release ensures a stable baseline (that's fully tested and stable on all platforms), which I would recommend should be the starting point for a first integration PR.

richardeakin:
@num3ric  this is precisely why I think there needs to be some system in place to ease updating the embedded version of Dear Imgui. If users have to figure out exactly what files to replace and where they go, it's going to create huge amounts of frustration (and worth noting, that many cinder users are going to be frustrated regardless, particularly those who have voiced that they don't think a framework should be shipping their own copies of thirdparty libraries).

To my mind, it makes a submodule in this case much more attractive.

tools/dearimgui_update.sh
```diff
@@ -0,0 +1,17 @@
...
+libcinder_path=$2
+
+target_inc=$libcinder_path/include/imgui
+target_src=$libcinder_path/src/imgui
+
+cd $dearimgui_path
+cp -v *.h $target_inc; cp -v *.cpp $target_src
+cp -v misc/cpp/*.h $target_inc; cp -v misc/cpp/*.cpp $target_src
+cp -v misc/freetype/*.h $target_inc; cp -v misc/freetype/*.cpp $target_src
+cp -v examples/imgui_impl_opengl3.h $target_inc; cp -v examples/imgui_impl_opengl3.cpp $target_src
```
num3ric: This script should allow easy updates via: `./dearimgui_update.sh ~/Code/imgui ~/Code/Cinder` A flat hierarchy on the cinder-side was preserved to prevent additional include paths. Furthermore, although we had discussed hiding/privatizing implementation files like `imgui_impl_opengl3.h/cpp` into `src\imgui`, I but I kept in public to allow uses cases where the automatic rendering is disabled.

num3ric:
Moved this PR to a new cleaner branch to address the discussion points here. Should be much easier to update now.

@ocornut Since there is (popular) demand for integrating the `docking` branch, would you recommend a particularly commit (history point) in that branch that would be deemed stable enough? Or would you recommend holding off on that for the moment?

PR #2146, opened 2020-03-06
Dear ImGui Integration

num3ric:
Continuation of the #2145 pull request, with cleaned-up commits.


num3ric:
Fyi, updating to the Dear ImGui `docking` branch doesn't seem to require any code changes (to get a working build at least). Running `python tools\scripts\dearimgui_update.py ~/Code/imgui ~/Code/Cinder` does the trick, and yields the following https://github.com/num3ric/Cinder/tree/dearimgui_docking

andrewfb:
This is looking good to me. I've tested it on macOS, MSW and Ubuntu 19.10. I'd be inclined to merge it soon.

felixfaire:
Very excited for these changes, well done / thanks to everyone involved. 

One feature that was very difficult with the CinderImGui block was loading multiple fonts with the same filename (to handle different ui text sizes etc), as the fonts were stored in a map<string, ImFont*>. This is partly a feature request but will it be possible to use the native dear Imgui AddFont methods or is there specific cinder font handling with this update?

num3ric:
@felixfaire I just pushed a change that delays the first `NewFrame()` call to `resize()`, and should therefore allow users to use custom fonts like so:
```
ImGui::Initialize();

ImGuiIO& io = ImGui::GetIO();
io.Fonts->AddFontFromFileTTF( getAssetPath( "Saint-Andrews Queen.ttf" ).string().c_str(), 20.0f, NULL, io.Fonts->GetGlyphRangesDefault() );
```

More references here: https://github.com/ocornut/imgui/blob/master/docs/FONTS.txt

I haven't personally used multiple fonts with `Dear ImGui`, but it certainly could make sense at some point to reintegrate those custom font/icon convenience functions.

num3ric:
Successful test on the RPI4:
rpi4 dear imgui

However, I did have to add `target_link_libraries(${PROJECT_NAME}  stdc++fs)` to the sample's cmake to fix a `std::filesystem` issue. TBD on how to best fix this globally, but unrelated to the `Dear ImGui` changes.

Everything else here should be quite solid now.

richardeakin:
I agree it's in a nice place for integration, glad to see it moving forward! We'll probably have to tweak build scripts as time goes by and different situations, we'll see how that goes with use.

PR #2147, opened 2020-03-10
Remove boost artifacts from macOS cmake configuration.

PetrosKataras:
This seems to be a residual from the recent Boost dependency removal that got missed on macOS and currently breaks compilation through CMake.

Tested on macOS 10.13.6 with Clang 9.0.0 and Cinder's current `master` branch.


andrewfb:
Thanks for this catch

PR #2149, opened 2020-03-20
Added quasi-random number generator and support for Rand<double>.

paulhoux:
Based on the excellent article by Martin Roberts:
http://extremelearning.com.au/unreasonable-effectiveness-of-quasirandom-sequences/

The `QuasiRand` class returns quasi random values in the range [0, 1) which are evenly distributed across the full range. When used to generate points on a sphere, it looks like this:

QuasiRandomSequencesApp 20-3-2020 15_40_28

By comparison, fully random points would look like this:

QuasiRandomSequencesApp 20-3-2020 15_40_56

The great thing about the algorithm is that it doesn't require you to know the number of values up front. Here's the same sphere after simply adding 4 times as many points:

QuasiRandomSequencesApp 20-3-2020 16_06_09

The class is templated and uses `float` by default, but also supports `double` and `long double`.

I will add a sample later.

Also refactored the `Rand` class by adding a template argument. This allows the use of `double` and `long double` random values. By default it uses `float` values.

### DIFF
--- a/include/cinder/QuasiRand.h
+++ b/include/cinder/QuasiRand.h
@@ -1,0 +1,208 @@
+/*
+ Copyright (c) 2020, The Cinder Project: http://libcinder.org
+ All rights reserved.
+
+ This code is intended for use with the Cinder C++ library: http://libcinder.org
+
+ Portions of this code based on the excellent article by Martin Roberts:
+ http://extremelearning.com.au/unreasonable-effectiveness-of-quasirandom-sequences/
+
+ Redistribution and use in source and binary forms, with or without modification, are permitted provided that
+ the following conditions are met:
+
+	* Redistributions of source code must retain the above copyright notice, this list of conditions and
+	the following disclaimer.
+	* Redistributions in binary form must reproduce the above copyright notice, this list of conditions and
+	the following disclaimer in the documentation and/or other materials provided with the distribution.
+
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED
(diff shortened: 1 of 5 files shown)


richardeakin:
Really cool. And I bet you could use something like a combination of `QuasiRand` + `Rand` to create a sort of blue noise.

One thought - does it need to live in a separate file? Doesn't seem like much code, and I would think this, along with similar random number generators, could all live within _cinder/Rand.h_ / _cinder/Rand.cpp_

paulhoux:
Actually I believe this quasi-random stuff should remain in its own `.h` and `.cpp` files. It's different enough from `Rand` and should be a deliberate opt-in, in my opinion. I guess it's a matter of taste.

totalgee:
Just a side note -- I've used the very good PCG random generator (http://www.pcg-random.org or https://github.com/imneme/pcg-cpp) in Cinder projects, when I needed random generation that was fast and produced the same results (for the same seeding) on different platforms (Mac and Windows, in my case). You don't get that with the standard library (for example `std::mt19937`, which I believe is what `ci::Rand` uses). It's easy enough to use (the header) on its own, but if you're "revisiting" or adding random stuff, it might be something to think about for Cinder...e.g. as a drop-in replacement for Rand?

That said, the Martin Roberts R<sub>2</sub> "golden ratio" noise thing is also a nice addition, thanks! (-;

include/cinder/QuasiRand.h
```diff
@@ -0,0 +1,208 @@
...
+	{
+	}
+
+	//! Resets the quasi-random generator to the specific seed \a seedValue.
+	void seed( uint32_t seedValue ) { mSeed = seedValue; }
+
+	//! Returns a quasi-random float in the range [0.0f,1.0f).
+	T nextFloat()
+	{
+		static T sIrrational = T{ 1 } / phi( 1 );
```
totalgee, line 55: All these static `sIrrational` (and A,B,C variations later in the file) variables should be declared `const`.

PR #2150, opened 2020-03-25
MediaTime

andrewfb:
This PR implements a new `MediaTime` class designed to represent time as a rational number. It is modeled after FFmpeg's [AVRational](https://ffmpeg.org/doxygen/3.3/group__lavu__math__rational.html) and in particular CoreMedia's [CMTime](https://developer.apple.com/documentation/coremedia/cmtime-u58). It's meant to help bridge Cinder into libraries such as those.

The `epoch` is particularly useful when comparing timestamps from one iteration of say, a looping video or audio clip and another.

We've been using this at Rare Volume for the last year and it's been quite helpful. This PR also includes basic unit tests.

### DIFF
--- a/include/cinder/MediaTime.h
+++ b/include/cinder/MediaTime.h
@@ -1,0 +1,205 @@
+/*
+ Copyright (c) 2020, The Cinder Project: http://libcinder.org All rights reserved.
+ This code is intended for use with the Cinder C++ library: http://libcinder.org
+
+ Redistribution and use in source and binary forms, with or without modification, are permitted provided that
+ the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright notice, this list of conditions and
+	the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright notice, this list of conditions and
+	the following disclaimer in the documentation and/or other materials provided with the distribution.
+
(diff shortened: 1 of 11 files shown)


richardeakin:
Looks great, one question - why not call this `ci::Rational`? It seems like it is useful for other time-based applications, is it that it is only intended for use with A/V/media things?

paulhoux:
I see the "rational" in that :)   

Perhaps a `using Rational = MediaTime;` (or vice versa) would suffice?

andrewfb:
Hey - I believe `MediaTime` is an appropriate name here as there are several aspects of its design which are time-centric, as opposed to a more general class like `std::ratio`. In particular, it makes some decisions around the base which mimic what FFmpeg does, defaulting to a common base (microseconds) when a denominator overflow would occur. Additionally, the notion of the `epoch` is very useful for its use case but is not appropriate for a generic Rational class. As a point of comparison, its inspiration's full respective names would be `CoreMediaTime` (CMTime) and `AudioVisualRational` (AVRational). And per Paul's comments, there are rounding choices that are appropriate in conversion from seconds. Later I'd like to look at some more sophisticated rounding behaviors (per CMTime) but I think that can be addressed later.

andrewfb:
I added a couple of convenience methods to this and fixed a bug when multing or div'ing with bases that would overflow. I'd like to merge this sometime tomorrow unless there are any objections.

PR #2151, opened 2020-04-01
Added convenience constructors to Scoped*Matrix classes.

paulhoux:
Allows users to directly set the desired matrices.

Old:
```
gl::ScopedMatrices scpMatrices;
gl::setMatrices( mCam );

gl::ScopedModelMatrix scpModel;
gl::setModelMatrix( mModelMatrix );
```

New:
```
gl::ScopedMatrices scpMatrices( mCam );
gl::ScopedModelMatrix scpModel( mModelMatrix );
```

### DIFF
--- a/include/cinder/gl/scoped.h
+++ b/include/cinder/gl/scoped.h
@@ -208,24 +208,44 @@ struct CI_API ScopedViewport : private Noncopyable {
 };
 
 struct CI_API ScopedModelMatrix : private Noncopyable {
-	ScopedModelMatrix()		{ gl::pushModelMatrix(); }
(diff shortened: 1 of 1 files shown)


richardeakin:
I'm all for this, thought to add it myself for ages. There are so many places where you could combine those two lines and things would be more uniform with other `gl::Scoped*` things.

You can also do things like this:

```
gl::ScopedModelMatrix scpTranslate( glm::translate( vec3( 0, 10, 0 ) ) );
gl::ScopedModelMatrix scpScale( glm::scale( vec3( 5 ) ) );
```

andrewfb:
Nice - thanks for these Paul.

Issue #2155, opened 2020-04-10
mouseWheel event on Windows with HiDPI display gives incorrect position


luksladky:
Scrolling on Windows with content scale set to something >100% gives incorrect position. Instead of returning position in points, it gives position in pixels * getWindowContentScale(). 

In AppImplMsw.cpp it clearly differs from other mouse events. If I change it to match the others, the problem is fixed. I'll make a pull request with a fix.

It can be seen clearly on this example, modified from HighDensityDisplay sample. Scrolling and mouseDragging should both draw under the mouse.

```
#include "cinder/app/App.h"
#include "cinder/app/RendererGl.h"
#include "cinder/gl/gl.h"
#include "cinder/PolyLine.h"

using namespace ci;
using namespace ci::app;

class HighDensityDisplayApp : public App {
  public:
	void mouseDrag( MouseEvent event ) override;
	void mouseWheel(MouseEvent event) override;
	void draw() override;
	
	PolyLine2f		mPoints;
};

void prepareSettings( HighDensityDisplayApp::Settings *settings )
{
	settings->setWindowSize( ivec2( 640, 480 ) );
	settings->setHighDensityDisplayEnabled( true ); // try removing this line
	settings->setMultiTouchEnabled( false );
}

void HighDensityDisplayApp::mouseDrag( MouseEvent event )
{
	mPoints.push_back( event.getPos() );
}

void HighDensityDisplayApp::mouseWheel(MouseEvent event)
{
	mPoints.push_back(event.getPos());
}

void HighDensityDisplayApp::draw()
{
	gl::clear( Color( 0.1f, 0.1f, 0.15f ) );

	gl::color( 1.0f, 0.5f, 0.25f );
	gl::lineWidth( toPixels( 1.0f ) );
	gl::draw( mPoints );
}

CINDER_APP( HighDensityDisplayApp, RendererGl, prepareSettings )
```

PR #2156, opened 2020-04-10
Fix windows mousewheel

luksladky:
Fixing  #2155  ,  mouseWheel event on Windows with HiDPI display gives incorrect position


andrewfb:
Thanks for this find & fix, especially the repro case in the initial issue. I cherry-picked the commit. Fixes #2155.

Issue #1849, opened 2017-05-24
app stops updating when you press the alt key (windows)


q-depot:
Hi,

I've already reported this bug on the forum:
http://discourse.libcinder.org/t/alt-key-on-windows-pause-the-app/696/3

richardeakin:
I did some investigating into this and it appears that a hidden menu is taking focus of the messaging in our [`WndProc`](https://github.com/cinder/Cinder/blob/246075e1ed64050252df161dbb5dd3fe0d3f26a1/src/cinder/app/msw/AppImplMsw.cpp#L769). Using [Spy++](https://msdn.microsoft.com/en-us/library/dd460725.aspx), you can see the messages `WM_INITMENU`, `WM_MENUSELECT`, and then `WM_ENTERIDLE` repeatedly until you do something that I presume cancels the menu. It seems like we should be able to solve this by either programatically cancelling the menu, or somehow disable that menu in the first place. I don't yet know how to do either.

There are a few things we could try in [this msdn doc](https://msdn.microsoft.com/en-us/library/windows/desktop/ms647553(v=vs.85).aspx#menu_destruction) to destroy the menu, but there might be a cleaner solution, perhaps with some [window style](https://msdn.microsoft.com/en-us/library/windows/desktop/ff700543(v=vs.85).aspx)?

luksladky:
I've pushed a pull request fixing this. We need to be handling command SC_KEYMENU in WM_SYSCOMMAND message.

brucelane:
same for me, also with F10 key!

luksladky:
Should be fixed in this commit: https://github.com/cinder/Cinder/commit/adcd158eb80b76876ecff66669c44e5fba9543dd

Let's close the issue.

PR #2157, opened 2020-04-10
Fixing application freeze on Alt key press

luksladky:
Solving #1849 , alt key was invoking native Windows menu.


andrewfb:
Thanks for this as well - tested and confirmed. I cherry-picked this commit. Fixes #1849.

PR #2158, opened 2020-04-10
Added Visual Studio Code ecosystem to enable cross-platform building with Cmake

kino-dome:
As an alternative to CLion, this could be a nice setup to utilize libcinder's powerful CMake integration in order to enable cross-platform development. Tested on Windows 10 with C++ build tools 2019 and Ubuntu 18.0.4 with Clang and GCC. The PR includes:

- Added `.workspace` files to Cinder and samples at `proj/cmake`
- Added a hidden `.vs-code` folder to the root of Cinder and every sample. This will hold a `settings.json` file with configs needed for the system to work
- Updated ghc::filesystem to the latest version. This solved an issue with `wchar_t` when targeting C++11 or C++14
- Updated `videoInput` lib to the latest version and added 'videoInput.cpp' to the Cmake Configuration since it seemed to be missing
- Updated the Cmake guide
- Updated the Ubuntu guide to remove installation for `filesystem` and added `libjpeg` ( I used this in the work I'm doing on Capture class in Linux which I'll make a separate PR for.

Let me know your thoughts guys,
Hesam


kino-dome:
I wrote an in-depth guide in the CMake guide. You can read it here:
https://docs.google.com/document/d/1efJFciMs4iq5r-FAGN7FaatgaINL6Y3-yL7Cv72fe3Y/edit?usp=sharing

richardeakin:
Thanks for these additions, lots of stuff going on here. Concerning any library updates, can we keep them isolated in separate PR's? Also there are a growing number pending cmake file updates, so that would be good to keep separate too.

Concerning Visual Studio Code integration - is it necessary to add project files for each sample? When I tested with VSCode it was easy to build out of the box, what do these extra config files buy us? Issue is mostly about maintanence, it would be nice not to have to carry around the extra per-sample settings on top of CMake. Although perhaps a guide on best ways to use Visual Studio Code would be nice, maybe even in the official docs.

kino-dome:
Hey Rich, thanks for the notes. I totally agree with dividing this PR into a couple. It makes a lot more sense. Sorry, this is my first PR to Cinder and one of my first PRs ever and I still have a lot to learn :-) 

About the VS Code integration, I've already added some notes in the CMake guide under the part about CLion. The notes describe the installation process and building using VS Code and also talks about configuration etc. Do you think that will suffice? or is there any other point you think needs to be talked about?
As for your question about having project files for each sample, I think it will ease up the learning process and doesn't require newcomers to mess around with creating new project files. Pretty much like how there's a VS Studio project folder for each sample.
Regarding the different settings files, for each workspace file, the cpp extension needs to know that it should use CMake and there needs to be relative paths to where Cmake files are for that sample. That's all there is to it to be honest. Nothing that will need maintenance of any sorts I think. Also the `.vs-code` folder that contain the settings file is hidden so there's less visual noise on the bright side :-) 
All in all I think it's necessary to keep the files to help new users learn Cinder easier. And in a free and cross-platform fashion which I think is the strongest point of VS Code in this context.

Lastly, I have a question about isolating the updates to different PRs. The thing is that for the VS Code PR to work the PRs for Cmake config and the update to ghc::filesystem needs to be merged first. So it's kind of dependent on those. How can we signify this? Sorry maybe it's a more of a Git question than a Cinder-specific one. Cheers Rich!

MikeGitb:
Doesn't VSCode automatically pick up a CML in the root folder of the workspace? That's how VS does it.

kino-dome:
> Doesn't VSCode automatically pick up a CML in the root folder of the workspace? That's how VS does it.

If there's no settings file at the root of the workspace, CMake extension prompts and asks if it should tell cpp tools to use it as the build tool. Afterwards a settings file with this appears on the root:
`"C_Cpp.default.configurationProvider": "vector-of-bool.cmake-tools"`

What I've done here is to put this (along with two attributes about the relative path of the cmake file and cmake build path) in the settings file beforehand. This way it never prompts and annoys the users and everything is setup and ready to build. Here are all the attributes that should be in the root's settings file for the system to work:
`"C_Cpp.default.configurationProvider": "vector-of-bool.cmake-tools"`
`"cmake.sourceDirectory": "${workspaceFolder}/proj/cmake"`
`"cmake.buildDirectory": "${workspaceFolder}/proj/cmake/build"`

For Cinder itself, here are the needed attributes:
`"C_Cpp.default.configurationProvider": "vector-of-bool.cmake-tools`
`"cmake.configureSettings": {"BUILD_SHARED_LIBS":false}`

Here configureSettings is a convenience attribute to let the users provide CMake's configuration with custom parameters.

So, again, I think we really shouldn't expect the users (especially newcomers) to go through all this.

MikeGitb:
OK, so the problem is that the CMLs are not in the root of the sample folders, but instead in proj/cmake

@richardeakin, @andrewfb: Is there a reason for it? 

Imho it would make more sense to just move the CMLs to the sample apps root folders, where various tools expect it to be (including VSCode and VS and I would guess many others too) instead of adding a tool specific config file/folder for every IDE/Editor to point it to the correct subfolder.

kino-dome:
I get it that it's a convention to have the CMakeLists files on the root but at the same time I understand the possible logic behind moving them to a subfolder. I would guess the reason to be to clean up the root folder and organize build system files in their respective folders. Files like the CML and the build folder (with lots of subfolders) created by Cmake. Although it's a different case for libcinder itself where the CML and the build folder are on the root. 

Personally I don't mind this arrangement especially when having to deal with extra files related to Cmake. Like in the case of blocks. Or in this case where a `.workspace` file needs to be shipped with the samples. It does definitely look more "organized" in my view.

Also, here in regards to VS Code workspace integration, even having the CML on the root won't negate the need to have the 'settings.json' file in the folder. This is because of the necessity to tell the C++ extension to connect to CMake, with this line:
`"C_Cpp.default.configurationProvider": "vector-of-bool.cmake-tools"`

PR #2167, opened 2020-05-18
Cmake update :: Visual Studio 2019 + videoInput

kino-dome:
Based on  MSVC version, the right toolset and compiler options are extracted and CXX_FLAGS are written. Solved a massive headache with VS 2019 not defaulting to c++17.

Also by adding `videoInput.cpp` to the cmake definition for Windows and setting the proper include path for it, CaptureBasic sample builds fine.


kino-dome:
I just found out that the section about videoInput was already in PR #2060 with a small difference that in that PR to correct the include paths, `videoInput.cpp` is modified but in my case the include paths are added to `platform_msw.cmake`. I'm not sure if there's any tangible difference or will modifying the source file break anything but just wanted to point that out.

andrewfb:
Thanks for this fix.

PR #2170, opened 2020-05-26
Dear ImGui v177

num3ric:
Updated via python script + include fix to avoid macro duplication warnings.


vinjn:
Please help merge :)

num3ric:
The v1.77 update (9714d4e) unfortunately broke key press events (on windows at least).

num3ric:
Looks like this recent change fixes the issue:
https://github.com/ocornut/imgui/commit/eefae082617cd36d9b1d412e481a7913ab0ef3f2#diff-e5e00627776ba59572e7d1d86554aee6R3370

However, I'd suggest reverting the v1.77 update commit for now.

PR #2174, opened 2020-06-07
Remove trailing slash from CINDER_LIB_DIRECTORY

markkorput:
The trailing slash is causing a double slash in my ```cinderConfig.cmake``` and subsequent cmake failures for apps. Removing the trailing slash seems to solve the issue and is also more in line with general policy to **not** add trailing slashes to cmake path variables.

If this is really a bug I'm a bit surprised because this doesn't look like it's a recent change and I'd expect it to have been causing a lot of issues.

Tested using:
* OSX 10.15.5
* CMAKE version 3.17.3


andrewfb:
Thank you for this fix

PR #2175, opened 2020-06-08
Add sqr function to CinderMath.h

IntellectualKitty:


paulhoux:
What's the use case for this?

IntellectualKitty:
Using a squaring function makes for shorter, easier code.  It also more closely resembles the mathematical equation the code is performing.

For example, this seems shorter, simpler, and easier to read:

```
float distance = cinder::math<float>::sqrt( cinder::math<float>::sqr( x2 - x1 ) + cinder::math<float>::sqr( y2 - y1 ) );
```
than this:

```
float deltaX = x2 - x1;
float deltaXSqr = deltaX  * deltaX;
float deltaY = y2 - y1;
float deltaYSqr = deltaY  * deltaY;
float distance = cinder::math<float>::sqrt( deltaXSqr + deltaYSqr );
```

andrewfb:
I am up for this change if there are no objections

totalgee:
Just a minor point: what about calling it `sq()` or `squared()`, so there is no ambiguity about the `sqr` being "square root"? Or alternatively (better?), there is already a `length2()` in GLM (GLX) that does this -- squaring a scalar and doing `dot(v,v)` on vectors (which returns the squared length).

paulhoux:
As a side note:
```
float deltaX = x2 - x1;
float deltaXSqr = deltaX  * deltaX;
float deltaY = y2 - y1;
float deltaYSqr = deltaY  * deltaY;
float distance = cinder::math<float>::sqrt( deltaXSqr + deltaYSqr );
```
can be simplified to:
```
auto delta = vec2( x2 - x1, y2 - y1 );
auto distance = glm::length( delta );
```
or:
```
auto distanceSquared = glm::dot( delta, delta );
````

PR #2181, opened 2020-06-18
Missing iterator increment

bradhowes:
Increment iterator in the cases where the current node is not
removed from the list.


richardeakin:
Thank you for this find and fix! Merged into master with #2331.

richardeakin:
@bradhowes I realized there is a path that your PR covers and mine didn't, so I'm re-opening this and will merge yours instead. Thanks again!

PR #2182, opened 2020-06-18
Add coordinate accessors to Rect.h

IntellectualKitty:


IntellectualKitty:
The way that `cinder::RectT` is implemented is unnecessarily ambiguous and confusing.  Every time I write new code using it, I have to go back to the header file to reference the names of member variables with their meaning:

```
	Vec2T		getUpperLeft() const	{ return Vec2T( x1, y1 ); };
 	Vec2T		getUpperRight() const	{ return Vec2T( x2, y1 ); };
 	Vec2T		getLowerRight() const	{ return Vec2T( x2, y2 ); };
	Vec2T		getLowerLeft() const	{ return Vec2T( x1, y2 ); };
```

For example, x1 is left, y2 is bottom, etc.

Adding specific coordinate accessors based on their meaning removes the ambiguity and confusion.  And it also does so without breaking the existing interface.

PR #2186, opened 2020-08-07
osc: fix crash when closing TCP receiver connections

totalgee:
- replaced remove_if() with find_if(), because removing an unique_ptr
  from a vector causes it to become null, causing a crash later when
  we try to dereference it.

### DIFF
--- a/blocks/OSC/src/cinder/osc/Osc.cpp
+++ b/blocks/OSC/src/cinder/osc/Osc.cpp
@@ -1605,7 +1605,7 @@ asio::error_code ReceiverTcp::closeConnection( uint64_t connectionIdentifier, as
 		return ec;
 	
 	std::lock_guard<std::mutex> lock( mConnectionMutex );
-	auto rem = remove_if( mConnections.begin(), mConnections.end(),
+	auto rem = find_if( mConnections.begin(), mConnections.end(),
 	[connectionIdentifier]( const UniqueConnection &cached ) {
 		return cached->mIdentifier == connectionIdentifier;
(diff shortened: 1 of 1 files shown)


totalgee:
I came across this bug when trying to write a server using `osc::ReceiverTcp`, with multiple client connections. It took me a while to figure out what was going on; I would get crashes in the server when disconnecting a client. However, the issue here is that `std::remove_if` not only moves any matching elements (which in this case is a `unique_ptr<Connection>`) to the end of the vector, but also "moves from" it -- which for a unique_ptr means it's set to `nullptr`. So later, when it is dereferenced to call `shutdown()`, we get undefined behaviour (a crash, generally).

The rest of the code is fine -- the later call to `mConnections.erase(rem)` to actually eliminate the item from the vector of connections works as expected (and the Connection is now destroyed at that point, rather than during the `remove_if`).

The only potential issue with calling `find_if` instead of `remove_if` would be if we expected multiple connections in the vector with the same identifier, but there can't...it always increments by one for each new connection. So "find" is the appropriate idiom anyhow!

totalgee:
Here is [an example illustrating the bug](https://gist.github.com/totalgee/a19831f6b4ea5f249606af21ba1a1bf0) in a standalone "dummy" example.

totalgee:
This is to fix issue https://github.com/cinder/Cinder/issues/2192, which I just opened.

andrewfb:
Wow - that's a subtle bug. For tracking it down and providing a helpful explanation as well as the fix.

Issue #2142, opened 2020-02-21
Upgrade CMake


MikeGitb:
There are a few updates I'd like to make to the cmake files, in order to make it easier for the user (and or a package management system) to control the cinder build via typical cmake configuration options. Two examples being the control of the used c++ standard for which CMAKE provides the `CMAKE_CXX_STANDARD` variable since cmake 3.1 or the hardcoded usage of MT (https://github.com/cinder/Cinder/issues/2116)

I think that would also be a good time to upgrade the minimal required cmake version, as 2.8 is really old and a lot of things have become easier since then.

I wanted to ask, what an acceptable baseline would be. To give a short overview: 
VS2019 currently ships with 3.15, Ubuntu 18.04 with cmake 3.10, Ubuntu16.04 with 3.5. Travis uses 3.12. I have no idea about macos, but generally it is very easy to install a newer version of cmake on any system.

Depending on whether Ubuntu 16.04 is still a common target, I'd recommend 3.5 or 3.10 (among other things, cmake 3.8 adds the cxx_std_14 / 17 target properties which are "nice to have" but not critical).

ufidstudios-ch:
I can tell you Android NDK uses 3.10.2, so no objections on that side.

kino-dome:
This could definitely be a plus.

MikeGitb:
@andrewfb , @richardeakin: Ok, so, would 3.10 be fine for you?

tmtmazum:
Is it strictly necessary to change the CMAKE_MINIMUM_VERSION in order to start using the newer features?

You could add a cmake version guard such as this around the use of the optional features:
```
    if( ${CMAKE_VERSION} VERSION_GREATER "3.8.2" )
        target_compile_features( cinder PRIVATE cxx_std_17)
    endif()
```
This would allow use of the newer features for those who have the latest cmake, without breaking the build for folks who have older cmake.

Otherwise there would always be a lag behind cmake features - i.e. if you settle on 3.10, then you're unable to use any features newer than that.

MikeGitb:
For one, I see little use in using the new features if I have to keep the old code around - just makes the code harder to understand and maintain. Second cmake_minimum_required not only checks the cmake version it also makes sure that newer versions behave as close as possible to the minimum cmake version specified by setting all kinds of policies.

 In turn, newer cmake versions start to spew out warnings when cmake_minimum_required is too old and hence is forced to behave in a way that might be suprising to the user.

[...]

PR #2190, opened 2020-08-21
[CMake] Set minimal required version to 3.10

MikeGitb:
In preparation of further refactorings.
See https://github.com/cinder/Cinder/issues/2142


andrewfb:
Thanks for this work

PR #2195, opened 2020-09-01
[CMake] Fix setting msvc link directories in make app

MikeGitb:
- CINDER_PATH should  be ARG_CINDER_PATH (argument passed to function instead of global variable)
- No apparent reason, why policy CMP0015 should be set to old.
  Remove this to silince cmake warnings about using outdated policies.


richardeakin:
Sorry for the long wait on this, both changes look good to me. Thanks!

PR #2196, opened 2020-09-01
[UnitTests] Fix unit tests with cmake and msvc

MikeGitb:
The test app is created via ci_make_app, which
creates a win32 application instead of a simple console app.
This requires a WinMain entry point instead of the main
provided by catch.


MikeGitb:
Rebased on master

It would be great, if this PR could get merged or if could get some feedback.

andrewfb:
Confirming this is not breaking other platforms (including MSVC via non-CMake), so I'm merging it (but have not tested it for CMake on MSW)

PR #2197, opened 2020-09-01
Get rid of most warnings when building OSC block

totalgee:
- these were particularly verbose on Windows, mainly related to possible loss of data during size conversions (`size_t` to `int32_t` or `uint32_t`).
- there are still a few other warnings from ASIO, but fixing those may require replacing/updating the ASIO library.

I also added asserts (for extra safety/correctness), but if you don't like those they can be removed. We can "probably" assume no one will try to send OSC messages, arguments or blobs that are more than 2GB or 4GB in size, but just in case... (-;

*(unrelated)* I also fixed a bug in the `Rand` unit test, where I assume it was intending to only run a performance test in Release builds, but was instead doing it in Debug (which took forever). At least on Windows, `DEBUG` wasn't defined in the project (although `_DEBUG` is) -- but the proper test is normally `#ifdef NDEBUG` if you want optimized (Release) builds. This should work across all platforms.

### DIFF
--- a/blocks/OSC/src/cinder/osc/Osc.cpp
+++ b/blocks/OSC/src/cinder/osc/Osc.cpp
@@ -312,7 +312,9 @@ void Message::append( const std::string& v )
 	mIsCached = false;
 	auto trailingZeros = getTrailingZeros( v.size() );
 	auto size = v.size() + trailingZeros;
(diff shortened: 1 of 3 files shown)


andrewfb:
Thanks for these, including the NDEBUG fix.

PR #2199, opened 2020-09-03
[CI] Update travis CI to use only supported ubuntu images

MikeGitb:
The oldest ubuntu distro, compiler and c++ standard and cmake version supported by cinder have changed quite a bit. 
Also travis itself has changed the environments it provides and what options are available. 
This PR cleans up the travis-ci file and tests cinder only with the default g++ and clang++ compiler on ubuntu bionic and focal respectively.

The main goal here is to make the ci-results usable again for development. Extending the test matrix to more compilers and more standard settings is left for a future PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,22 +90,23 @@ matrix:
         - TOOL="xcode-ios-sim"
         - DESCRIPTION=OS X build/test via Xcode"
 
-    # OS X CMake
+    #OS X CMake
     - os: osx
       osx_image: xcode9.2
       env:
         - TOOL="cmake"
         - DESCRIPTION="OS X build/test via CMake"
-    - os: osx
-      osx_image: xcode8.3
-      env:
-        - TOOL="cmake"
-        - DESCRIPTION="OS X build/test via CMake"
-    - os: osx
-      osx_image: xcode7.3
-      env:
-        - TOOL="cmake"
-        - DESCRIPTION="OS X build/test via CMake"
+    # TODO: find fast way to install cmake 3.10 on these systems
+    #- os: osx
+    #  osx_image: xcode8.3
+    #  env:
+    #    - TOOL="cmake"
+    #    - DESCRIPTION="OS X build/test via CMake"
+    #- os: osx
+    #  osx_image: xcode7.3
+    #  env:
+    #    - TOOL="cmake"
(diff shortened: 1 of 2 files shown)


MikeGitb:
TODO: also fix OSX runs

MikeGitb:
Note: This PR also deactivates two test on macos with older xcode versions, as the cmake version there is 3.9. As I don't have a mac, I don't know what the fastest way to install a newer cmake version would be. Using `brew upgrade cmake` seems possible, but that leeds to excessive build times as it first also installs several dependencies.

andrewfb:
Thanks for this update

ufidstudios-ch:
I'd be happy to set up a Bitrise CI integration, they've offered us a free plan:

https://blog.bitrise.io/free-developer-plan-features-for-open-source-projects-on-bitrise

ufidstudios-ch:
The reason for insisting on CMake 3.10.2 is that it's what is bundled with Android Studio, so it's available on Mac, PC and Linux.

MikeGitb:
> it's what is bundled with Android Studio, so it's available on Mac, PC and Linux.

Only when android studio is installed. People like me don't write Software for Android ;). Well,  I'm using latest cmake anyway but I've no Idea what the typical MacOS dev environment looks like.

ufidstudios-ch:
Right. But the object of this exercise is to make it as easy as possible for people who do. 

Ideally, and Android developer wants to install Android Studio and nothing else, and be able to build Cinder from source, and deploy it to the Play Store. This is a normal use case.

ufidstudios-ch:
Anyway. You don't need a mac or an android device, because bitrise now runs a CI pipeline which checks everything.

https://app.bitrise.io/build/88e8d685acd39547#?tab=log

With regard to toolchains - I don't know a single Android developer who doesn't use Android studio. It's theoretically possible with a lot of expensive and time consuming custom scripting, but this is best avoided if at all possible. We want Cinder for Android to work as easily as possible, out of the box.

ufidstudios-ch:
@MikeGitb Just out of interest, do you mainly write for Windows & Linux? I'm keen to cover those bases too.

MikeGitb:
I think we should move this discussion back to https://github.com/cinder/Cinder/pull/2208 (wrote a comment there) I only linked this PR, because the question of how to install recent versions of cmake on mac came up when I created this PR too.

Issue #2193, opened 2020-08-27
ImGui right shift key stays "stuck" (acts as toggle)


totalgee:
Hi there. When I use an `ImGui::InputTextMultiline`, the left shift key works as expected (only works while the key is held down). However, the right one acts as a toggle, which is really annoying. You can see this by typing something with a (right) shift in it, then using the arrow keys to move the cursor. You'll see you are highlighting selected text, because it things shift is still held down. You need to press it again to toggle off.

It makes using the multiline widget as a code editor pretty much useless. I'm not sure if this might be a bug (or "feature") in ImGui itself, but I haven't found anything related to it in the repo there.

totalgee:
I should clarify that it doesn't act the same as shift-lock, in that typed letters *don't* come out uppercase. But it *does* act as though shift were down for the purposes of selection, when you move the cursor.

totalgee:
This is **not** to do with ImGui, after all. If I write a minimal Cinder app and just overload `keyDown()` and `keyUp()`, I can see the problem. Just print a different message for L/R shift down and up, and you'll see it... E.g.

```c++
void MyApp::keyDown(KeyEvent e)
{
    if (e.getCode() == KeyEvent::KEY_LSHIFT) console() << "Left shift down" << endl;
    if (e.getCode() == KeyEvent::KEY_RSHIFT) console() << "Right shift down" << endl;
}

void MyApp::keyUp(KeyEvent e)
{
    if (e.getCode() == KeyEvent::KEY_LSHIFT) console() << "Left shift up" << endl;
    if (e.getCode() == KeyEvent::KEY_RSHIFT) console() << "Right shift up" << endl;
}
```

Now run your app, and press and release right shift several times. You'll see:

```txt
Right shift down
Right shift up
Right shift down
Left shift up
```

I think the bug is in `prepNativeKeyCode()` in `AppImplMsw.cpp`, but not sure of the fix yet. Maybe Windows only gives you "shift" events (not left/right-specific, which you need to poll for with `GetKeyState()`), it looks like, that's maybe why it was written the way it was...? @andrewfb any idea?

PR #2202, opened 2020-09-04
ImGui: fix "stuck" right shift key in InputText on Windows (#2193)

totalgee:
The way modifier key state was sent to ImGui seemed flaky (it failed on Windows in the right-shift case because of the issue I describe in #2193); there are already `KeyEvent` tests that work well to indicate which modifiers are down.

This fixes the issue with `InputText` (and `InputTextMultiline`) on Windows, and should also be valid for Mac (even if it doesn't specifically fix an issue there).


totalgee:
Thanks for merging; very happy to have this in!

PR #2203, opened 2020-10-02
Improves doc.

Divyanshu9151:


richardeakin:
Love the opinionated enthusiasm. :)

andrewfb:
We appreciate the compliment :) but will elect to leave this copy as written.

PR #2209, opened 2020-11-15
Adding install dependencies script for Fedora Linux

violetcraze:
Hello! 🌸

Seeing if I can get cinder working on Fedora Linux. Just spent about an hour tracking down all the equivalent packages from the apt script and figured I'd make a pull request with the resulting script. Don't know if it'll be helpful but figured I'd offer it up.

Not sure if missed any packages so I'm marking this pr as a draft in case I need to add some later in the process. Also don't have a clean install to test this at the moment. (I already have a ton of gstreamer packages installed from things like openFrameworks)

Couldn't find the equivalent of `gstreamer1.0-alsa` and `gstreamer1.0-pulseaudio`.

Happy to make changes if you like. 💖

### DIFF
--- a/tools/linux/scripts/trusty-install-deps-fedora.sh
+++ b/tools/linux/scripts/trusty-install-deps-fedora.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+set -e
+
+# Packages
+sudo dnf -y upgrade
+sudo dnf -y install \
+  libXcursor-devel \
+  libXrandr-devel \
+  libXinerama-devel \
+  libXi-devel \
+  mesa-libGL-devel \
+  libglvnd-devel \
+  zlib-devel \
+  expat-devel \
+  freetype-devel \
+  uuid-devel \
+  libsndfile-devel \
+  pulseaudio-libs-devel \
+  alsa-lib-devel \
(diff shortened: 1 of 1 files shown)


violetcraze:
Removed draft setting because Cinder seems to be working fine so I don't think there's any missing libraries.

andrewfb:
Hi - currently Ubuntu is our only officially supported Linux distribution, and unfortunately given the testing burden new platforms represent, that's unlikely to change. That said, if you found an issue which prevented Cinder from running correctly on Fedora we'd be interested to hear about it and would almost certainly accept a PR to that effect. Will close this for now.

richardeakin:
I think something like this would be suitable as an addition to the [linux guide](https://libcinder.org/docs/guides/linux-notes/index.html), at least to ease the process of getting cinder building on other linux distros. Or even easier, a simple forum post with your findings and build tweaks on the Fedora OS would help others who are trying to do the same.

violetcraze:
Thanks for considering! I did take @richardeakin and post on the Cinder forum. Here's a link for anyone that might come across this in the future: https://discourse.libcinder.org/t/fedora-linux-install-dependencies-script/1742.

PR #2212, opened 2020-11-20
Replace std::result_of with decltype

### LINKED ISSUE #2206: C++20 support on Windows
With GCC 10.2 on linux, I can compile with C++20. It seems to have kept backward compatibility, despite the C++20 standard not requiring it.
With Clang or VS2019 on Windows 10, compilation fails. There a couple of other issues in addition to result_of.

You can check here (godefv@f33b943) what I have done to successfully compile with C++20. This commit is on top of this PR.

(see PR #2133 for C++17 with VS2019)

bjude:
std::result_of is deprecated in C++17 and removed in C++20, breaking downstream projects that use C++20 with MSVC

This could be replaced by a wrapper type trait that dispatches to std::invoke_result_t for C++17 and later or to std::result_of for
earlier versions but since it is only used in one place for nullary functors, decltype will be simpler.

The vendored version of boost::asio uses std::result_of so this PR will not allow Cinder to be built in C++20 mode yet, but since asio is only used internally and not in Cinder's interface it should allow downstream project to use C++20.

Fixes #2206

### DIFF
--- a/include/cinder/app/AppBase.h
+++ b/include/cinder/app/AppBase.h
@@ -408,7 +408,7 @@ class CI_API AppBase {
 	void	dispatchAsync( const std::function<void()> &fn );
 	
 	template<typename T>
-	typename std::result_of<T()>::type dispatchSync( T fn );
+	decltype(std::declval<T>()()) dispatchSync( T fn );
 
 	//! Returns the default Renderer which will be used when creating a new Window. Set by the app instantiation macro automatically.
 	RendererRef	getDefaultRenderer() const { return mDefaultRenderer; }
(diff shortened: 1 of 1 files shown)


bjude:
Pinging @andrewfb for a review, I'd prefer to keep my project tracking the master branch instead of my fork and the issues this PR and #2211 are blocking that

andrewfb:
I apologize for the crazy delay on merging this - thanks for submitting it.

godefv:
Thanks for fixing my issue. I also prefer to work with the official branch rather than my fork !

PR #2214, opened 2020-12-14
Support for EXR grayscale images

gaborpapp:


richardeakin:
Nice add! I didn't run the code (don't have a grayscale EXR file either), but it all looks good to me.

I did see that the target side of things already supports Y / YA channel layouts so all good there, guess we just forgot to add it on the source side at the time of adding this.

gaborpapp:
Thanks for reviewing it.
Yes, the target side already works. I tested it with saving and loading `Channel32f`.

Issue #2216, opened 2021-01-17
Update Asio to 1.18?


morphogencc:
Is there any reason that the asio block is still on the 1.11.0 version?  The current stable release is 1.18, and some features/functionalities from the 1.11 have been deprecated in the older releases, and uses features that are deprecated in C++17 (specifically, members of `std::allocator` and `std::result_of` are deprecated).

The looked up past Asio version upgrades and found this single issue about it : https://github.com/cinder/Cinder/pull/1668  This issue seems to suggest that cinder wants to use asio's dev branch, but it doesn't look like they have a development branch anymore.

I'm happy to do the legwork to update the library and any dependencies, but wanted to check in before I started doing the work myself.

Alternatively -- I believe that if the app's `asio::io_service` instance was removed, it'd be possible for end users to include their own version of asio.  Currently the fact that `asio::io_service` is included by default as a part of `AppBase` causes problems because `io_service` has been deprecated in newer versions of asio.  It doesn't look like any samples or classes use this `io_service` instance, so it could be relatively straightforward to remove it.

andrewfb:
Hi - there's no technical reason we're on a previous version - would gladly accept a PR that upgrades us.

I haven't tracked with the more recent ASIO design changes but the value of an app-maintained `io_service` is a convenient pattern in my experience. Is there some similar concept that has replaced it?

morphogencc:
@andrewfb Absolutely; it's been replaced by the `io_context` class that's pretty much a drop-in replacement.  I'll get started on the change and I'll submit a PR when I'm finished!

PR #2217, opened 2021-01-19
Update asio from v1.11 to v1.18.1

morphogencc:
In response to issue #2216.

* replaced files in `include/asio` with the latest version of asio (v1.18.1)
* Updated the following files to use the new `io_context` object that replaces the deprecated `io_service`:
  * `AppBase.h`
  * `AppBase.cpp`
  * `Osc.h`
  * `Osc.cpp`
* Updated the following examples to use the new method `get_io_context()` of `AppBase`:
  * `osc/BroadcastSender`

All changes were tested with the samples in `osc/samples`.


andrewfb:
Thanks a lot for this. I have tested this on both MSW and macOS via the OSC samples.

I'd be up for merging this but want to give a little opportunity for anyone to speak their piece since it's a reasonably large change. Assuming there are no objections I think we can merge this soon.

morphogencc:
Sounds great; thanks!

PR #2218, opened 2021-01-25
Add a boolean to cinder::gl::Fbo::readPixels8u and cinder::gl::Fbo::r…

willemkempers:
…eadPixels32f to enable or disable the alpha channel in the returned Surface8u or Surface32f.

### DIFF
--- a/include/cinder/gl/Fbo.h
+++ b/include/cinder/gl/Fbo.h
@@ -171,10 +171,10 @@ class CI_API Fbo : public std::enable_shared_from_this<Fbo> {
 	//! Sets the debugging label associated with the Fbo. Calls glObjectLabel() when available.
 	void				setLabel( const std::string &label );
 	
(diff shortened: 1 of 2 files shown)


willemkempers:
This doesn't seem to work on a Raspberry Pi 4 running Raspbian GNU/Linux 10 (buster) unfortunately. Using `GL_RGBA` works fine but using `GL_RGB` results in garbage. I've found these threads on the topic: https://www.raspberrypi.org/forums/viewtopic.php?t=109474 and https://www.raspberrypi.org/forums/viewtopic.php?f=68&t=101903.

I'd personally still really like to see something along these lines - I think it could be really useful in combination with LibOLA for example. I'd love to see the readPixels method be expanded to be able to return a single ChannelT, rather than a SurfaceT. Would be very useful for outputting to white LEDs.

Issue #2222, opened 2021-01-31
Exr writing broken


gaborpapp:
No matter what surface channel order is set, the blue and alpha channels seems to be broken.

The following code
```
Surface32f test( 2, 2, true, SurfaceChannelOrder::ABGR );
test.setPixel( ivec2( 0, 0 ), ColorAf( 1, 2, 3, 4 ) );
test.setPixel( ivec2( 1, 0 ), ColorAf( 5, 6, 7, 8 ) );
test.setPixel( ivec2( 0, 1 ), ColorAf( -1, -2, -3, -4 ) );
test.setPixel( ivec2( 1, 1 ), ColorAf( -5, -6, -7, -8 ) );

writeImage( getAppPath() / "test.exr", test );

Surface32f test2 = loadImage( loadFile( getAppPath() / "test.exr" ) );
app::console() << test2.getPixel( ivec2( 0, 0 ) ) << std::endl;
app::console() << test2.getPixel( ivec2( 1, 0 ) ) << std::endl;
app::console() << test2.getPixel( ivec2( 0, 1 ) ) << std::endl;
app::console() << test2.getPixel( ivec2( 1, 1 ) ) << std::endl;
```

generates this output:
```
[1,2,-9.47913e+08,2]
[5,6,-9.47913e+08,6]
[-1,-2,-9.47913e+08,-2]
[-5,-6,-9.47913e+08,-6]
```

PR #2224, opened 2021-02-05
Fixes Exr invalid channel write

gaborpapp:
The vector storing the channel data should have a reserved size otherwise the pointers might get invalid if the vector is resized.

### DIFF
--- a/src/cinder/ImageFileTinyExr.cpp
+++ b/src/cinder/ImageFileTinyExr.cpp
@@ -284,6 +284,7 @@ void ImageTargetFileTinyExr::finalize()
 {
 	// turn interleaved data into a series of planar channels
 	vector<Channel32f> channels;
+	channels.reserve( mNumComponents );
 	unsigned char *    imagePtr[4];
 	for( int c = 0; c < mNumComponents; ++c ) {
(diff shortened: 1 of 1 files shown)


richardeakin:
Huh, well I guess when the vector resize gets called, the Channel [copy constructor](https://github.com/cinder/Cinder/blob/master/src/cinder/Channel.cpp#L141-L144) is getting called instead of the move, which is what I thought C++11 was all about. We can merge this PR to fix the problem, but it also seems like we need to figure out how to enable the move constructor for `vector<ChannelT>`s..

Also only just realized that this PR fixes #2222 (I think it didn't get auto linked because of the 'cinder' prefix in your commit comment).

gaborpapp:
I checked this in Linux and macOS too. GCC and Clang both wrote broken exr's without this change. So it's probably not a compiler specific issue.

richardeakin:
I guess worth noting that we should look into making Channel and Surface classes movable when held in `std::vector`, but as I don't have the time to look into that at the moment will merge this as it is a clear bugfix for the current version.

Thanks for checking on other platforms, good find!

PR #2227, opened 2021-02-28
macOS & Xcode version requirements update

andrewfb:
This PR modernizes our macOS build requirements to:
• macOS 10.13 (High Sierra) deployment target
• Xcode 11.3.1, implying macOS 10.14 (Mojave) as the baseline for development
• C++17

The primary motivation for this update is to better reflect the subset of platforms which are being regularly being used and tested. More advanced users can downgrade these requirements if necessary. The reasoning for selecting 10.13 as the baseline deployment OS is that it is the most recent version which is no longer supported by Apple (as of Dec 2020). Similarly, 10.14 as a baseline for development was selected as it is the oldest version still supported by Apple.

If these requirements do not reflect your active deployment or development baselines please weigh in here.


gaborpapp:
Thanks for this. I still need #2221 when built with CMake on Catalina 10.15.7, though.

PR #2228, opened 2021-03-01
OSMesa/glad headless fix

PetrosKataras:
Fixes a crash that manifested when linking against a packaged version of OSMesa and was caused by improper initialisation of the appropriate GL function pointers.

It also adds a more modern version of the `FindOSMesa.cmake` script for finding and linking against the library.

Tested on Ubuntu 20.04 with `libosmesa6-dev` ( Mesa 20.2.6 )


andrewfb:
Thanks for this Petros - tested on both Ubuntu 18.04 and 20.04 with default packages and a built-from-source Mesa.

Issue #2234, opened 2021-04-03
Build doesn't work on arm64 / M1


nside:
I had to change the following to build on arm64
```
diff --git a/proj/cmake/platform_macosx.cmake b/proj/cmake/platform_macosx.cmake
index db8005ae3..fa164282f 100644
--- a/proj/cmake/platform_macosx.cmake
+++ b/proj/cmake/platform_macosx.cmake
@@ -152,7 +152,7 @@ if( NOT ( "Xcode" STREQUAL "${CMAKE_GENERATOR}" ) )
                message(STATUS "Found libtool - ${CMAKE_LIBTOOL}")
                get_property(languages GLOBAL PROPERTY ENABLED_LANGUAGES)
                foreach(lang ${languages})
-                       set(CMAKE_${lang}_CREATE_STATIC_LIBRARY "${CMAKE_LIBTOOL} -static -arch_only x86_64 -o <TARGET> <LINK_FLAGS> <OBJECTS> ")
+                       set(CMAKE_${lang}_CREATE_STATIC_LIBRARY "${CMAKE_LIBTOOL} -static -arch_only arm64 -o <TARGET> <LINK_FLAGS> <OBJECTS> ")
                endforeach()
        endif()
 endif()
diff --git a/src/cinder/System.cpp b/src/cinder/System.cpp
index 36bec8727..128d77b71 100644
--- a/src/cinder/System.cpp
+++ b/src/cinder/System.cpp
@@ -455,7 +455,7 @@ int System::getOsMajorVersion()
                instance()->mOSMajorVersion = [[sysVerComponents firstObject] intValue];
 #elif defined( CINDER_MAC )
                if( [[NSProcessInfo processInfo] respondsToSelector:@selector(operatingSystemVersion)] ) {
-                       ShadowOSVersion version = ((ShadowOSVersion(*)(id, SEL))objc_msgSend_stret)([NSProcessInfo processInfo], @selector(operatingSystemVersion));
+                       ShadowOSVersion version = ((ShadowOSVersion(*)(id, SEL))objc_msgSend)([NSProcessInfo processInfo], @selector(operatingSystemVersion));
                        instance()->mOSMajorVersion = (int32_t)version.majorVersion;
                } else {
        #pragma clang diagnostic push
@@ -487,7 +487,7 @@ int System::getOsMinorVersion()
                instance()->mOSMinorVersion = [[sysVerComponents objectAtIndex:1] intValue];
 #elif defined( CINDER_MAC )
[...]

PR #2235, opened 2021-04-03
Fix macosx build on arm64

nside:
https://github.com/cinder/Cinder/issues/2234


Morpheu5:
I verified that this produces a working build for macOS on Apple Silicon (arm64). I built this with Xcode 12.5.1 (12E507).

Merging would be great 😃

andrewfb:
Thanks very much for this PR - I am not in a position to test this, but it makes sense.

PR #2237, opened 2021-04-10
Make Ref arguments `const` for ScopedVao and ScopedGlslProg

totalgee:
Minor change, just to be consistent with other scoped Ref types, and allowing const `draw()` methods to use these scoped objects.

### DIFF
--- a/include/cinder/gl/scoped.h
+++ b/include/cinder/gl/scoped.h
@@ -42,7 +42,7 @@ typedef std::shared_ptr<Sampler>			SamplerRef;
 
 struct CI_API ScopedVao : private Noncopyable {
 	ScopedVao( Vao *vao );
-	ScopedVao( VaoRef &vao );
+	ScopedVao( const VaoRef &vao );
 	~ScopedVao();
 
   private:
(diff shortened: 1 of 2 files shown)


totalgee:
Bump...? 😉

richardeakin:
I believe there was a specific reason for this though I can't remember exactly what that was. Something @andrewfb found as an optimization when there were many many `ScopedVao`s and `ScopedGlslProg`s (I think it was related to the Earth sample).

totalgee:
Actually, these changes shouldn't affect anything to do with whether the Scoped object (GlslProg or Vao) has const or non-const access, because this is just the `shared_ptr` (the "Ref") we're talking about. I see that some of the Scoped helpers take a const pointer (GlslProg) vs a non-const (Vao), but this won't be affected by this PR. (also, `ScopedGlslProg` actually takes a `shared_ptr` to a *const* GlslProg). Anyhow, it's not super important and I do get what you're saying @richardeakin , but the truth is that making the `shared_ptr` argument itself const reference or not should make no difference, since the constructor then takes a pointer to the pointed-to object using `get()`. Do you want to integrate this, or should I retract it?

Issue #2238, opened 2021-04-24
Camera method `getInverseViewMatrix()` returns out-of-date value


totalgee:
If you have a `Camera` and modify its transform (by calling `lookAt()`, `setEyePoint()`, `setViewDirection()`, etc.) and then immediately call `getInverseViewMatrix()`, you don't get the correct matrix. This normally didn't show up as an issue because (presumably) many places call `getViewMatrix()` first, and many internal methods call `calcMatrices()`.

totalgee:
Fixed by #2239.

PR #2239, opened 2021-04-24
Fix inverse viewing matrix caching bug in Camera class

totalgee:
- add method dirtyViewCaches() to dirty both viewing and inverse
  viewing matrix caches whenever the camera transform is modified.

Fixes bug #2238.


totalgee:
Any word on this one?

andrewfb:
Apologies for such an unreasonable delay on this - thanks for this find & fix

Issue #2241, opened 2021-05-03
ImGui glitches when using multiple Windows


totalgee:
See https://discourse.libcinder.org/t/multiple-windows-and-interop-with-imgui/1819

In an application with multiple Windows, even if only one of them should be using ImGui, things get glitchy (the UI starts jumping around) when you have multiple windows, especially notable when the two are on different displays.

totalgee:
Resolved by #2242 and #2243.

PR #2242, opened 2021-05-03
Fix glitches when using ImGui in an app with multiple windows

totalgee:
- two calls to `toPixels()` were using the default `App` version (which
  uses the "current" Window to do the conversion), but the current
  Window is indeterministic when in an `update` handler, so with
  multiple windows, it is likely to reference the wrong one.
- fixes Cinder issue #2241.

### DIFF
--- a/src/cinder/CinderImGui.cpp
+++ b/src/cinder/CinderImGui.cpp
@@ -457,7 +457,7 @@ static void ImGui_ImplCinder_NewFrameGuard( const ci::app::WindowRef& window ) {
 	IM_ASSERT( io.Fonts->IsBuilt() ); // Font atlas needs to be built, call renderer _NewFrame() function e.g. ImGui_ImplOpenGL3_NewFrame() 
 
 	// Setup display size
-	io.DisplaySize = ci::app::toPixels( window->getSize() );
+	io.DisplaySize = window->toPixels( window->getSize() );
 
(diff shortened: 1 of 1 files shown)


totalgee:
Note that there are some other calls to `toPixels()` in the Cinder ImGui code, e.g. in the various mouse and keyboard Event handlers. These seem to work fine (so I didn't change them, though they could easily be fixed by using `event->getWindow()->toPixels()`). Presumably the current window is correct in these callbacks, since you're directly interacting with them... (let me know if you want me to change these, too, I initially opted for the minimum fix that corrected my issue ;-)  This also seems to be the case with the resize handler, although there we have no direct way to get the window, but it could easily be "fixed" by binding the window in the signal handler, similar to what is done for the update signal handler. Again, let me know if you want that change to be made.

richardeakin:
Nice find, this makes sense to me.

richardeakin:
I think if it makes sense to you for these other winder->toPixel changes to happen it would be more robust and can review it, though for now happy to merge this.

totalgee:
> I think if it makes sense to you for these other winder->toPixel changes to happen it would be more robust and can review it, though for now happy to merge this.

Okay, I decided to make these changes, since I was in the code, and if I don't do it now I'll forget about it. See #2243 (this one can stay closed). Thanks!

PR #2243, opened 2021-05-10
Make ImGui multi-window more robust; fix issues noted in #2242

totalgee:
Just adding the rest of the fixes I'd suggested earlier. They aren't strictly "necessary" (since the current window seems to be correct when the event and resize callbacks are called), but are cleaner and more consistent.


totalgee:
@richardeakin Any chance this could get merged, so the whole (related) topic can be closed? Thanks!

richardeakin:
Sure, and thanks for getting these changes in. Apologies for the delay, I've been deep in work work and hoping to emerge soon to address more of these outstanding PRs..

PR #2245, opened 2021-05-27
Logging updates

num3ric:
Surfaced log level getter & setter to the logger base class to allow changes at runtime across all loggers. Also added a `limitDirectoryFileCount` function in `Utilities.h` to remove log files exceeding a file count.


num3ric:
Note that this is a (minor) breaking interface change with the previous `setTriggerLevel` and `setLoggingLevel` functions superseded by the base `setLevel` function.

PR #2253, opened 2021-09-29
Fix to Screensaver on MacOS

timmb:
This commit fixes an error preventing new screensaver projects from rendering on MacOS.

It implements a solution found by athens and shared in [this forum post](https://discourse.libcinder.org/t/screensaver-on-osx-just-black-screen/1686/4).


richardeakin:
So, I guess this extra `[mCinderView draw]` is needed now. I'm not in a position to test Mac OS screensavers, but as it appears that multiple people found this to be a necessary change I'm inclined to merge this PR.

Thanks for submitting and apologies for the delay!

PR #2266, opened 2022-01-05
Fixes copyRawRgbFullAlpha incorrect src offset

chaoticbob:
The constant for src should be 3 instead of 4 since the src surface is
Rgb and not Rgba.

### DIFF
--- a/src/cinder/Surface.cpp
+++ b/src/cinder/Surface.cpp
@@ -477,7 +477,7 @@ void SurfaceT<T>::copyRawRgbFullAlpha( const SurfaceT<T> &srcSurface, const Area
 	int32_t width = srcArea.getWidth();
 
 	for( int32_t y = 0; y < srcArea.getHeight(); ++y ) {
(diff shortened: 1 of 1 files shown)


src/cinder/Surface.cpp
```diff
@@ -477,7 +477,7 @@ void SurfaceT<T>::copyRawRgbFullAlpha( const SurfaceT<T> &srcSurface, const Area
 	int32_t width = srcArea.getWidth();
 
 	for( int32_t y = 0; y < srcArea.getHeight(); ++y ) {
-		const T *src = reinterpret_cast<const T*>( reinterpret_cast<const uint8_t*>( srcSurface.getData() + srcArea.x1 * 4 ) + ( srcArea.y1 + y ) * srcRowBytes );
+		const T *src = reinterpret_cast<const T*>( reinterpret_cast<const uint8_t*>( srcSurface.getData() + srcArea.x1 * 3 ) + ( srcArea.y1 + y ) * srcRowBytes );
```
totalgee, line 480: Just wondering: to be consistent, should this be using `srcPixelInc` and `dstPixelInc` as some of the surrounding methods do? Or do those not give the right result here?

Issue #2272, opened 2022-02-10
Mysterious crash when printing to console on Windows (with specific size message)


totalgee:
Whoa, I just spent some time debugging a mysterious intermittent crash in my program... It was crashing when printing to the console...very rarely.

I just figured out that the issue is here:

https://github.com/cinder/Cinder/blob/e83f5bb9c01a63eec20168d02953a0879e5100f7/src/cinder/msw/CinderMsw.cpp#L146-L152

(the check `if (wideSize == ERROR_NO_UNICODE_TRANSLATION)` is incorrect)

The return value from `MultiByteToWideChar()` is supposed to be [0 on an error](https://docs.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-multibytetowidechar#return-value), after which you're supposed to call `GetLastError()` if you want a more detailed error code (one of which can be `ERROR_NO_UNICODE_TRANSLATION`).

I was convinced there was a bug or some kind of memory corruption in my code, but I think this is likely a long-standing bug in Cinder (maybe never encountered by anyone else...it only throws/crashes when you happen to try to log a string that is exactly 1113 bytes long :wink:).

For the fix, I will just remove that first `if` check, and throw the error from the current `else` clause (could explicitly check `GetLastError()`, but probably not much more benefit here). I will submit a PR.

PR #2273, opened 2022-02-10
Fix mysterious console logging crash

totalgee:
- an error would be thrown when trying to log a message to the
  console that was of length 1113 bytes.
- fixes issue #2272.


totalgee:
Yes, lucky number 11**13**...  Thanks for merging!

PR #2276, opened 2022-02-20
Added optional ${CINDER_IMGUI_DIR} cmake variable to build with an external version of ImGui

simongeilfus:
This PR allows to ship _Cinder_ along with an external version of _DearImGui_. 

It's great to have a script to update the version cinder is shipping with but it's a far less convenient solution when trying to setup a repository with a custom version of imgui. When working with a team it is easier to rely on submodules and cmake than having to depend on users to execute a python script.

The cmake changes are minimal but there's a couple of small required changes on the cpp side. Mostly the addition of a new `CINDER_IMGUI_EXTERNAL` preprocessor define to switch between the cinder friendly `#include "imgui/imgui.h"` and the standard `#include "imgui` used by _DearImGui_ (and all libraries built upon _DearImGui_). This is taken care of by appending to the `CINDER_DEFINES` cmake variable.

This PR also removes the unused `#include "imgui/imgui_stdlib.h"` present at the top of `CinderImGui.h`. This hopefully shouldn't cause any major issue as cinder is not using the content of that file.

### DIFF
--- a/include/cinder/CinderImGui.h
+++ b/include/cinder/CinderImGui.h
@@ -26,8 +26,11 @@
 #if ! defined( IMGUI_USER_CONFIG )
 #define IMGUI_USER_CONFIG "cinder/CinderImGuiConfig.h"
 #endif
+#if ! defined( CINDER_IMGUI_EXTERNAL )
 #include "imgui/imgui.h"
-#include "imgui/imgui_stdlib.h"
+#else
+#include "imgui.h"
+#endif
 
 #include "cinder/Filesystem.h"
 #include "cinder/CinderGlm.h"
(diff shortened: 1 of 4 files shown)


richardeakin:
I think this is a good move, to allow users to update their Dear Imgui copy locally without having a dirty git tree.

totalgee:
> This PR also removes the unused #include "imgui/imgui_stdlib.h" present at the top of CinderImGui.h. This hopefully shouldn't cause any major issue as cinder is not using the content of that file.

Definitely doesn't cause a major problem ;-), but if I have a project that relied on the overload of `ImGui::InputText()` that takes a string argument (instead of c-string plus length), I should just include the (now) missing `imgui_stdlib.h` header? Thanks!

totalgee:
(To answer myself: it seems to work fine that way...thanks and sorry for the noise.)

Issue #2236, opened 2021-04-06
Some samples with ImGui crash on macOS when window is resized


gaborpapp:
When the window is resized or made full screen the following assertion is thrown:

```
Assertion failed: (g.WithinFrameScope), function Begin, file /cinder/src/imgui/imgui.cpp, line 5380.
```

It is reproducible with PickingFBO (resized or full screen), DeferredShading (full screen), but it does not happen with other ImGui samples like DeferredShadingAdvanced, ClothSimulation, NormalMapping, ShadowMapping.

macOS Catalina 10.15.7, latest cinder HEAD bb15730

simongeilfus:
Actually I think @num3ric 's PR here: https://github.com/cinder/Cinder/pull/2178/files would fix the issue on all platforms. The app draw function is not guarantied to be called for each call to the update function. This is true on Windows and on Mac (probably on Linux too?).

This in turns produces sometimes calls to `ImGui::Render` that are not paired with their respective `ImGui::NewFrame`. This produces more or less the same thing as calling `ImGui::Begin` without calling `ImGui::End`.

PR #2277, opened 2022-02-20
Fixes DearImGui crash when the App issues more than one draw per update

simongeilfus:
As mentioned in https://github.com/cinder/Cinder/issues/2236 the app draw function is not guarantied to be called only once for each call to the update function. And as explained [here](https://github.com/ocornut/imgui/blob/master/imgui_internal.h#L2476) it is illegal to call ImGui functions after `ImGui::EndFrame()/ImGui::Render()`. Failing to do so will trigger or an assert or crash the app. I believe the `NewFrameGuard` was slightly changed when adapted from `Cinder-ImGui` and is now missing a check before the app emits its draw calls.

This means that currently it is only safe to use ImGui inside the scope of the update function.

This can be easily reproduced on Windows by adding a `ImGui::Text( "WillCrash" )` in the app draw method and dragging the window to the corner of the screen.


richardeakin:
Thank you Simon and sorry to have missed this! 🙏

PR #2278, opened 2022-03-16
Update protocol on submodules

afrancois:
If you pull a fresh copy of cinder, Submodules will no longer download as the git: protocol is no longer supported. This chages the  protocol used to https. You may need to do a `git submodule sync` to update the submodules.


andrewfb:
Thanks for this

PR #2282, opened 2022-03-30
Added missing GLFW window move callback

gaborpapp:
Quick simple fix for the window move callback to work in Linux.


andrewfb:
Thanks for this and sorry for the delay! Tested under Ubuntu 24.04 with Wayland.

Issue #1176, opened 2015-11-03
Accessor methods for geom::Primitives?


simongeilfus:
I'm using the ```geom``` API a lot and was wondering if it would make sense to add accessor methods to the different primitive classes. I understand the idea for classes like ```ci::geom::Capsule``` or ```ci::geom::Cube``` is more about carrying the geometry and attributes around than representing the primitive itself. But as there's very little reasons for creating more math classes along the existing ```ci::Sphere```, ```ci::AxisAlignedBox```, etc..., I was wondering if we could extend the geom classes to have accessor methods (just to get access to the protected ```mRadius, mSize, mLength```,...).

The downside of this is that the interface would be a tiny bit more cluttered, but on the other hand it would offer more flexibility. It would for example allow us to use the same data for both gl objects and a physics engine. To be honest I don't see a lot of other use cases, except for scene serialisation and physics engine integration. 

So for instance having this:

```c++
class Sphere : public Source {
  public:
	Sphere();
	Sphere( const ci::Sphere &sphere );

	....

	vec3		getCenter() const { return mCenter; }
	float		getRadius() const { return mRadius; }
	int		getSubdivisions() const { return mSubdivisions; }

  protected:
        ....
};
```

Would allow that kind of code :

```c++
auto source = geom::Sphere().radius( 10.0f ).center( vec3( 0.0f, 100.0f, 0.0f ) );
auto sphereBatch = gl::Batch::create( source, glslProg );
auto sphereRigidBody = bt::createRigidBody( source, 2.0f );
```

I can work on a PR if this is something you would accept.

andrewfb:
I'm open to this - does anybody else have an opinion?

paulhoux:
To be honest, I am opposed to this. The classes in the ```geom``` namespace currently aren't meant to be small, concise, mathematical descriptions of primitives. We have separate ```AxisAlignedBox``` and ```Sphere``` classes for that. Instead, they are "just" there to create a bunch of vertices and normals.

I would personally like to see a few clean, small classes just for intersection and collision purposes, in the cinder namespace, or a different namespace altogether. Apart from ```AxisAlignedBox``` and ```Sphere```, I'd add an ```OrientedBox```, a ```Cone``` and a ```Capsule``` and that's about it. Those classes would contains functions for efficient culling, intersection and collision, no more and no less.

simongeilfus:
Thanks for your feedback guys,
[...]

PR #2289, opened 2022-05-24
added getters to each geom::Object based off of available setters

morphogencc:
Following the suggestion in [#1176](https://github.com/cinder/Cinder/issues/1176), I created getters for each `geom::Object` type based only on what setters were available.


andrewfb:
Thanks for this, Nathan. I'm up to merge this (the presence of these accessors doesn't really conflict with any of the "philosophical" questions regarding the purpose of the `geom` types raised previously, IMO). That said, are you able to re-submit the PR where just `GeomIo.h` has been modified? It looks like some other changes made their way into this PR (hence the 46 files changes).

morphogencc:
Ah, I missed that!  Will resubmit.

PR #2293, opened 2022-07-14
added getters to each geom::Object based off of available setters [Take Two]

morphogencc:
Second attempt at resolving https://github.com/cinder/Cinder/issues/1176! I created getters for each geom::Object type based only on what setters were available.


richardeakin:
Hi @morphogencc, you'll also want to turn off the auto-format settings in your IDE so you aren't creating all the whitespace changes in there. Usually if you want to get your changes into an open source library, the best approach is to create a PR with the minimal set of changes possible. In this case I believe that would be just the `.getXX()` methods you've added.

PR #2300, opened 2022-09-27
Add mp3 support on linux

mincequi:
This PR adds mp3 support on linux within FileAudioLoader.cpp

Please tell me what you think.

### DIFF
--- a/src/cinder/audio/linux/FileAudioLoader.cpp
+++ b/src/cinder/audio/linux/FileAudioLoader.cpp
@@ -128,6 +128,9 @@ struct IStreamSndFile {
 struct IStreamMpg123 {
 	static ssize_t read( void* userData, void* ptr, size_t count ) {
 		ci::IStreamCinder* stream = static_cast<ci::IStreamCinder*>( userData );
+		if (stream->isEof()) {
+			return 0;
+		}
 		size_t pos0 = stream->tell();
 		stream->readData( ptr, count );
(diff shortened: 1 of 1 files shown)


src/cinder/audio/linux/FileAudioLoader.cpp
```diff
@@ -609,8 +612,7 @@ void SourceFileAudioLoader::performSeek( size_t readPositionFrames )
 
 std::vector<std::string> SourceFileAudioLoader::getSupportedExtensions()
 {
-	std::vector<std::string> result;
-	return result;
+	return { "mp3" };
```
richardeakin, line 615: We should add this, but also might as well include file formats that libsndfile supports (https://libsndfile.github.io/libsndfile/formats.html)

src/cinder/audio/linux/FileAudioLoader.cpp
```diff
@@ -128,6 +128,9 @@ struct IStreamSndFile {
 struct IStreamMpg123 {
 	static ssize_t read( void* userData, void* ptr, size_t count ) {
 		ci::IStreamCinder* stream = static_cast<ci::IStreamCinder*>( userData );
+		if (stream->isEof()) {
```
richardeakin, line 131: Did you encounter an issue with this? Not sure if `read()` returning zero bytes during a failure scenario is an ideal solution, but would like to understand how this came about.

PR #2308, opened 2023-05-31
MSW: Decoupling App from Platform

richardeakin:
A while back, we refactored `app::Platform` out of the main App (`AppBasic`, `AppMsw`, `AppLinux`, etc) so that you can use that functionality in an app-less environment. In my case, I'm not even building the cpps for those app classes as I experiment with combining cinder and Diligent Engine ([here](https://github.com/richardeakin/diligent-playground)), and in trying this I see that `PlatformMsw` still relies on an app in a couple places:

1) `getOpenFilePath()`, `getSaveFilePath()` and `getFolderPath()` methods were relying on `AppImplMsw`. This is straightforward to fix by moving the implementations to `PlatformMsw.cpp`
2) `PlatformMsw::refreshDisplays()` is calling the `AppBase` signals for when displays change. I'm not yet sure what to do about this, to me it seems like if it requires AppBase than this method should live on the App. Or the signals should be living on the Platform.. open to discussion about this before I proceed. 

Lastly, I updated the ImageFileBasic app to build as x64 since we no longer support 32-bit builds on windows.


richardeakin:
After some discussion, opted to solve 2) by changing `PlatformMsw::refreshDisplays()` to:

```cpp
void refreshDisplays( std::vector<DisplayRef> *connectedDisplays, std::vector<DisplayRef> *changedDisplays, std::vector<DisplayRef> *disconnectedDisplays );
```

Then the `AppMswImplBasic` calls that and emits any signals back to the app layer on display changes.

Issue #2316, opened 2023-10-26
std::filesystem being used with c++11


dimateos:
The following define has a an OR that ignores the c++ version check, it even ignores the `__has_include` check

https://github.com/cinder/Cinder/blob/4e27a4a5e66fd63000e98afa09d0d0419119ee6b/include/cinder/Filesystem.h#L27-L33

> probably there is no need to check for MSVC when `__cplusplus` is defined

andrewfb:
This should be addressed with #2361

PR #2317, opened 2023-10-26
Fix using std::filesystem when not available

dimateos:
> Ref https://github.com/cinder/Cinder/issues/2316

I simply removed the `_MSC_VER` check.

Feel free to edit or discard.
Cheers!

### DIFF
--- a/include/cinder/Filesystem.h
+++ b/include/cinder/Filesystem.h
@@ -24,12 +24,12 @@
 
 #pragma once
 
-#if ( (! defined(__APPLE__)) && defined(__cplusplus) && __cplusplus >= 201703L && defined(__has_include) && __has_include(<filesystem>)) || defined( _MSC_VER )
-		#define GHC_USE_STD_FS
-		#include <filesystem>
-		namespace cinder {
-			namespace fs = std::filesystem;
-		}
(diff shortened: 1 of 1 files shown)


paulhoux:
The Debug target seems to have a few errors. If you can fix those, I'll run the tests again.

dimateos:
Seems like MSVC does not define `__cplusplus` according to the picked `std=c++XX` prior to 2017.
Moreover, due to legacy code expecting it to still be `199711L` this is disabled by default and requires `/Zc:__cplusplus`.
> ref : https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/

Instead, we can additionally check `_MSVC_LANG` which always reports the c++ std version 
Basically, when `/Zc:__cplusplus` is enabled: `__cplusplus == _MSVC_LANG`
> ref: https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

I added the explanation to the file too

dimateos:
Well there were a few more issues, like the source .cpp not added to the solution (hence I did not update it to begin with). Basically the project compiled in c17 because ghc was not being used, otherwise it would have failed at the linker stage.

> Now I tested both and succesfully compiled the both in c17 and c14, and also samples using the lib :)

PR #2321, opened 2023-12-27
Added comparison functions for floats and doubles…

paulhoux:
…providing a safer way to do things like `double value; if( value == 0.0 ){}`.

Compared to previous versions of the code, I have explicitly added the `std::` namespace to `abs`. While this is not necessary in Visual Studio (it already uses std::abs by default), it may be required for other compilers. Using the C++ function `std::abs` if preferred over the C-function `fabsf`.

I also added versions for `double`s.

### DIFF
--- a/include/cinder/CinderMath.h
+++ b/include/cinder/CinderMath.h
@@ -123,41 +123,74 @@ struct CI_API math<float>
 #define M_PI           3.14159265358979323846
 #endif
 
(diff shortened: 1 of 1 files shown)


andrewfb:
Thanks for this Paul - these look like nice additions, and I'm inclined to merge this shortly unless anyone has any additional comments.

paulhoux:
Once accepted, please `--squash` this branch to a single commit when merging.

PR #2322, opened 2023-12-29
Added string utility functions to trim, filter and convert.

paulhoux:
Compared to previous version of the code, I have renamed the in-place versions to `*InPlace`. They usually perform way better than their copy versions. To prevent issues with the current locale, I have removed calls to `std::setlocale` and left these for the user. The `valueToString` functions use `std::to_string` which performs slightly better than the older `ci::toString()` function.


paulhoux:
I am thinking of renaming `filter` and `filterInPlace` to `removeAnyOf` and `removeAnyOfInPlace`. 

Furthermore, it should be possible to rewrite `findReplace` in such a way that it becomes UTF8 compatible. I could try [this](https://codereview.stackexchange.com/a/216624) code.

totalgee:
Nice. But how come none of the new `valueToString()` functions are declared in the `Utilities.h` file? Also, though Cinder is not yet on C++20, the implementation of these functions should possibly be switched from `std::to_string()` to `std::format()` in C++20, which is very fast and particularly good for (by default) printing floating point numbers with just the required precision. (See the examples [here](https://en.cppreference.com/w/cpp/string/basic_string/to_string).

PR #2324, opened 2023-12-29
Adds gl::ScopedColorMask and gl::ScopedStencilMask.

paulhoux:


andrewfb:
Thanks Paul - this looks good to me.

PR #2325, opened 2024-01-02
Path2d and Shape2d additions: basic shapes and commands.

paulhoux:
In preparation of path rendering, these additions will make it much easier to define paths and shapes from code.


src/cinder/Path2d.cpp
```diff
@@ -264,6 +313,14 @@ void Path2d::arcSegmentAsCubicBezier( const vec2 &center, float radius, float st
 				center.y + r_sin_B - h * r_cos_B, center.x + r_cos_B, center.y + r_sin_B );
 }
 
+float Path2d::angleHelper( const vec2 &u, const vec2 &v ) const
```
andrewfb, line 316: Minor request - unless we are trying to expose this, I'd propose it be an anonymous namespace function and we remove it from the interface

PR #2326, opened 2024-01-03
Added Canvas UI

paulhoux:
This is a 2D equivalent of CameraUi. It allows users to easily add pan and zoom to their 2D applications. CanvasUi is header-only.

### DIFF
--- a/include/cinder/CanvasUi.h
+++ b/include/cinder/CanvasUi.h
@@ -39,13 +39,12 @@ class CanvasUi {
 	{
 		if( this != &rhs ) {
 			mScale = rhs.mScale;
-			mMouse = rhs.mMouse;
-			mClick = rhs.mClick;
+			mMouseDownPos = rhs.mMouseDownPos;
 			mAnchor = rhs.mAnchor;
 			mPosition = rhs.mPosition;
 			mOriginal = rhs.mOriginal;
 			mModelMatrix = rhs.mModelMatrix;
-			mIsDirty = rhs.mIsDirty;
+			mModelCached = rhs.mModelCached;
 			mMouseWheelMultiplier = rhs.mMouseWheelMultiplier;
 			mWindow = rhs.mWindow;
 			mSignalPriority = rhs.mSignalPriority;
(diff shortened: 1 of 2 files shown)


andrewfb:
Thanks for this - I think it could be a really nice addition to Cinder. Any thoughts on a mode - maybe a constructor parameter - that lets me opt-in to left / right / middle buttons serving as drag? My thought being, for many applications, I'd want the left and potentially right mouse buttons to be free for scene interaction, and the middle button for panning.

Also, thoughts on `zoomIn( float percent )` and `zoomOut( float percent )` function that a user could easily map to keyboard hotkeys or similar?

paulhoux:
Those are good suggestions. Will make changes as soon as time permits.

paulhoux:
Actually, while the suggestions are good, they are not straightforward to implement. I based this implementation upon the `CameraUi`, which also doesn't support custom mouse/keyboard mappings. If I am going to add that to `CanvasUi`, we should also add it to `CameraUi` and come up with a `struct` of some kind which handles the mapping.

Regarding `zoomIn` and `zoomOut`: scaling is done relative to the last recorded mouse position. Keeping track of this position is the most important part of `CanvasUi`, as it requires conversion to and from window position. Adding keyboard shortcuts to zoom in or out would not be sufficient: to use this intuitively would also require us to reposition the scale origin to the center of the window (I think). At the very least, we would have to think about the desired behavior.

Again, the `CameraUi` does not have support for assigning keys and mouse buttons. Can't we accept `CanvasUi` as-is and add those features later if needed?

include/cinder/CanvasUi.h
```diff
@@ -0,0 +1,222 @@
...
+	//! Connects to mouseDown, mouseDrag and mouseWheel signals of \a window, with optional priority \a signalPriority.
+	void connect( const app::WindowRef &window, int signalPriority = 0 )
+	{
+		if( window == mWindow && signalPriority == mSignalPriority )
+			return;
+
+		disconnect();
+
+		mWindow = window;
+		mWindowSize = window->getSize();
```
diasbruno, line 67: Should this go on the `if` below?

andrewfb:
Thanks for all this Paul - continued and completed by #2354

PR #2333, opened 2024-07-03
Fix Could NOT find MPG123 on Ubuntu 24.04 LTS

smdaa:
Platform : Ubuntu 24.04 LTS

I installed dependencies found here https://libcinder.org/docs/guides/linux-notes/ubuntu.html
I faced this error:
```
CMake Error at /usr/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find MPG123 (missing: MPG123_INCLUDE_DIR)
Call Stack (most recent call first):
  /usr/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  proj/cmake/modules/FindMPG123.cmake:35 (find_package_handle_standard_args)
  proj/cmake/platform_linux.cmake:143 (find_package)
  CMakeLists.txt:29 (include)
```

Here is a fix i found

### DIFF
--- a/proj/cmake/modules/FindMPG123.cmake
+++ b/proj/cmake/modules/FindMPG123.cmake
@@ -8,7 +8,7 @@
 
 set( MPG123_FOUND false )
 
-set( MPG123_INCLUDE_DIRS /opt/local/include /usr/local/include /usr/include )
+set( MPG123_INCLUDE_DIRS /opt/local/include /usr/local/include /usr/include /usr/include/x86_64-linux-gnu )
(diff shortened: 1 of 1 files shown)


richardeakin:
Ran into this today as well and the fix looks good to me. I don't yet see a more generic way of solving this so going to merge as Ubuntu 24.04 builds are broken without.

Only info I could find on the new (to me, anyway) include path: https://unix.stackexchange.com/questions/586792/difference-between-usr-lib-and-usr-lib-x86-64-linux-gnu-folders

PR #2332, opened 2024-06-18
Update OpenGL loader to one generated by glad 0.1.36

totalgee:
- in a later step, we should update to glad2.

PR #2339, opened 2025-01-21
Fix Windows warnings about APIENTRY from GLAD library

totalgee:
This is an alternative to my PR #2332, which didn't actually solve the problem with these GLAD warnings with MSVC, and not as "heavy" as updating to GLAD 2:

```
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.20348.0\shared\minwindef.h(130,1): warning C4005: 'APIENTRY': macro redefinition
1>(compiling source file '../../src/cinder/gl/wrapper.cpp')
1>C:\Users\holag\Documents\Work\...\cinder\include\glad\glad.h(518,1):
1>see previous definition of 'APIENTRY'
```

This happens when building with MSVC, depending on the include order of GLAD headers vs Windows headers that define `APIENTRY`. Adding the push/pop seems like a simple solution to the problem.

### DIFF
--- a/include/glad/glad.h
+++ b/include/glad/glad.h
@@ -466,15 +466,14 @@
 #ifndef __glad_h_
 #define __glad_h_
 
-#pragma push_macro("APIENTRY")
-
-
 #ifdef __gl_h_
 #error OpenGL header already included, remove this include, glad already provides it
 #endif
 #define __gl_h_
 
 #if defined(_WIN32) && !defined(APIENTRY) && !defined(__CYGWIN__) && !defined(__SCITECH_SNAP__)
+#pragma push_macro("APIENTRY")
+#define GLAD_PUSHED_APIENTRY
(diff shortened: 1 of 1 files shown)


totalgee:
Hmm, unfortunately this doesn't work on other platforms (see compile error e.g. on Ubuntu above). The macro (referenced by other macros) is undefined when it is needed later (compiling `glad.c`)...or something like that.

totalgee:
Okay, fixed the issue so it works on all platforms, and still builds Cinder projects without those annoying warnings on Windows!

totalgee:
I would be really happy if this got merged, because it's so nice to have no more compile warnings on my projects.

andrewfb:
This looks good to me and fixes a long-standing annoyance - thanks for this.

totalgee:
Thank you for merging!

PR #2344, opened 2025-02-28
Added gl::StencilOp, gl::stencilOpSeparate, gl::StencilFunc…

paulhoux:
…and gl::stencilFuncSeparate methods, as well as their push/pop/get variants, to control stencil buffer state. Also added matching gl::ScopedStencilOp and gl::ScopedStencilFunc classes.

While not many people often use the stencil buffer, in recent projects I have used it extensively for effects and path rendering. Being able to properly set and restore the stencil state comes in really handy, especially when doing nested operations.

### DIFF
--- a/include/cinder/gl/Context.h
+++ b/include/cinder/gl/Context.h
@@ -76,6 +76,44 @@ class CI_API Context {
 #endif
 	};
 
+	struct CI_API StencilOp
+	{
+		StencilOp() = default;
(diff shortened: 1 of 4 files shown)


paulhoux:
If it's all the same, I think I'll go ahead and merge this. I have been using it for the past couple of weeks and it's working well for me. Will give it a few more days, so let me know if you have any objections.

andrewfb:
Tested these locally - thanks for this Paul and sorry for the delay

PR #2345, opened 2025-04-16
Fix a bug when getting a gzip header extra field with inflate()

npt-1707:
**Description**
This PR fixes a potential vulnerability in inflate() that was cloned from lib but did not receive the security patch. The original issue was reported and fixed under https://github.com/madler/zlib/commit/eff308af425b67093bab25f80f1ae950166bece1.
This PR applies the same patch to eliminate the vulnerability.

References
https://nvd.nist.gov/vuln/detail/cve-2022-37434
https://github.com/madler/zlib/commit/eff308af425b67093bab25f80f1ae950166bece1

### DIFF
--- a/src/zlib/inflate.c
+++ b/src/zlib/inflate.c
@@ -758,9 +758,10 @@ int flush;
                 copy = state->length;
                 if (copy > have) copy = have;
                 if (copy) {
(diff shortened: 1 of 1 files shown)


andrewfb:
Thanks for this - it should be addressed in #2355

PR #2349, opened 2025-07-02
Added missing MouseEvent handling in ImGui_ImplCinder_MouseUp

simongeilfus:
Current behavior prevents other parts of the app from using mouseUp events. When ImGui windows are interacted with, they shouldn't trigger mouseUp handlers elsewhere in the app.

This PR fixes that by handling the event using `io.WantCaptureMouse` in the `ImGui_ImplCinder_MouseUp` callback.


andrewfb:
Thanks for this - merged manually with 0597e8d

PR #2350, opened 2025-08-11
Fix ImGui Ctrl+Click functionality on Linux

gaborpapp:
Update mouse event handlers in CinderImGui to set modifier key states (KeyCtrl, KeyShift, KeyAlt, KeySuper) from MouseEvent. Previously these were only set during keyboard events, causing ImGui features like Ctrl+Click on sliders to not work reliably on Linux.

The fix ensures modifier states are synchronized with ImGui during all mouse events (MouseDown, MouseUp, MouseMove, MouseDrag, MouseWheel).


andrewfb:
Thanks for this, Gabor - tested this locally under Linux.

PR #2351, opened 2025-08-11
Fix keyboard event propagation when ImGui window has focus

gaborpapp:
When an ImGui window has focus, all keyboard events are blocked from reaching the application, preventing other keyboard shortcuts handled in the app `keyDown` from working. This occurs because keyboard navigation is enabled by default, causing ImGui to capture all keyboard input for navigation purposes.

The proposed fix with the `NavNoCaptureKeyboard` flag prevents ImGui from setting `WantCaptureKeyboard=true` when only navigation is active (not text input). This allows the event handlers to correctly propagate keyboard events to the application when ImGui doesn't actually need them for text input.

### DIFF
--- a/src/cinder/CinderImGui.cpp
+++ b/src/cinder/CinderImGui.cpp
@@ -557,7 +557,10 @@ bool ImGui::Initialize( const ImGui::Options& options )
 	auto context = ImGui::CreateContext();
 	
 	ImGuiIO& io = ImGui::GetIO();
-	if( options.isKeyboardEnabled() ) io.ConfigFlags |= ImGuiConfigFlags_NavEnableKeyboard; // Enable Keyboard Controls
+	if( options.isKeyboardEnabled() ) {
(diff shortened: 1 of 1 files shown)


andrewfb:
Thanks for this, Gabor. This sounds promising but I was having trouble testing it meaning, I didn't see the behavior change I'd expect. Do you mind walking me through how to test it?

gaborpapp:
I usually have Escape quitting the application:
```
void CinderApp::keyDown( KeyEvent event )
{   
    switch ( event.getCode() )
    {
        case KeyEvent::KEY_ESCAPE:
            quit();
            break;
    }
} 
```

Without this change, if the ImGui window is in focus Esc does not do anything. I mainly tested it on Linux and Mac, though.

andrewfb:
Thanks for this - I tested it locally on both Linux and Windows. I don't believe MSW behavior changed, but the fix was apparent with Linux.

PR #2353, opened 2025-09-17
Dear ImGui v1.92.3

andrewfb:
This updates Cinder's embedded Dear ImGui to version [1.92.3](https://github.com/ocornut/imgui/releases/tag/v1.92.3), released a few hours ago. 

This update also should provide a path for macOS Retina applications to make better use of Dear ImGui, and provides a new sample `ImGui` that exercises some of the basics as well as a few Cinder-specific bits of functionality:
image

Tested on macOS, MSW, and Linux.


andrewfb:
This replaces #2311

PR #2383, opened 2025-11-21
Windows-based support for ending a move/resize window interaction.

Gazoo101:
As this is my first contribution, I'll both apologize and in advance do my best to address necessary changes.

I've found a need to be able to tell if a resizing operation has ceased on windows, at which point I wish to re-render a windows content but not before. Determining this without the actual OS signal isn't straight-forward so I opted to add it in for the platform I'm working on. On Mac, admittedly, this will not trigger, and I'm a bit of a Mac novice, so testing it isn't so feasible for me at the moment.

An explicit test for a user interaction is a bit tricky, but I'm open to suggestions.

### DIFF
--- a/include/cinder/app/Window.h
+++ b/include/cinder/app/Window.h
@@ -404,6 +404,9 @@ class CI_API Window : public std::enable_shared_from_this<Window> {
 	EventSignalWindow&	getSignalResize() { return mSignalResize; }
 	void 				emitResize();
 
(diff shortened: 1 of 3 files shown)


Gazoo101:
Makes sense to leverage GLFW if it can address this need universally on multiple platforms.

Let me take this opportunity to thank you for all your work, and recent continued contributions to LibCinder @andrewfb . I personally appreciate it so very much, and I reckon lots of less other vocal individuals do as well.

PR #2387, opened 2025-12-01
postResize and isResizing()

andrewfb:
This begins with and supersedes #2383. It implements the `postResize` signal proposed by that PR and adds a `Window::isResizing()` method. It does change the semantics some so that it only fires on resize, and not move. @Gazoo101 let me know if that meets your use case too but I think it's a bit cleaner. This also implements a macOS version of the same, and updates test/windowTest to exercise it, specifically changing the background color while a resize is underway. I don't believe GLFW has native support for this concept and a bit of reading made me think it may be possible under Wayland, but it's not currently implemented.

### DIFF
--- a/test/windowTest/src/windowTestApp.cpp
+++ b/test/windowTest/src/windowTestApp.cpp
@@ -40,6 +40,7 @@ class WindowTestApp : public App {
 	void windowClose();
 	void windowMouseDown( MouseEvent &mouseEvt );
 	void displayChange();
+	void windowPostResize();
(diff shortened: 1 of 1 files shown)


Gazoo101:
@andrewfb Thanks for creating this PR - Yep - it definitely meets my needs, and avoids me having to use a flag separately to track whether a resize or move was occurring. I tested it in my own application and it appears to work as intended 👍 

Thank you again for implementing it!

Issue #2337, opened 2024-12-04
Can't detect question mark '?' from KeyEvent code (due to shift being pressed?)


totalgee:
When I press the question mark key '?' on the a keyboard (for different layouts, such as US English or Spanish), I never get the `KeyEvent::KEY_QUESTION` code, but a different one.

For example, in my `keyDown()` handler:
```
    if (event.getChar() == '?') {
        fmt::print("Got key: '{}' code: {} native: {} translated: {}", ch,
            event.getCode(),
            event.getNativeKeyCode(),
            KeyEvent::translateNativeKeyCode(event.getNativeKeyCode()));
    }
```
For a US English layout, I get this:
`Got key: '?' code: 47 native: 191 translated: 47` -- this is '/', which is the unshifted key I'm pressing.
For a Spanish keyboard layout, I get this:
`Got key: '?' code: 91 native: 219 translated: 91` -- this is '[', which is not the unshifted key; you would have ''' (apostrophe) on a Spanish keyboard as the unshifted '?' key.

The actual `KeyEvent::KEY_QUESTION` value is 63, which is also the ASCII value of '?'. But I never see that value as the "Cinder key code". I know I can (and do) just test the `getChar()` value, but it would be nice to be able to just use a single `switch` statement on the KeyEvent codes. I can't check `event.isShiftDown()` plus a different code in a compatible way, since the code is different for different layouts.

totalgee:
Sort of related, but not exactly the same issue. I think this issue is just the way things are... I suppose the "proper" way to deal with different keyboard layouts is to allow a way for the user to make their own mapping of keys to functionality, so if it's wrong, they can adjust it as desired. Or, in my case if I really want to check for `'?'`, just I'll use `event.getChar()` for some that and similar cases, and use `event.getCode()` for others.

PR #2388, opened 2025-12-04
Add keyChar() event; modifier support for FileDropEvent

### LINKED ISSUE #2194: ImGui does not support non-English keyboard layouts (e.g. Spanish)
Hi there, this is to note that other (e.g. Spanish) keyboard layouts don't play well with the InputText widgets in Cinder's ImGui implementation. I have tried running an [online demo](https://pthom.github.io/imgui_manual_online/manual/imgui_manual.html) (built using emscripten) which does work properly...so not sure if this is a Cinder issue, a GLFW issue, or an ImGui issue (unlikely).

In particular -- certain keys on a Spanish keyboard work by pressing the "AltGr" key (which is equivalent to Ctrl+Alt) and then another key. For example, if you're a programmer, you regularly need to use these keys: `{}[]|@#\~`. All these keys require a special key combination. When typing in an InputText (and Multiline) widget in Cinder, you currently get nothing when trying to type these keys (switching to an EN-US keyboard works of course).

### LINKED ISSUE #2270: Text input does not support modifier keys other than Alt,Ctrl,Shift,Meta on linux.
I am using the UK extended keyboard layout. So, for example, typing ` then a writes à, typing AltGr+6 then a writes â.
I am also using the compose key. So for example, typing Compose then o then e writes œ.
See https://en.wikipedia.org/wiki/Modifier_key#Accented_characters .

None of these examples work on the TexBox sample, only the unmodified keys are taken into account. Some debugging seems to show that the modifier keys like accents and compose are not implemented. I guess it should be around this file src/cinder/app/linux/AppImplLinuxGlfw.cpp, maybe in extractKeyModifiers() and modifyChar() ?

I volunteer to contribute if needed.

andrewfb:
## Summary
- Add `keyChar()` App-level virtual method and `getSignalKeyChar()` for text input handling
  - Properly handles dead keys and composed characters (e.g., `'` + `e` = `é`)
- Adds modifier key support to `FileDropEvent` (Shift, Ctrl, Alt, Meta) on macOS and Windows
  - Platform limitations prevent this on Linux, at least without X11 and Wayland-specific code
- Update ImGui integration to use `keyChar` for text input

### When to use keyChar() vs keyDown()/keyUp()
  **Use `keyChar()`** for text entry - when you need the actual character the user intended to type. Handles dead keys, IME, AltGr combinations, and non-English layouts correctly.

  **Use `keyDown()`/`keyUp()`** for controls and shortcuts - when you need to detect physical key presses regardless of text output (e.g., WASD movement, hotkeys, game controls).

### Related Issues
- Fixes #2194 - ImGui does not support non-English keyboard layouts (e.g. Spanish)
- Fixes #2270 - Text input does not support modifier keys other than Alt,Ctrl,Shift,Meta on linux
- Affects #2337 - Clarifies distinction between key codes (physical keys) and characters

### DIFF
--- a/include/cinder/app/FileDropEvent.h
+++ b/include/cinder/app/FileDropEvent.h
@@ -55,15 +55,15 @@ class CI_API FileDropEvent : public Event {
 	//! Returns the path for file number \a index.
 	const fs::path&					getFile( size_t index ) const { return mFiles.at(index); }
 
-	//! Returns a bitwise-or of modifier key flags active during the drop
+	//! Returns a bitwise-or of modifier key flags active during the drop. Unimplemented on Linux.
 	unsigned int	getModifiers() const { return mModifiers; }
-	//! Returns whether the Shift key was held during the drop
+	//! Returns whether the Shift key was held during the drop. Unimplemented on Linux.
 	bool			isShiftDown() const { return ( mModifiers & SHIFT_DOWN ) != 0; }
-	//! Returns whether the Alt (Option on macOS) key was held during the drop
+	//! Returns whether the Alt (Option on macOS) key was held during the drop. Unimplemented on Linux.
 	bool			isAltDown() const { return ( mModifiers & ALT_DOWN ) != 0; }
-	//! Returns whether the Control key was held during the drop
+	//! Returns whether the Control key was held during the drop. Unimplemented on Linux.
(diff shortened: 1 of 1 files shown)


totalgee:
This is great -- thank you!

Issue #2411, opened 2026-07-28
Defining ImGui config macros in app code silently causes an ODR violation


totalgee:
`CinderImGui.h` sets `IMGUI_USER_CONFIG` and pulls in `imgui.h`, so apps naturally reach ImGui through it. But Cinder's bundled `src/imgui/*.cpp` are compiled **without** any app-side configuration. If an app defines an ImGui config macro before including `CinderImGui.h`, its translation units get a different ImGui struct layout than the one compiled into `cinder.lib` -- an ODR violation with no diagnostic from the compiler or linker.

In my case:

```cpp
#define IMGUI_DISABLE_OBSOLETE_FUNCTIONS   // in my own header
#include "cinder/CinderImGui.h"
```

That macro removes a 32-byte block from `ImGuiIO` (`FontGlobalScale`, the legacy clipboard function pointers, and `ClipboardUserData`). Measured in translation units on both sides:

| | app TU | cinder TU |
|---|---|---|
| `sizeof(ImGuiIO)` | 3000 | 3032 |
| `sizeof(ImGuiContext)` | 10784 | 10816 |
| `offsetof(ImGuiContext, InputTextState)` | 9208 | 9240 |

### Symptom

Clicking **any** text field appeared to focus it for one frame, then focus was lost. `InputTextEx()` was calling `ClearActiveID()` from its "ActiveId was set through another widget" edge case, because `GetInputTextState(id)` returned `NULL` even though `g.InputTextState.ID == id` (an "impossible" result, from reading the code ;-).

`GetInputTextState()` is `inline` in `imgui_internal.h`. At `/Od` (which implies `/Ob0`), every TU referencing it emits an out-of-line COMDAT, and the linker keeps one arbitrarily. My app's copy -- compiled with the smaller `ImGuiIO` -- won, so it read `InputTextState.ID` 32 bytes off. Release builds were unaffected, because `/O2` expands the function at each call site and no COMDAT is ever selected.

### Why it was so hard to find

- Latent for a long time. It only surfaced once my app referenced an ImGui internal (`ImGui::GetInputTextState()`) for custom `InputText` behaviour. Merely *compiling* that reference was enough; the code never had to run.
- Debug-only, so it looked like a compiler or optimisation bug.
- A `git bisect` landed on the commit that added the reference, which contained no code on any executed path.

### What would have caught it immediately

[...]

PR #2412, opened 2026-07-28
Stop using obsolete ImGui functions and enums

totalgee:
Summary:

- enable IMGUI_DISABLE_OBSOLETE_FUNCTIONS in imconfig.h and port the deprecated uses: tab/nav colour enums, the NavNoCaptureKeyboard flag, clipboard callbacks, Image() to ImageWithBg(), and DX12 legacy SRV descriptors.
- make an inline version of Initialize() so IMGUI_CHECKVERSION() runs in the caller's translation unit, where it can catch an app/library config mismatch instead of only comparing the Cinder lib version to itself.
- to resolve issue #2411.

### Breaking change -- needs a maintainer decision

Enabling `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` affects **downstream apps**, not just Cinder. The most likely thing people hit is renamed colour enums, since many people might customize `style.Colors[]`:

| Deprecated | Current | Since |
|---|---|---|
| `ImGuiCol_TabActive` | `ImGuiCol_TabSelected` | 1.90.9 |
| `ImGuiCol_TabUnfocused` | `ImGuiCol_TabDimmed` | 1.90.9 |
| `ImGuiCol_TabUnfocusedActive` | `ImGuiCol_TabDimmedSelected` | 1.90.9 |
| `ImGuiCol_NavHighlight` | `ImGuiCol_NavCursor` | 1.91.4 |
| `ImGuiChildFlags_Border` | `ImGuiChildFlags_Borders` | 1.91.1 |
| `ImGuiTreeNodeFlags_AllowItemOverlap` | `...AllowOverlap` | 1.89.7 |

plus a handful of functions (`SetWindowFontScale()`, `GetContentRegionMax()`, `SetItemAllowOverlap()`, the old `Combo()`/`ListBox()` callback form) and the `io.FontGlobalScale` / `io.*ClipboardTextFn` members.

The blast radius is bounded: everything obsoleted before ~1.89 is already fully commented out in 1.92, so it is unavailable with or without this macro change! Only the dozen or so items still live in the obsolete section are affected here.

If that's too disruptive, the macro is a single line in `imconfig.h` and I'm happy to comment it out -- the internal migration stands on its own, since these APIs will be deleted upstream eventually.

### `Initialize()` / `InitializeImpl()` split

`IMGUI_CHECKVERSION()` only detects a config mismatch when it is compiled in the **caller's** translation unit. The existing call in `CinderImGui.cpp` compares Cinder against itself and can never fail, which is why #2411 went undiagnosed. Making `Initialize()` an inline header wrapper is the smallest way to get the check instantiated in application code with no change to how apps call it.

Worth flagging: this moves the exported symbol from `Initialize` to `InitializeImpl`, so it is an ABI change for anyone linking a prebuilt `cinder.lib`. Rebuilding resolves it. If you'd prefer a different shape (a separate `ImGui::CheckVersion()` helper apps call themselves, or a differently named impl function), I could adjust it.

### One migration that is not a simple rename

`Image()` → `ImageWithBg()` is the only case where the compiler won't catch you: the old signature took `(tint_col, border_col)`, the new one takes `(bg_col, tint_col)`. Forwarding positionally compiles cleanly and renders incorrectly -- I hit that before spotting it.

Cinder's public `ImGui::Image( Texture2dRef, size, uv0, uv1, tint, border )` signature is unchanged, and `border_col` is still honoured via `ImGuiCol_Border` plus `ImGuiStyleVar_ImageBorderSize`, mirroring ImGui's own obsolete shim in `imgui_widgets.cpp` (including forcing a minimum thickness when the colour has alpha, so a border still appears purely because a colour was supplied). However, the Cinder version of ImGui::Image() still takes the same arguments and order as before, so anyone using that (such as `samples/ImGui` will continue to work as before).

### DX12 is untested

I wrote the SRV descriptor allocator without a DX12 setup to run it on. That said, the previous code could not have worked here regardless: the `LegacySingleSrv*` fields don't exist in 1.92, and the backend now allocates one descriptor per texture rather than one in total. I replaced them with the documented `SrvDescriptorAllocFn` / `SrvDescriptorFreeFn` callbacks over a simple free list on the existing heap, following the pattern in ImGui's `example_win32_directx12`.

It compiles, but someone with a D3D12 project (`samples/D3d12ImageBasic` would be the obvious test) should confirm it before this is relied on.

### How I verified

- Cinder builds clean in Debug and Release (MSVC v143, x64).
- `samples/ImGui` builds and runs; verified interactively that the Tint Color and Border Color controls both affect the image as before.
- grepped `samples/`, `src/cinder/`, `include/cinder/` and `blocks/` for remaining deprecated identifiers -- none found.
- DX12 and non-Windows platforms not built.

### DIFF
--- a/include/cinder/CinderImGui.h
+++ b/include/cinder/CinderImGui.h
@@ -23,6 +23,16 @@
 
 #pragma once
 
+// NOTE: this is not the place to configure ImGui from an application.
+// ImGui configuration macros (IMGUI_DISABLE_OBSOLETE_FUNCTIONS and friends) belong
+// in include/imgui/imconfig.h, which imgui.h includes in *every* translation unit.
+// CinderImGuiConfig.h below is only reached by code that goes through this header --
+// cinder's own imgui*.cpp never define IMGUI_USER_CONFIG and so never see it.
+// Defining such a macro in application code therefore changes the layout of
+// ImGuiIO/ImGuiContext for that application only, while cinder.lib keeps the
+// unmodified layout. That is an ODR violation which neither the compiler nor the
+// linker reports, and it silently corrupts ImGui state. Initialize() below calls
+// IMGUI_CHECKVERSION() from the caller's translation unit to catch it.
 #if ! defined( IMGUI_USER_CONFIG )
 #define IMGUI_USER_CONFIG "cinder/CinderImGuiConfig.h"
 #endif
@@ -112,7 +122,19 @@ namespace ImGui {
 	//! For D3D12 apps: auto-render is not supported since apps manage their own command lists.
 	//! Use GetD3D12SrvHeap() to get the descriptor heap, then call NewFrame(), your GUI code,
 	//! Render(), and ImGui_ImplDX12_RenderDrawData() manually.
-	CI_API bool Initialize( const Options& options = Options() );
+	CI_API bool InitializeImpl( const Options& options );
+
+	//! Deliberately inline: IMGUI_CHECKVERSION() compares the sizes of ImGuiIO,
+	//! ImGuiStyle, ImVec2/4 and ImDrawVert/Idx as seen by the *caller* against those
+	//! compiled into cinder.  Calling it from inside cinder would only ever compare
+	//! cinder against itself, so it must be instantiated in the application's
+	//! translation unit to detect a configuration mismatch across the library
+	//! boundary (see the note at the top of this header).
+	inline bool Initialize( const Options& options = Options() )
+	{
+		IMGUI_CHECKVERSION();
+		return InitializeImpl( options );
+	}
 
 	//! Returns true if ImGui is using the D3D12 backend
 	CI_API bool IsUsingD3D12();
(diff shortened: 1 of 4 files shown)


totalgee:
Verified against samples/D3d12ImageBasic: the allocator and free callbacks are exercised for ImGui's font atlas and the UI renders correctly. The multi-descriptor path (more than one texture live at once in DX12) is not exercised by any existing sample.

totalgee:
Fixed broken Release build for `samples/ImGui` VS2022 project

- removed IgnoreSpecificDefaultLibraries (`LIBCMT;LIBCPMT`) from `samples/ImGui`, which broke its Release x64 link (`/MT` requests exactly the libs being ignored, so the CRT was excluded with thousands of unresolved symbols).
- the same line appears in 67 sample projects; fixing the others seems out of scope here but is probably worth a separate pass.

totalgee:
One last comment -- the most important part of this fix is the protection provided by doing `IMGUI_CHECKVERSION()` in the now-inlined `ImGui::Initialize()`, so the same class of bug can't go silent again on any platform.

totalgee:
Also fixed regression when using the default height (-1) with `ImGui::ListBox()`, due to the switch to `IMGUI_DISABLE_OBSOLETE_FUNCTIONS`. See 7f49ffa

totalgee:
Sorry, that latest commit cc0b1f3 (about the ImGui assert) probably shouldn't have been added to this PR related to "obsolete ImGui functions". It probably deserved its own PR.

Is anyone reviewing these PRs and likely to merge them in? 🤞 🙏  Thanks!

